PR #78, opened 2026-04-26
fix(fuzz): add missing bytes dep for filter pipeline fuzzer

nerdalert:
## Summary

Resolved a compilation error (E0433) in fuzz_filter_pipeline.rs where the bytes crate was referenced but not declared as a dependency in tests/fuzz/Cargo.toml.

## Validation
Successfully runs `cargo build --manifest-path tests/fuzz/Cargo.toml --bin fuzz_filter_pipeline` with the patch.

Ty!


nerdalert:
Closing, got patched on another commit 🎉

PR #80, opened 2026-04-28
feat: add Prometheus metrics foundation with /metrics admin endpoint

### LINKED ISSUE #8: Epic: Prometheus Metrics
## Summary

Expose Prometheus-compatible metrics via a scrape endpoint on the admin
listener (`/metrics`). Uses the `metrics` crate as facade with
`metrics-exporter-prometheus` for export.

## Scope

Individual metrics and cardinality controls are tracked as sub-issues of this
epic (see the Sub-issues list), grouped into HTTP, TCP, Upstream, Error, and
Cardinality Controls.

nerdalert:
Trying to keep this not obnoxiously large. Scoping to metrics foundation: `/metrics`, the Prom recorder, and low cardinality HTTP request metrics to make sure cardinality is a first class citizen here. With `/metrics` we can start thinking about token usage, quota, rate-limit, PoCs and how to scale them.

## Summary

  - Add /metrics scrape endpoint to the admin listener alongside /healthy and /ready
  - Install metrics + metrics-exporter-prometheus as workspace dependencies
  - Emit praxis_http_requests_total counter and praxis_http_request_duration_seconds histogram from both HTTP handler logging hooks
  - Labels: method, status_class (2xx/3xx/4xx/5xx), route (placeholder unknown), cluster (from router or none)

## Details

- The metrics module (protocol/src/http/pingora/metrics.rs) installs a global Prometheus recorder via OnceLock when the admin listener is configured. Both with_body and no_body handlers call emit_request_metrics() in their logging() hook, reading the response status from session.response_written() and the cluster name from a metrics_cluster snapshot on PingoraRequestCtx (since cluster is consumed by filter context construction before logging runs).
- Body-byte histograms (request_body_bytes, response_body_bytes) are intentionally excluded — the ctx byte counters are only populated when filter body hooks run, so the no-body handler path would emit misleading zeros. These will be reintroduced with independent transport-level byte counting.
- Added method_label() in metrics.rs that passes through the 9 RFC 9110 standard methods and collapses everything else to "OTHER". The caller in emit_request_metrics now runs the raw method through this function before labeling. Custom methods like PURGE or garbage strings can no longer create unbounded cardinality.

Deferred to follow-up PRs to try and keep this already too large PR reviewable 😬: active request gauge, TCP metrics, upstream metrics, error counters, configurable label sets, path grouping, route name schema changes.

## Cardinality

  I intentionally kept the initial label set conservative to give some time to think about scaling.

  - `status_class` is used instead of raw status codes to avoid creating one series per status.
  - `route` is currently emitted as `unknown` instead of using raw request paths. Raw paths are intentionally avoided because IDs, tenant names, model names, or arbitrary user-controlled path segments can create unbounded series.
  - `cluster` comes from the configured router cluster name, or `none` when no cluster was selected. This assumes clusters remain operator-controlled config values, not per-request dynamic values.
  - `method` is the only request-derived label. It is expected to remain low-cardinality; a follow-up can normalize unknown/custom methods to `OTHER` if needed.

Follow-up work for #8 should keep this rule: metrics labels must come from bounded config or normalized enums, not raw request data. Route naming/path grouping should be added before `route` is populated with anything more specific than `unknown`.

## Test plan

  - Unit tests for status_class() conversion (7 cases covering all classes + edge cases)
  - Unit test for prometheus_response() — verifies 200 status, correct content type, valid UTF-8 body
  - All 224 protocol crate tests pass
  - Clippy clean, cargo +nightly fmt clean
  - Manual: start proxy with admin config, curl localhost:9901/metrics returns Prometheus text format
  - Manual: send requests through proxy, verify praxis_http_requests_total and praxis_http_request_duration_seconds appear with correct labels

Manual validation/demo output: https://gist.github.com/nerdalert/e35609d4b38693fb27ae01e4b9644abd

Refs #8

Ty!

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -74,15 +74,22 @@ See [hot-reload.yaml] for an example.
 ## Admin
 
 `admin.address` binds a separate HTTP listener that serves
-`/ready` and `/healthy`. `/healthy` returns `200 OK` with
-`{"status":"ok"}` once the server is accepting
-connections (liveness). `/ready` returns per-cluster
-health status with healthy/unhealthy/total counts when
-active health checks are configured; it returns 503
-when any cluster has zero healthy endpoints. Without
-health checks, `/ready` returns `{"status":"ok"}`. Any
-other path returns 404. Useful for orchestrator health
-checks without exposing them on the main listeners.
+`/healthy`, `/ready`, and `/metrics`.
+
+- `/healthy` returns `200 OK` with `{"status":"ok"}`
+  once the server is accepting connections (liveness).
+- `/ready` returns per-cluster health status with
+  healthy/unhealthy/total counts when active health
+  checks are configured; it returns `503 SERVICE UNAVAILABLE` when any
+  cluster has zero healthy endpoints. Without health
+  checks, `/ready` returns `{"status":"ok"}`.
+- `/metrics` returns Prometheus text exposition format
+  with HTTP request metrics (`praxis_http_requests_total`,
+  `praxis_http_request_duration_seconds`).
+
+Any other path returns `404 NOT FOUND`. Useful for orchestrator
+health checks and monitoring without exposing them on
+the main listeners.
 
 ```yaml
 admin:
--- a/protocol/src/http/pingora/handler/mod.rs
+++ b/protocol/src/http/pingora/handler/mod.rs
@@ -12,7 +12,7 @@ use praxis_filter::{CompressionConfig, FilterPipeline};
 use tokio::sync::Semaphore;
 use tracing::{debug, warn};
 
-use super::context::PingoraRequestCtx;
+use super::{context::PingoraRequestCtx, metrics};
 
 /// Shared hop-by-hop header stripping logic.
 mod hop_by_hop;
@@ -211,6 +211,41 @@ async fn logging_cleanup(pipeline: &FilterPipeline, ctx: &mut PingoraRequestCtx)
     }
 }
 
+/// Emit Prometheus metrics for a completed HTTP request.
+///
+/// No-op when the Prometheus recorder has not been installed.
+fn emit_request_metrics(session: &Session, ctx: &PingoraRequestCtx) {
+    if !metrics::is_recorder_installed() {
+        return;
+    }
+
+    let status_code = session.response_written().map_or(0, |resp| resp.status.as_u16());
+    let status_class = metrics::status_class(status_code);
+
+    let request_method = session.req_header().method.as_str();
+    let raw_method = if request_method.is_empty() {
+        ctx.request_snapshot.as_ref().map_or("UNKNOWN", |r| r.method.as_str())
+    } else {
+        request_method
+    };
+    let method = metrics::method_label(raw_method);
+
+    let cluster = ctx.metrics_cluster.as_ref().map_or_else(
+        || ::metrics::SharedString::const_str("none"),
+        |cluster| ::metrics::SharedString::from(Arc::clone(cluster)),
+    );
+
+    let labels = metrics::RequestMetricLabels {
+        method,
+        status_class,
+        route: "unknown",
+        cluster,
+    };
+
+    let duration_secs = ctx.request_start.elapsed().as_secs_f64();
+    metrics::record_request_metrics(labels, duration_secs);
+}
+
 /// Record a passive health observation for the selected upstream endpoint.
 ///
 /// Called from the `logging` hook on every completed request. Determines
@@ -220,7 +255,8 @@ async fn logging_cleanup(pipeline: &FilterPipeline, ctx: &mut PingoraRequestCtx)
 /// No-op when no upstream was selected, no health registry is available,
 /// or passive checking is not configured for the cluster.
 fn record_passive_health(pipeline: &FilterPipeline, error: Option<&pingora_core::Error>, ctx: &PingoraRequestCtx) {
-    let Some(ref cluster_name) = ctx.cluster else {
+    let cluster_name = ctx.cluster.as_ref().or(ctx.metrics_cluster.as_ref());
+    let Some(cluster_name) = cluster_name else {
         return;
     };
     let Some(idx) = ctx.selected_endpoint_index else {
@@ -516,8 +552,26 @@ mod tests {
     fn passive_health_missing_cluster_is_noop() {
         let (pipeline, mut ctx) = make_passive_scenario(Some(1), Some(1));
         ctx.cluster = None;
+        ctx.metrics_cluster = None;
         let error = make_error();
         record_passive_health(&pipeline, Some(&error), &ctx);
+    }
+
+    #[test]
+    fn passive_health_falls_back_to_metrics_cluster() {
+        let (pipeline, mut ctx) = make_passive_scenario(Some(2), Some(1));
+        ctx.cluster = None;
+        ctx.metrics_cluster = Some(Arc::from("test-cluster"));
+        let error = make_error();
+        record_passive_health(&pipeline, Some(&error), &ctx);
+        record_passive_health(&pipeline, Some(&error), &ctx);
+
+        let registry = pipeline.health_registry().unwrap();
+        let entry = registry.get("test-cluster").unwrap();
+        assert!(
+            !entry.endpoints()[0].is_healthy(),
+            "fallback to metrics_cluster should still record passive health"
+        );
     }
 
     #[test]
--- a/protocol/src/http/pingora/handler/no_body.rs
+++ b/protocol/src/http/pingora/handler/no_body.rs
@@ -18,8 +18,8 @@ use tokio::sync::Semaphore;
 use tracing::{debug, warn};
 
 use super::{
-    adjust_compression, handle_connect_failure, logging_cleanup, record_passive_health, request_filter,
-    response_filter, upstream_peer, upstream_request, via,
+    adjust_compression, emit_request_metrics, handle_connect_failure, logging_cleanup, record_passive_health,
+    request_filter, response_filter, upstream_peer, upstream_request, via,
 };
 use crate::http::pingora::context::PingoraRequestCtx;
 
@@ -172,8 +172,9 @@ impl ProxyHttp for PingoraHttpHandlerNoBody {
         upstream_peer::execute(ctx)
     }
 
-    async fn logging(&self, _session: &mut Session, e: Option<&pingora_core::Error>, ctx: &mut Self::CTX) {
+    async fn logging(&self, session: &mut Session, e: Option<&pingora_core::Error>, ctx: &mut Self::CTX) {
         let pipeline = self.pipeline.load();
+        emit_request_metrics(session, ctx);
         record_passive_health(&pipeline, e, ctx);
         logging_cleanup(&pipeline, ctx).await;
     }
--- a/protocol/src/http/pingora/handler/with_body.rs
+++ b/protocol/src/http/pingora/handler/with_body.rs
@@ -19,8 +19,8 @@ use tokio::sync::Semaphore;
 use tracing::{debug, warn};
 
 use super::{
-    adjust_compression, handle_connect_failure, logging_cleanup, record_passive_health, request_body_filter,
-    request_filter, response_body_filter, response_filter, upstream_peer, upstream_request, via,
+    adjust_compression, emit_request_metrics, handle_connect_failure, logging_cleanup, record_passive_health,
+    request_body_filter, request_filter, response_body_filter, response_filter, upstream_peer, upstream_request, via,
 };
 use crate::http::pingora::context::PingoraRequestCtx;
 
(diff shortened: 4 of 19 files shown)


docs/configuration.md
```diff
@@ -20,15 +20,22 @@ insecure_options:      # Optional. Dev/test overrides. See development.md.
...
-health checks, `/ready` returns `{"status":"ok"}`. Any
-other path returns 404. Useful for orchestrator health
-checks without exposing them on the main listeners.
+`/healthy`, `/ready`, and `/metrics`.
+
+- `/healthy` returns `200 OK` with `{"status":"ok"}`
+  once the server is accepting connections (liveness).
+- `/ready` returns per-cluster health status with
+  healthy/unhealthy/total counts when active health
+  checks are configured; it returns 503 when any
```
shaneutt:
```suggestion
  checks are configured; it returns `503 SERVICE UNAVAILABLE` when any
```

docs/configuration.md
```diff
@@ -20,15 +20,22 @@ insecure_options:      # Optional. Dev/test overrides. See development.md.
...
+  cluster has zero healthy endpoints. Without health
+  checks, `/ready` returns `{"status":"ok"}`.
+- `/metrics` returns Prometheus text exposition format
+  with HTTP request metrics (`praxis_http_requests_total`,
+  `praxis_http_request_duration_seconds`).
+
+Any other path returns 404. Useful for orchestrator
```
shaneutt:
```suggestion
Any other path returns `404 NOT FOUND`. Useful for orchestrator
```

protocol/src/http/pingora/health/service.rs
```diff
@@ -164,6 +187,7 @@ impl ServeHttp for PingoraHealthService {
 
         match path.as_str() {
             "/healthy" => json_response(200, br#"{"status":"ok"}"#),
+            "/metrics" => prometheus_response(),
```
shaneutt, line 200: It looks like we might not have integration tests for `/metrics`?

We'll need to add those, in addition to examples + example tests.

protocol/src/http/pingora/health/service.rs
```diff
@@ -463,4 +487,18 @@ mod tests {
...
+        metrics::install_prometheus_recorder();
+        let resp = prometheus_response();
+        assert_eq!(resp.status(), 200, "should be 200 when recorder is installed");
+        assert_eq!(
+            resp.headers()["Content-Type"],
+            "text/plain; version=0.0.4; charset=utf-8",
+            "content-type should be Prometheus text format"
+        );
+        let body_str = std::str::from_utf8(resp.body());
+        assert!(body_str.is_ok(), "prometheus body should be valid UTF-8");
```
shaneutt: I believe an empty string will pass all three assertions here. Consider `assert!(!body.is_empty(), "prometheus body should not be empty")` or some additional more extensive validity checks 🤔

protocol/src/http/pingora/metrics.rs
```diff
@@ -0,0 +1,209 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Shane Utt
```
shaneutt:
```suggestion
// Copyright (c) 2026 Praxis Contributors
```

protocol/src/http/pingora/metrics.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Status class (e.g. `"2xx"`).
+    pub status_class: &'a str,
+    /// Route name or `"unknown"`.
+    pub route: &'a str,
+    /// Cluster name or `"none"`.
+    pub cluster: &'a str,
+}
+
+/// Record HTTP request metrics for a completed request.
+pub fn record_request_metrics(labels: &RequestMetricLabels<'_>, duration_secs: f64) {
```
shaneutt: Doesn't need to be pub:


```suggestion
pub(crate) fn record_request_metrics(labels: &RequestMetricLabels<'_>, duration_secs: f64) {
```

protocol/src/http/pingora/metrics.rs
```diff
@@ -0,0 +1,209 @@
...
+    let label_set = [
+        ("method", labels.method.to_owned()),
+        ("status_class", labels.status_class.to_owned()),
+        ("route", labels.route.to_owned()),
+        ("cluster", labels.cluster.to_owned()),
```
shaneutt: **micro-optimization**: the first three here don't need to be allocated on every HTTP request, we should look into making them `Cow::Borrowed` or something to avoid it.

nerdalert:
praxis-bot closed this on size. Will split it up, following up with the review fixes here though for brevity:

  1. Method label source - emit_request_metrics now reads from session.req_header().method first, falling back to ctx.request_snapshot only if empty.
  2. Admin function rename - Introduced add_admin_endpoints_to_pingora_server as the primary function; old name kept as a thin wrapper for existing
  callers.
  3. Per-request allocations - Static labels (method, status_class, route) are now &'static str and use the => macro syntax (zero-alloc). Only cluster
  allocates once per request.
  4. Narrower visibility - record_request_metrics and RequestMetricLabels are now pub(crate).
  5. Integration/example tests - Added smoke test for /metrics, schema test for admin-interface.yaml, and integration test that sends a proxy request then
   asserts metric names and label values.
  6. Stronger prometheus unit test - Records a unique metric before calling prometheus_response(), asserts body is non-empty and contains the metric name.
  7. Copyright header - Changed to 2026 Praxis Contributors.
  8. Docs status wording — 503 → `503 SERVICE UNAVAILABLE`, 404 → `404 NOT FOUND`.

nerdalert:
I'm reopening with convention bypass. Trying to get a smaller PR might be a painful review since tests will be all over the place. This is 400 LOC with the cap of 250. Going to add the convention bypass if cool. Have the broken up PR plan in an issue if you would rather that way!

protocol/src/http/pingora/handler/mod.rs
```diff
@@ -210,6 +210,32 @@ async fn logging_cleanup(pipeline: &FilterPipeline, ctx: &mut PingoraRequestCtx)
     }
 }
 
+/// Emit Prometheus metrics for a completed HTTP request.
+fn emit_request_metrics(session: &Session, ctx: &PingoraRequestCtx) {
```
shaneutt, line 217: If I'm not mistaken, we're running this function on every request, whether or not a metrics recorder has been installed? We should ensure this is a no-op when not needed.

protocol/src/http/pingora/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+
+// -----------------------------------------------------------------------------
+// Metric Recording
+// -----------------------------------------------------------------------------
+
+/// Labels for a completed HTTP request.
+///
+/// Static labels (`method`, `status_class`, `route`) use `&'static str`
+/// so the metrics facade can intern them without per-request allocation.
+/// Only `cluster` is dynamic.
```
shaneutt, line 125: Actually: I think we're allocating for every invocation. Look into using `Arc<str>` to eliminate per-request allocations, and see if we can optimize this a bit more.

protocol/src/http/pingora/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+    .increment(1);
+    histogram!(
+        HTTP_REQUEST_DURATION_SECONDS,
+        "method" => labels.method,
+        "status_class" => labels.status_class,
+        "route" => labels.route,
+        "cluster" => cluster
+    )
+    .record(duration_secs);
+}
```
shaneutt, line 156: This function creates multiple allocations on a hot path that appear to be unnecessary. For instance: we're doing `to_owned()` and then `clone()` on cluster, resulting in two allocations on every request just there. Take a pass at switching to borrows and Arcs.

protocol/src/http/pingora/handler/no_body.rs
```diff
@@ -172,8 +172,9 @@ impl ProxyHttp for PingoraHttpHandlerNoBody {
         upstream_peer::execute(ctx)
     }
 
-    async fn logging(&self, _session: &mut Session, e: Option<&pingora_core::Error>, ctx: &mut Self::CTX) {
+    async fn logging(&self, session: &mut Session, e: Option<&pingora_core::Error>, ctx: &mut Self::CTX) {
         let pipeline = self.pipeline.load();
+        emit_request_metrics(session, ctx);
         record_passive_health(&pipeline, e, ctx);
         logging_cleanup(&pipeline, ctx).await;
```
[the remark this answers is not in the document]
nerdalert: Same shared fix applies from the previous comment. `record_passive_health()` is used by both handlers, so the fallback covers both `with_body.rs` and `no_body.rs` without changing either logging hook.

protocol/src/http/pingora/handler/with_body.rs
```diff
@@ -217,8 +217,9 @@ impl ProxyHttp for PingoraHttpHandler {
         upstream_peer::execute(ctx)
     }
 
-    async fn logging(&self, _session: &mut Session, e: Option<&pingora_core::Error>, ctx: &mut Self::CTX) {
+    async fn logging(&self, session: &mut Session, e: Option<&pingora_core::Error>, ctx: &mut Self::CTX) {
         let pipeline = self.pipeline.load();
+        emit_request_metrics(session, ctx);
```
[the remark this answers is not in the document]
nerdalert: Fixed in `record_passive_health()` itself rather than in both logging hooks. The function now falls back to `ctx.metrics_cluster` when `ctx.cluster` is `None`: `ctx.cluster.as_ref().or(ctx.metrics_cluster.as_ref())`.

This centralizes the fix and avoids mutating `ctx.cluster` back to a value that was intentionally consumed. Added `passive_health_falls_back_to_metrics_cluster`, which proves passive health still records when `ctx.cluster` is `None` but `metrics_cluster` is set.

protocol/src/http/pingora/handler/mod.rs
```diff
@@ -210,6 +210,32 @@ async fn logging_cleanup(pipeline: &FilterPipeline, ctx: &mut PingoraRequestCtx)
...
+    let labels = metrics::RequestMetricLabels {
+        method,
+        status_class,
+        route: "unknown",
+        cluster: cluster_name,
+    };
+
+    let duration_secs = ctx.request_start.elapsed().as_secs_f64();
+    metrics::record_request_metrics(&labels, duration_secs);
+}
```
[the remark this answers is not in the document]
nerdalert: Added `metrics::is_recorder_installed()` and made it the first check in `emit_request_metrics()`.

When no admin listener is configured, no label computation, no timer read, no allocation happens.

protocol/src/http/pingora/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+    )
+    .increment(1);
+    histogram!(
+        HTTP_REQUEST_DURATION_SECONDS,
+        "method" => labels.method,
+        "status_class" => labels.status_class,
+        "route" => labels.route,
+        "cluster" => cluster
+    )
+    .record(duration_secs);
```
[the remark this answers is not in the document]
nerdalert: Changed `RequestMetricLabels.cluster` to `metrics::SharedString` and build it from the existing `ctx.metrics_cluster: Arc<str>` snapshot:

```rust
 let cluster = ctx.metrics_cluster.as_ref().map_or_else(
     || ::metrics::SharedString::const_str("none"),
     |cluster| ::metrics::SharedString::from(Arc::clone(cluster)),
 );
```

`record_request_metrics()` now consumes `RequestMetricLabels` by value, clones the `SharedString` for the counter, and moves the original into the histogram. For a real cluster label, that clone is a cheap shared-string / `Arc` clone rather than a `String` allocation. For the `"none"` path, the label is static and borrowed.

protocol/src/http/pingora/metrics.rs
nerdalert: Also fixed *I think*. cluster now uses metrics::SharedString built from the existing Arc<str> in ctx.metrics_cluster, or the static "none" label. record_request_metrics() clones the shared label once for the first metric and moves it into the second, so we no longer do to_owned() + String::clone() per request.

protocol/src/http/pingora/metrics.rs
nerdalert: I **think** this does it by changing cluster from String to metrics::SharedString, and now build it from the existing Arc<str> in ctx.metrics_cluster:

```
  ::metrics::SharedString::from(Arc::clone(cluster))
```

So it no longer does to_owned() + String::clone() on every recorded request and when metrics are disabled, emit_request_metrics() returns before doing any label work. Ty!

protocol/src/http/pingora/handler/mod.rs
nerdalert: Added the `is_recorder_installed()` guard as the first line of `emit_request_metrics()`. It is a single `OnceLock::get()` check. When the admin listener is not configured, the entire function is a no-op and does not compute labels or allocate. Ty!

shaneutt (approved):

PR #90, opened 2026-04-29
fix: add missing body mode fields.

usize:
This unbreaks fuzz tests that are missing required fields.


usize:
@shaneutt fuzz tests failing, seeing if this cleans it up.

shaneutt (approved):
Thank you! I'll get another patch in that will help make sure we catch this kind of thing.

Issue #17, opened 2026-04-03
Epic: External Processing


shaneutt:
External processing filter for request/response mutation via gRPC callout, compatible with the Envoy ext_proc protocol. Praxis acts as an ext_proc **client**, calling out to external processing services during request/response handling.

## Relationship to extproc/

The `extproc/` directory in the monorepo is the **server** side: Praxis acting as an ext_proc server for Envoy. This epic covers the **client** side: Praxis calling out to external processors from within its own filter pipeline.

## ext_proc Protocol Phases

Six processing phases, each independently configurable:

- `request_headers`: before forwarding request headers upstream
- `request_body`: before forwarding request body upstream
- `response_headers`: before forwarding response headers downstream
- `response_body`: before forwarding response body downstream
- `request_trailers`: before forwarding request trailers upstream
- `response_trailers`: before forwarding response trailers downstream

## Processing Modes

Per-phase processing mode configuration:

- **skip**: do not send this phase to the external processor
- **send**: stream data to processor as it arrives
- **buffered**: buffer entire content before sending to processor
- **buffered_partial**: buffer up to a configured limit, then send

Default: send headers, skip body.

## Mutation Types

Mutations returned by the external processor:

- **HeaderMutation**: add, remove, or overwrite request/response headers
- **BodyMutation**: replace request/response body content
- **ImmediateResponse**: short-circuit the pipeline, return a response directly to the client

## Filter Integration

- ext_proc implemented as an `HttpFilter` under `filter/src/builtins/http/payload_processing/`
- Delegates to external gRPC service at each configured phase
- Applies returned mutations to the request/response flowing through the pipeline

## Connection Management

- gRPC connection pooling to external processor endpoints
- Configurable keepalive interval and timeout
- Reconnection with backoff on connection failure
- Timeout cascading: connection timeout < request timeout < filter timeout

## Failure Handling

- Configurable fail-open (continue with unmodified request) vs fail-closed (reject request)
- Default: fail-closed (safe default)
- Timeout treated as failure, policy applies
- Per-filter failure mode configuration

## Related Issues

- #164 Extensibility Platform (v0.5.0) - parent category
- #133 Dynamic Filter Loading via Shared Libraries (v0.5.0) - related extensibility

shaneutt:
We have decided to _not_ generally support ext_proc as it goes against some of the general philosophies of Praxis, in addition to being a very large additional support and maintenance burden.

PR #94, opened 2026-04-29
feat: add praxis-proto crate with Envoy ext_proc protocol definitions

usize:
## Summary

Establishes support for gRPC via a tonic enabled library, and pulls in the Envoy ExternalProcessor (and related) protocols. This is wired into the filter crate via a feature, in anticipation of supporting external processors via gRPC.

Related: #17

## Details

**Proto crate (`proto/`):**
- Vendors `ext_proc.proto` (ExternalProcessor service, ProcessingRequest/Response, ProcessingMode, header/body/trailer mutations) and `envoy_common.proto` (HeaderValue, HeaderValueOption, StatusCode, HttpStatus, Metadata)
- Compiles with `protox` + `tonic-prost-build`, maps `google.protobuf.Value`/`Struct` to `prost-wkt-types`
- Generates both client and server stubs

**Feature wiring:**
- `praxis-filter` gains `ext-proc` feature with optional deps on `praxis-proto`, `tonic`, `prost-wkt-types`
- `praxis` (server) forwards the feature

## Test plan

- [x] `cargo build -p praxis-proxy-proto` builds standalone
- [x] `cargo build -p praxis-proxy-filter --features ext-proc` compiles with gRPC
- [x] `cargo build -p praxis-proxy-filter --no-default-features --features ai-inference` compiles without gRPC
- [x] `make build` full workspace passes
- [x] `make lint` clippy + fmt pass


filter/Cargo.toml
```diff
@@ -14,8 +14,9 @@ publish = true
 name = "praxis_filter"
 
 [features]
-default = ["ai-inference"]
+default = ["ai-inference", "ext-proc"]
```
shaneutt: We shouldn't add `ext-proc` as a default feature, it pulls in a lot of transitive dependencies and it seems very likely that the feature will only be used by a small subset of users. Let's make it an opt-in.

proto/build.rs
```diff
@@ -0,0 +1,30 @@
...
+    let cwd = std::env::current_dir()?;
+    let proto_files = ["proto/envoy_common.proto", "proto/ext_proc.proto"]
+        .iter()
+        .map(|name| cwd.join(name))
+        .collect::<Vec<_>>();
+    let include_dirs = [cwd.join("proto")];
+
+    let config = {
+        let mut c = prost_build::Config::new();
+        c.disable_comments(Some("."));
```
shaneutt: I believe this will strip ALL proto documentation, which can be very useful for developers (otherwise they have to cross-reference). We should consider removing this 🤔

proto/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+name = "praxis-proxy-proto"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+description = "Protobuf and gRPC definitions for Praxis external processing"
+license.workspace = true
+repository.workspace = true
+readme = "../README.md"
+publish = true
```
shaneutt: At least for the moment, let's make no suggestion that we're going to make this a public crate.

```suggestion
publish = false
```

Cargo.toml
```diff
@@ -53,6 +54,13 @@ pingora-core = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingo
 pingora-http = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingora.git", rev = "514d06b852e731848fd36574da9fcdaee101ccb4" }
 pingora-proxy = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingora.git", rev = "514d06b852e731848fd36574da9fcdaee101ccb4", features = ["rustls"] }
 percent-encoding = "2.3.2"
+praxis-proto = { version = "0.1.0", path = "proto", package = "praxis-proxy-proto" }
+prost = "0.14"
+prost-build = "0.14"
+prost-types = "0.14"
+prost-wkt-build = "0.7"
```
shaneutt: Our conventions call for full semver on every dependency. We need to update this for several deps here, and if there's some kind of linter than can catch this let's add that? 🤔

shaneutt (commented):
Thank you @usize!

A few comments, and copilot has some too.

At a high level: I feel weird about having `proto/` be at the root, given that it's specific to filters and it shouldn't be enabled in a standard build 🤔 Consider if there's a better place, maybe under filters?

proto/build.rs
usize: Fair enough. Let's leave it only revisit if it's a problem.

usize:
> Thank you @usize!
> 
> A few comments, and copilot has some too.
> 
> At a high level: I feel weird about having `proto/` be at the root, given that it's specific to filters and it shouldn't be enabled in a standard build 🤔 Consider if there's a better place, maybe under filters?

I spent a bit thinking about where to put it as well. Putting it at the root felt natural to me because I thought of bringing in gRPC support as its own "thing". But it is true that it's bound up with filters. So Having it live there makes sense. I'll move it under filters.

filter/proto/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+//! This crate compiles vendored `.proto` files from the Envoy project into
+//! Rust types with [`tonic`] gRPC client and server stubs.
+
+#![deny(unsafe_code)]
+
+pub mod envoy {
+    pub mod service {
+        pub mod common {
+            #[allow(
+                warnings,
```
shaneutt, line 12: Allowing `warnings` ends up covering many of the checks below. We should simply stick to allowing the warnings we know we need to allow to make things more clear, and make it harder for something unexpected to fall through.

filter/proto/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+                missing_docs,
+                unreachable_pub,
+                trivial_casts,
+                unused_qualifications,
+                clippy::all,
+                clippy::pedantic
```
shaneutt, line 18: Are we sure we need all of these exclusions?

filter/proto/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+            )]
+            pub mod v3 {
+                tonic::include_proto!("envoy.service.common.v3");
+            }
+        }
+
+        pub mod ext_proc {
+            #[allow(
+                warnings,
```
shaneutt, line 27:
```suggestion
```

filter/proto/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+                missing_docs,
+                unreachable_pub,
+                trivial_casts,
+                unused_qualifications,
+                clippy::all,
+                clippy::pedantic
```
shaneutt, line 33: ?

```suggestion
```

filter/proto/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+            )]
+            pub mod v3 {
+                tonic::include_proto!("envoy.service.common.v3");
+            }
+        }
+
+        pub mod ext_proc {
+            #[allow(
+                warnings,
+                missing_docs,
```
shaneutt, line 28:
```suggestion
```

filter/proto/build.rs
```diff
@@ -0,0 +1,45 @@
...
+    let config = {
+        let mut c = prost_build::Config::new();
+        c.extern_path(".google.protobuf.Value", "::prost_wkt_types::Value");
+        c.extern_path(".google.protobuf.Struct", "::prost_wkt_types::Struct");
+        c
+    };
+
+    let fds = protox::compile(&proto_files, &include_dirs)?;
+    tonic_prost_build::configure()
+        .build_server(true)
```
shaneutt, line 18: We should have .build_client(true) added explicitly for clarity.

filter/proto/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[lib]
+name = "praxis_proto"
+
+# Workspace lints are not inherited: this crate wraps generated protobuf code
+# that cannot satisfy the strict style rules applied to hand-written code.
+# Lint suppression is handled by crate-level `#![allow(...)]` attributes in
+# lib.rs instead.
+[lints.rust]
+unsafe_code = "deny"
```
shaneutt, line 23: I don't know that we want to do this. It seems reasonable to us to keep checking `dead_code`, `ununsed_imports`, `unused variables`, etc as examples?

filter/proto/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[dependencies]
+prost = { workspace = true }
+prost-types = { workspace = true }
+prost-wkt-types = { workspace = true }
+tonic = { workspace = true }
+tonic-prost = { workspace = true }
```
shaneutt, line 30: Double check: do we actually need this at _runtime_? 🤔 _

Cargo.toml
```diff
@@ -53,6 +54,13 @@ pingora-core = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingo
 pingora-http = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingora.git", rev = "514d06b852e731848fd36574da9fcdaee101ccb4" }
 pingora-proxy = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingora.git", rev = "514d06b852e731848fd36574da9fcdaee101ccb4", features = ["rustls"] }
 percent-encoding = "2.3.2"
+praxis-proto = { version = "0.1.0", path = "filter/proto", package = "praxis-proxy-proto" }
+prost = "0.14.3"
+prost-build = "0.14.3"
+prost-types = "0.14.3"
+prost-wkt-build = "0.7.1"
+prost-wkt-types = { version = "0.7.1", features = ["vendored-protox"] }
```
shaneutt, line 62: Do we need the vendored option? 🤔

It seems like currently we'll end up with two copies compiled internally.

shaneutt (changes requested):

shaneutt (approved):
Approving as there's nothing that really needs to hold us up.

usize:
I see we landed this. I'm going to fix up the requested changes in the follow up here. Where we start to actually build the ext_proc support out. Thanks for the patience and reviews!

PR #169, opened 2026-05-04
chore: add an xtask to detect missing minor versions in dependencies.

usize:
In some cases it may be desirable/necessary to leave out the minor version. In those cases, a comment with the string `semver:ignore` either above the dependency or inline with it will disable the check.

This check has also been added to our github ci workflows.

Implementation was assisted by Opus 4.6.


usize:
Reminder here, that I think we should not count tests toward the line limit in our incoming PR checks.

PR #178, opened 2026-05-07
feat: add --dump / -T CLI flag for effective config output

### LINKED ISSUE #131: Config Validation and Dump Mode
Validate configuration files without starting the proxy, and dump the effective configuration for debugging.

Scope:
- `--validate` / `-t` flag: parse and validate config, exit with success/failure
- `--dump` / `-T` flag: validate and print the effective configuration (resolved chains, defaults applied)
- Exit codes: 0 for valid, non-zero for invalid with error details on stderr

nerdalert:
## Summary

2nd half of #131. This is the `--dump` functionality. It grew a bit more than I planned as it moves some `--validate` functions out of main.rs to avoid growing the binary.

- Moved load_and_validate_for_cli, validate_config_for_startup, run_dump, and resolve_config_source from main.rs into commands.rs so main.rs stays a thin CLI dispatcher and the check-mode behavior is testable without growing the binary entry point.
- Add `praxis --dump` / `-T` as a non-serving CLI mode.
- Emit YAML with the config source, effective parsed configuration with defaults, and resolved top-level listener chains.
- Add serialization support for config types needed by the effective config dump.
- `--dump` and `--validate` are mutually exclusive; both exit 0 on success, non-zero with errors to stderr on failure; neither starts the server

  ## Testing

  - `cargo check -p praxis`
  - `cargo test -p praxis`
  - `cargo clippy -p praxis --all-targets -- -D warnings`

Example output can be found here: https://gist.github.com/nerdalert/0a2028d9581bd17ea8b5463f56e85870

### DIFF
--- a/server/src/commands.rs
+++ b/server/src/commands.rs
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Check-mode command helpers shared by `--validate` and `--dump`.
+
+use praxis_core::config::Config;
+
+use crate::dump;
+
+// -----------------------------------------------------------------------------
+// Shared Validation
+// -----------------------------------------------------------------------------
+
+/// Load and fully validate configuration without starting the server.
+///
+/// Shared by `--validate` and `--dump`. Runs the same validation
+/// checks used during server startup: log override validation,
+/// filter factory instantiation, chain expansion, ordering checks,
+/// and body-limit application.
+///
+/// # Errors
+///
+/// Returns an error if loading or validation fails.
+pub(crate) fn load_and_validate_for_cli(
+    explicit: Option<&str>,
+) -> Result<Config, Box<dyn std::error::Error + Send + Sync>> {
+    let config = praxis::load_config(explicit)?;
+    validate_config_for_startup(&config)?;
+    Ok(config)
+}
+
+/// Validate a parsed configuration by building filter pipelines.
+pub(crate) fn validate_config_for_startup(config: &Config) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    praxis_core::logging::validate_log_overrides(config)?;
+    let registry = praxis_filter::FilterRegistry::with_builtins();
+    let health_registry = praxis_core::health::build_health_registry(&config.clusters);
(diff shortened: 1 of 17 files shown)


server/src/commands.rs
```diff
@@ -0,0 +1,130 @@
...
+    let dump_model = dump::build_dump(&config, &source)?;
+    dump::write_dump(&dump_model)
+}
+
+/// Determine the human-readable config source label.
+fn resolve_config_source(explicit: Option<&str>) -> String {
+    if let Some(path) = explicit {
+        return path.to_owned();
+    }
+    if let Ok(env_path) = std::env::var("PRAXIS_CONFIG") {
```
shaneutt: Practically speaking I don't think this branch is reachable because `main.rs` already does `cli.config_or_else(|| std::env::var("Praxis_Config").ok())`. While we're here, it seems like it would be worth it to refactor this function to no longer take an `Option`.

server/src/dump.rs
```diff
@@ -0,0 +1,309 @@
...
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used, clippy::expect_used, clippy::indexing_slicing, reason = "tests")]
+mod tests {
+    use praxis_core::config::{Config, FailureMode};
+
+    use super::*;
+
+    #[test]
+    fn dump_defaults_appear_under_configuration() {
```
shaneutt, line 219: We should to add a test that round-trips a representative config (with conditions, branch chains, multiple filter types, etc) through `serde_yaml::to_string` and back through `Config::from_yaml`. This would be a good way to look out for any serialization regressions going forward.

server/src/commands.rs
nerdalert: Thanks!! 😵‍💫 - Removed the unreachable env-var branch and split the implicit source handling out so we only pass the already-resolved explicit source through this path.

server/src/dump.rs
nerdalert: Nice, ty - test now serializes a representative config with request conditions, response conditions, branch chains, inline chains, and multiple filter types, then parses it back through `Config::from_yaml`. It also checks that conditions serialize as `when:` / `unless:` mappings instead of YAML tags.

shaneutt (approved):

Issue #179, opened 2026-05-07
request_body_filter not called when downstream body consumed before forwarding


nerdalert:
## Problem

This is lengthy but it's not an obvious solution and will need to manifest into an upstream fix so worth some optionality. For the unblocking of dev I'm going with option 1 with our fork to get unblocked for agentic > 64KiB sessions.

When a `ProxyHttp` implementation reads the downstream request body before upstream forwarding begins, the downstream body can be fully consumed by the time Pingora enters the proxy forwarding loop.

This can happen when an implementation needs to inspect the request body before choosing an upstream, for example body-based routing, JSON field extraction, request validation, or request normalization.

In that state:

- `is_body_done()` returns `true`
- `is_body_empty()` returns `false` because the request did have a body
- `get_retry_buffer()` returns `None` unless retry buffering was explicitly enabled

The forwarding loop currently only enters the initial body-send path when it has a retry buffer, or in the h1 path when the body is empty:

```rust
if buffer.is_some() || session.as_mut().is_body_empty() {
    send_body_to_pipe(...)
}
```

For a non-empty body that was already consumed before forwarding, both conditions are false. Pingora skips the initial body-send path and then enters the normal duplex loop. The downstream side is already done, so `read_body_or_idle(done = true)` idles instead of producing another body chunk. The upstream waits for the request body, the client waits for a response, and the request can hang until timeout.

```text
Downstream: body already consumed, is_body_done() = true
Upstream:   waiting for request body
Proxy:      idle(), waiting for downstream close
Result:     request stalls until timeout
```

The missing behavior is a terminal `request_body_filter()` callback after a non-empty body has already been consumed by the implementation. Without that callback, the implementation has no chance to emit a transformed or replayed body to upstream.

## Impact

Any `ProxyHttp` implementation that pre-reads a request body and expects to provide the upstream body through `request_body_filter()` can stall on non-empty requests.

The current workaround is to call `enable_retry_buffering()` before the pre-read and then rely on `get_retry_buffer()` to trigger the initial body-send path. That workaround is not ideal:

- Retry buffering is meant for upstream retry/reuse decisions, not pre-read body replay.
- The retry buffer is capped by `BODY_BUF_LIMIT` at 64 KiB.
- Bodies larger than the retry buffer limit can be truncated or become unavailable for retry-buffer replay.
- Increasing the retry buffer limit does not fix the lifecycle issue; it only stretches a workaround.

[...]

Issue #75, opened 2026-04-24
StreamBuffer pre-read body forwarding limited to 64 KiB


usize:
## Problem

The `StreamBuffer` pre-read in `stream_buffer.rs` consumes the downstream request body via `read_request_body()` so that body filters can promote headers before the request-phase pipeline runs (e.g. `json_body_field` extracting a model name).

After the pre-read, Pingora's body forwarding loop never calls `request_body_filter` because `is_body_done()` returns `true`. The pre-read body stored in `ctx.pre_read_body` is never drained to upstream.

## Current Workaround
In https://github.com/praxis-proxy/praxis/pull/62 I added  `enable_retry_buffering()`  before the pre-read so that
Pingora's retry buffer captures the body data. 

Pingora then uses `get_retry_buffer()` to replay the body to upstream via `send_body_to_pipe`.

However, **Pingora's retry buffer is capped at 64 KiB** via (`BODY_BUF_LIMIT` in `pingora-core`). 

Bodies exceeding that limit are silently truncated and will not be forwarded to upstream.

## Proposed Fix

Add an upstream Pingora API (e.g. `set_preread_body()`) that feeds Pingora's body forwarding loop without the 64 KiB retry buffer limit. Or, if we still need to enable retry buffering, ensure that we have a way to up the limit. @shaneutt mentioned that this can be resolved upstream.

## Impact

Any `StreamBuffer` filter that processes request bodies larger than 64 KiB will silently drop the body when forwarding to upstream. Currently only `json_body_field` uses `StreamBuffer`, and its integration tests use `start_header_echo_backend` which does not read the request body, masking the issue.

PR #180, opened 2026-05-07
feat: strip reserved internal headers to prevent client spoofing

### LINKED ISSUE #148: Epic: AI Agentic Protocol Support
Protocol support for AI agent communication patterns: MCP (Model Context Protocol) and A2A (Agent-to-Agent).

## Status

The initial protocol foundation and useful production paths are implemented. The remaining work is primarily sessionful MCP backend routing, shared/distributed state, and explicitly scoped advanced/security integrations.

This epic was originally written as a nine-PR implementation plan. That plan is now historical: the filters moved from Praxis core into this repository, the MCP stateless profile was added, and several original tracking issues were completed or superseded. The sections below describe the current state.

## Shipped

- [x] Shared JSON-RPC parsing and durable filter metadata (praxis-proxy/ai#152, praxis-proxy/praxis#101)
- [x] Reserved internal-header hygiene (praxis-proxy/praxis#180)
- [x] Reusable MCP and A2A integration-test servers (praxis-proxy/praxis#369)
- [x] MCP classification: method, name, protocol version, session presence, and header/body validation (praxis-proxy/ai#154)
- [x] Static MCP catalog aggregation and namespace prefixes
- [x] Stateless MCP `tools/call` routing, prefix removal, backend path rewrite, and framing repair (praxis-proxy/ai#247, praxis-proxy/ai#340)
- [x] A2A classification, task/context extraction, local ownership routing, and SSE response capture (praxis-proxy/ai#157, praxis-proxy/ai#226)
- [x] StreamBuffer body-mutation adapter behavior (praxis-proxy/ai#212)
- [x] Broker-to-load-balancer pipeline validation without a dummy router (praxis-proxy/ai#213)

## Operating modes on `main`

| Mode | Pipeline | Current behavior |
| --- | --- | --- |
| Pass-through | `router -> load_balancer` | Proxies MCP/A2A HTTP traffic without protocol interpretation. |
| Classifier | `mcp` or `a2a` -> `router` -> `load_balancer` | Promotes body-derived protocol metadata for normal routing. |
| Stateless MCP broker | `mcp` -> `load_balancer` | Serves discovery/catalog operations and routes `tools/call` through the configured static catalog. |
| Current/sessionful MCP broker | `mcp` -> `load_balancer` | Handles local initialize/catalog operations, but backend `tools/call` remains unsupported until backend-session infrastructure lands. |
| A2A ownership routing | `a2a` -> `router` -> `load_balancer` | Captures task/context ownership and routes follow-up traffic using the local store. |

The old `mcp_gateway` filter name is no longer the public configuration shape; broker behavior is configured on the `mcp` filter through `servers` and `protocol_profile`.

## Remaining core work

- [ ] Complete the bounded, reusable MCP backend callout and lazy initialization boundary (praxis-proxy/ai#179).
- [ ] Complete current-profile backend session lifecycle and settle the shared/distributed state design (praxis-proxy/ai#156).
- [ ] Reconcile the remaining historical gateway tracker with the shipped stateless broker and the sessionful gap (praxis-proxy/ai#155).
- [ ] Reconcile the A2A task-routing tracker: local task/context routing is shipped; shared/distributed storage remains a separate concern (praxis-proxy/ai#158).
- [ ] Reconcile the remaining safety tracker with the header-hygiene and mock-server work that already shipped (praxis-proxy/ai#153).

## Advanced, security, and integration follow-ups

These are related workstreams, but should not silently expand the definition of the initial protocol-support milestone:

- MCP SSE aggregation and replay: praxis-proxy/ai#143
- MCP authorization policy: praxis-proxy/ai#144
- MCP notifications, elicitation, and conformance: praxis-proxy/ai#159
- Anthropic Messages MCP toolsets: praxis-proxy/ai#638
- Tool argument schema enforcement: praxis-proxy/ai#1125
- Responses agentic-loop authorization propagation: praxis-proxy/ai#1128

## Completion boundary

Close this epic when:

1. Current-profile MCP calls can initialize and reuse backend sessions with bounded I/O, or that profile is explicitly deferred and documented.
2. The local versus distributed state contract for MCP and A2A has an explicit implemented or deferred outcome.
3. Historical trackers (`#153`, `#155`, `#156`, `#158`, `#159`) are closed, narrowed, or moved to follow-up scope so their state matches the code.
4. Advanced/provider-specific issues are either accepted as part of this epic or reparented to their owning feature epics.

Last reconciled with `main`: 2026-09-24.

nerdalert:
The full _AI Agentic Protocol Support_ plan and PR/feature orders are in praxis-proxy/ai#148. This is PR-2 in the planned stack. The `BODY_BUF_LIMIT` test Is currently ignored until the Pingora workaround PR lands: https://github.com/praxis-proxy/pingora/pull/1 and issue: praxis-proxy/praxis#179 

  ## Summary

  - Strip client-supplied `x-praxis-*`, `x-mcp-*`, and `x-a2a-*` headers from incoming requests before the filter pipeline runs, preventing clients from spoofing
  body-derived routing decisions.
  - Strip the same reserved prefixes from upstream requests before forwarding to backends, preventing internal routing metadata from leaking.
  - Standard MCP protocol headers (`MCP-Session-Id`, `Mcp-Method`, `Mcp-Name`, `MCP-Protocol-Version`) are preserved — they lack the `x-` prefix and are not part of
  the reserved set.
  - Add a >64 KiB StreamBuffer body regression test that documents the Pingora `BODY_BUF_LIMIT` cap (#75). 

  ## Addresses

 MCP and A2A body-aware filters (planned in later PRs) will parse JSON-RPC request bodies and promote internal headers like `x-praxis-mcp-method: tools/call` so the `router` filter can route by them. Without this PR, a client could send `X-Praxis-Mcp-Method: initialize` directly and the router would trust it over the body-derived value. This is the same class of problem that Kuadrant MCP Gateway (`x-mcp-servername`) and Envoy AI Gateway (`x-ai-eg-mcp-backend`) face with their  internal headers.

  ## What changed

  | File | Change |
  | --- | --- |
  | `protocol/.../request_filter/mod.rs` | `strip_reserved_internal_headers()` — runs after normalization, before request snapshot |
  | `protocol/.../upstream_request.rs` | `strip_reserved_internal()` — runs after hop-by-hop stripping, before upstream forwarding |
  | `protocol/.../with_body.rs`, `no_body.rs` | Wire upstream stripping into both handler paths |
  | `tests/.../security.rs` | 6 new tests: 2 acceptance (spoofed header vs body routing, spoofed header not reaching backend), 4 prefix (x-praxis-*, x-mcp-*, x-a2a-*
  stripped; standard MCP headers preserved) |
  | `tests/.../body.rs` | 1 new test: >64 KiB StreamBuffer regression (currently failing, blocked on Pingora fix) |

  ## Test plan

  - [x] `spoofed_internal_header_overridden_by_body_derived_routing` — client sends `X-Praxis-Mcp-Method: initialize` with body `method: tools/call`, router picks
  `tools-call` backend
  - [x] `spoofed_internal_header_does_not_reach_backend` — header-echo backend confirms no `x-praxis-mcp-method` forwarded
  - [x] `reserved_x_praxis_headers_stripped_from_client` — `x-praxis-mcp-method`, `x-praxis-mcp-name`, `x-praxis-a2a-method` stripped
  - [x] `reserved_x_mcp_headers_stripped_from_client` — `x-mcp-servername`, `x-mcp-toolname` stripped
  - [x] `reserved_x_a2a_headers_stripped_from_client` — `x-a2a-method`, `x-a2a-family` stripped
  - [x] `standard_mcp_protocol_headers_preserved` — `MCP-Session-Id`, `Mcp-Method`, `Mcp-Name`, `MCP-Protocol-Version` forwarded to backend
  - [x] `stream_buffer_body_above_64kib_forwarded_intact` — set to ignore until Pingora fork PR merges
  - [x] 378 total integration tests pass (1 expected failure above), 0 clippy errors

### DIFF
--- a/protocol/src/http/pingora/handler/reserved_headers.rs
+++ b/protocol/src/http/pingora/handler/reserved_headers.rs
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Reserved internal header helpers for proxy-owned routing metadata.
+
+/// Built-in reserved header prefixes for Praxis agentic protocol routing.
+///
+/// Headers with these prefixes are proxy-internal metadata used for
+/// body-derived routing decisions. Clients must not be able to inject
+/// them directly, and they should not be forwarded to upstream backends.
+///
+/// Standard MCP protocol headers (`mcp-session-id`, `mcp-method`,
+/// `mcp-name`, `mcp-protocol-version`, `mcp-param-*`) do NOT match these
+/// prefixes because they lack the `x-` prefix.
+///
+/// TODO(#186) Spike: consider additive operator-managed reserved prefixes
+/// once the broader config model defines global vs listener/filter-chain
+/// scope and additive vs override semantics.
+const RESERVED_HEADER_PREFIXES: &[&str] = &["x-praxis-", "x-mcp-", "x-a2a-"];
+
+/// Return whether a header name belongs to Praxis reserved internal metadata.
+pub(in crate::http::pingora::handler) fn is_reserved_internal_header(name: &http::HeaderName) -> bool {
+    let name = name.as_str();
+    RESERVED_HEADER_PREFIXES.iter().any(|prefix| name.starts_with(prefix))
+}
--- a/tests/integration/tests/suite/body.rs
+++ b/tests/integration/tests/suite/body.rs
@@ -695,3 +695,67 @@ filter_chains:
 "#
     )
 }
+
+// -----------------------------------------------------------------------------
+// StreamBuffer Large-Body Regression (#75)
+// -----------------------------------------------------------------------------
+
+#[test]
+fn stream_buffer_body_above_64kib_forwarded_intact() {
+    let backend_guard = start_echo_backend_with_shutdown();
+    let proxy_port = free_port();
+
+    let yaml = format!(
+        r#"
+listeners:
+  - name: default
+    address: "127.0.0.1:{proxy_port}"
+    filter_chains: [main]
+filter_chains:
+  - name: main
+    filters:
+      - filter: json_rpc
+        max_body_bytes: 131072
+      - filter: router
+        routes:
+          - path_prefix: "/"
+            cluster: "backend"
+      - filter: load_balancer
+        clusters:
+          - name: "backend"
+            endpoints:
+              - "127.0.0.1:{}"
+"#,
+        backend_guard.port()
+    );
+    let config = Config::from_yaml(&yaml).unwrap();
+    let proxy = start_proxy(&config);
+
+    let payload = format!(
+        r#"{{"jsonrpc":"2.0","id":1,"method":"test","params":{{"data":"{}"}}}}"#,
+        "x".repeat(70_000)
+    );
+    let request = format!(
+        "POST / HTTP/1.1\r\n\
+         Host: localhost\r\n\
+         Content-Type: application/json\r\n\
+         Content-Length: {}\r\n\
+         \r\n\
+         {payload}",
+        payload.len(),
+    );
+
+    let raw = http_send(proxy.addr(), &request);
+    let status = parse_status(&raw);
+    assert_eq!(status, 200, "request should succeed");
+
+    let echoed = praxis_test_utils::parse_body(&raw);
+    assert_eq!(
+        echoed.len(),
+        payload.len(),
+        "backend should receive the full body ({} bytes), but got {} bytes — \
+         Pingora retry buffer truncates at 64 KiB (BODY_BUF_LIMIT)",
+        payload.len(),
+        echoed.len()
+    );
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1042,6 +1042,8 @@ version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
 dependencies = [
+ "allocator-api2",
+ "equivalent",
  "foldhash 0.1.5",
 ]
 
@@ -1061,11 +1063,6 @@ name = "hashbrown"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4f467dd6dccf739c208452f8014c75c18bb8301b050ad1cfb27153803edb0f51"
-dependencies = [
- "allocator-api2",
- "equivalent",
- "foldhash 0.2.0",
-]
 
 [[package]]
 name = "heck"
@@ -1854,7 +1851,7 @@ checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 [[package]]
 name = "pingora-cache"
 version = "0.8.0"
-source = "git+https://github.com/praxis-proxy/pingora.git?rev=2e452af034fb23a9cd57ec808e7d374037c0d664#2e452af034fb23a9cd57ec808e7d374037c0d664"
+source = "git+https://github.com/praxis-proxy/pingora.git?rev=d066f6818038dc296326566afb8e7bf39fb463ff#d066f6818038dc296326566afb8e7bf39fb463ff"
 dependencies = [
  "ahash",
  "async-trait",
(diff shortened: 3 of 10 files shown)


twghu:
RESERVED_HEADER_PREFIXES appears twice in 2 separate files.  

Rather than hard code the strings a configuration option would be more flexible.  Any new or unexpected string requires a build/release cycle. 

Likely this would be managed by an operator at some point so the config could like something like:
```
  # adds to the built-in x-praxis-*, x-mcp-*, x-a2a-* set                                                                               
  reserved_header_prefixes:                                                                                                             
    - x-mycompany-
    - x-secret-squirrel-
```
This makes support in production easier if something arises that was unexpected.   

Maybe mark this as #todo for now until the full config picture is defined?

twghu:
Possible existing test assertion flipped without renaming — 
the InjectInternalFilter test in security.rs had its assertion reversed:
 it  previously proved that filter-injected x-praxis-secret reached upstream; it now proves the opposite. The test name and description no  longer match what it's testing. It should be split or renamed to make the intent clear.

nerdalert:
> RESERVED_HEADER_PREFIXES appears twice in 2 separate files.
> 
> Rather than hard code the strings a configuration option would be more flexible. Any new or unexpected string requires a build/release cycle.
> 
> Likely this would be managed by an operator at some point so the config could like something like:
> 
> ```
>   # adds to the built-in x-praxis-*, x-mcp-*, x-a2a-* set                                                                               
>   reserved_header_prefixes:                                                                                                             
>     - x-mycompany-
>     - x-secret-squirrel-
> ```
> 
> This makes support in production easier if something arises that was unexpected.
> 
> Maybe mark this as #todo for now until the full config picture is defined?

SGTM, duped and added a todo. Cheers.

protocol/src/http/pingora/handler/reserved_headers.rs
```diff
@@ -0,0 +1,25 @@
...
+///
+/// Headers with these prefixes are proxy-internal metadata used for
+/// body-derived routing decisions. Clients must not be able to inject
+/// them directly, and they should not be forwarded to upstream backends.
+///
+/// Standard MCP protocol headers (`mcp-session-id`, `mcp-method`,
+/// `mcp-name`, `mcp-protocol-version`, `mcp-param-*`) do NOT match these
+/// prefixes because they lack the `x-` prefix.
+///
+/// TODO(agentic): consider additive operator-managed reserved prefixes
```
shaneutt: Let's add an issue (maybe a spike) to follow up on this, and link it here so this doesn't become an errant TODO.

tests/integration/tests/suite/body.rs
```diff
@@ -695,3 +695,68 @@ filter_chains:
 "#
     )
 }
+
+// -----------------------------------------------------------------------------
+// StreamBuffer Large-Body Regression (#75)
+// -----------------------------------------------------------------------------
+
+#[test]
+#[ignore = "manual regression for Praxis issue #179; bodies >64 KiB through StreamBuffer may fail or not replay intact"]
```
shaneutt: Do we have what we need now, can we update our dependency?

shaneutt (changes requested):
Thanks @nerdalert! A couple of small comments.

Because of the tone we're trying to set in the early days of Praxis (e.g. strict, fail-fast) we should take a more austere approach to this change:

Let's assume that any time one of these headers would come in, it would have to be either:

a) malicious
b) a programming mistake from the client

By default, let's _error_ on requests that try to spoof these headers instead of just stripping them. We can add an insecure option to strip them also, but then we should warn. LMKWYT?

protocol/src/http/pingora/handler/reserved_headers.rs
nerdalert: Done! Opened spike #186

tests/integration/tests/suite/body.rs
nerdalert: Updated the Pingora rev to current praxis-proxy/pingora main. It passed all of the IT along with the uncommented test here. 🤞

nerdalert:
> Thanks @nerdalert! A couple of small comments.
> 
> Because of the tone we're trying to set in the early days of Praxis (e.g. strict, fail-fast) we should take a more austere approach to this change:
> 
> Let's assume that any time one of these headers would come in, it would have to be either:
> 
> a) malicious b) a programming mistake from the client
> 
> By default, let's _error_ on requests that try to spoof these headers instead of just stripping them. We can add an insecure option to strip them also, but then we should warn. LMKWYT?

Awesome, here are the changes:


Client-supplied reserved internal headers now fail fast with 400 before Max-Forwards handling or filter execution. 
Filter-generated reserved headers are still allowed internally for routing, then stripped before upstream forwarding. 
Integration coverage was added for spoof rejection, body-derived routing, generated-header stripping, and all reserved prefix families.

shaneutt (approved):

PR #182, opened 2026-05-10
feat: add ext_proc filter crate with Envoy-compatible config

usize:
Introduce `praxis-proxy-ext-proc` as a standalone workspace crate at filter/ext-proc/, alongside filter/proto/. Isolated from the filter builtins to contain the gRPC/tonic/proto dependency surface.

The config mirrors Envoy's ExternalProcessor proto shape for easy migration from Envoy deployments:

- target, failure_mode, message_timeout_ms
- processing_mode (header/body/trailer send modes)
- allow_mode_override, observability_mode, disable_immediate_response, max_message_timeout_ms

Features not yet implemented are parsed but rejected during validation with a clear error. The filter is not registered in FilterRegistry::with_builtins(); consumers register it explicitly via registry.register() or register_filters!.

Opus 4.6 aided in codebase exploration and feature implementation.


usize:
Experimenting with saving summarized agent session for prs: https://gist.github.com/usize/ebb03b9963383cf01f433bd5bf7719e3

usize:
@shaneutt Aside from config parsing there's not much happening here. Except it does establish the structure of the `filter/ext-proc` crate. I wanted to check your thoughts on the structure.

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+    reason = "mirrors Envoy ExternalProcessor proto fields"
+)]
+struct ExtProcConfig {
+    /// gRPC endpoint URI of the external processing server.
+    target: String,
+
+    /// Behaviour when the external processor is unreachable or returns
+    /// an error. Maps to Envoy's `failure_mode_allow`.
+    #[serde(default)]
+    failure_mode: FailureMode,
```
shaneutt: `failure_mode` may be getting stripped in `strip_structural_keys`, meaning it will always deserialize as the serde default even if someone explicit sets open.

The tests don't cover this behavior, we need to expand them.

We should remove `failure_mode` here anyway, and instead read it from the `FilterEntry` at pipeline build.

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+    /// Whether the external processor may override the processing
+    /// mode via `mode_override` in its responses.
+    #[serde(default)]
+    allow_mode_override: bool,
+
+    /// Upper bound in milliseconds for `override_message_timeout`
+    /// values sent by the external processor. When set, the server
+    /// may extend the per-message timeout up to this limit.
+    #[allow(dead_code, reason = "parsed for config compatibility; used in subsequent PRs")]
+    max_message_timeout_ms: Option<u64>,
```
shaneutt, line 132: Missing validation: value `0` and a value less than `message_timeout_ms` would be accepted (silently), when you do the alluded to follow-up PRs, just make sure we add validation.

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+    /// Send body to the processor as it arrives without waiting for
+    /// the header response. Only applies to `streamed` body mode.
+    #[serde(default)]
+    send_body_without_waiting_for_header_response: bool,
+
+    /// Allowlist of processing modes the processor may override to.
+    /// Only evaluated when `allow_mode_override` is `true`.
+    #[allow(dead_code, reason = "parsed for config compatibility; used in subsequent PRs")]
+    #[serde(default)]
+    allowed_override_modes: Vec<ProcessingModeConfig>,
```
shaneutt, line 177: There seems to be no test for when this is empty

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+    }
+    Ok(())
+}
+
+/// Reject unsupported [`ProcessingModeConfig`] values.
+fn validate_processing_mode(pm: &ProcessingModeConfig) -> Result<(), FilterError> {
+    if pm.request_header_mode == HeaderSendMode::Skip {
+        return Err("ext_proc: request_header_mode 'skip' is not yet supported".into());
+    }
+    if pm.response_header_mode == HeaderSendMode::Skip {
```
shaneutt, line 424: technically no test covering this path

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+        return Err(format!(
+            "ext_proc: response_body_mode '{}' is not yet supported (only 'none')",
+            pm.response_body_mode,
+        )
+        .into());
+    }
+    if pm.request_trailer_mode == HeaderSendMode::Send {
+        return Err("ext_proc: request_trailer_mode 'send' is not yet supported".into());
+    }
+    if pm.response_trailer_mode == HeaderSendMode::Send {
```
shaneutt, line 438: technically no test covering this path

filter/ext-proc/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+name = "praxis_ext_proc"
+
+[lints]
+workspace = true
+
+[dependencies]
+async-trait = { workspace = true }
+praxis-core = { workspace = true }
+praxis-filter = { workspace = true }
+praxis-proto = { workspace = true }
```
shaneutt: Are we using this anywhere in here? 🤔

Cargo.toml
```diff
@@ -56,6 +57,7 @@ pingora-core = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingo
 pingora-http = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingora.git", rev = "d066f6818038dc296326566afb8e7bf39fb463ff" }
 pingora-proxy = { version = "0.8.0", git = "https://github.com/praxis-proxy/pingora.git", rev = "d066f6818038dc296326566afb8e7bf39fb463ff", features = ["rustls"] }
 percent-encoding = "2.3.2"
+praxis-ext-proc = { version = "0.2.0", path = "filter/ext-proc", package = "praxis-proxy-ext-proc" }
```
shaneutt: Technically should be up to date with workspace version

```suggestion
praxis-ext-proc = { version = "0.3.1", path = "filter/ext-proc", package = "praxis-proxy-ext-proc" }
```

shaneutt (changes requested):

filter/ext-proc/Cargo.toml
usize: dropped praxis-core, but others are indeed needed.

filter/ext-proc/src/lib.rs
usize: fixed up.

filter/ext-proc/src/lib.rs
usize: Covered under the default feature flags test. Could be more explicit, but may be overkill.

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+fn validate_processing_mode(pm: &ProcessingModeConfig) -> Result<(), FilterError> {
+    if pm.request_header_mode == HeaderSendMode::Skip {
+        return Err("ext_proc: request_header_mode 'skip' is not yet supported".into());
+    }
+    if pm.response_header_mode == HeaderSendMode::Skip {
+        return Err("ext_proc: response_header_mode 'skip' is not yet supported".into());
+    }
+    if pm.request_body_mode != BodySendMode::None {
+        return Err(format!(
+            "ext_proc: request_body_mode '{}' is not yet supported (only 'none')",
```
twghu: Convention nit:
  
// Old style — positional
  format!("invalid mode '{}'", pm.request_body_mode)

  // Inline format args — what the convention requires
  let mode = pm.request_body_mode;
  format!("invalid mode '{mode}'")

As format args don't allow for field access expressions this is probably super nit pick

Similar patternss on lines 419–421, 425–427, 499

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+    /// Returns [`FilterError`] if the YAML config is malformed, the
+    /// target URI is invalid, or an unsupported feature is requested.
+    pub fn from_config(config: &serde_yaml::Value) -> Result<Box<dyn HttpFilter>, FilterError> {
+        let cfg: ExtProcConfig = parse_filter_config("ext_proc", config)?;
+        validate_config(&cfg)?;
+
+        let endpoint: Endpoint = cfg
+            .target
+            .parse()
+            .map_err(|e: tonic::transport::Error| -> FilterError {
```
twghu: maybe just |e| ?

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+/// target: "http://127.0.0.1:50051"
+/// message_timeout_ms: 200
+/// status_on_error: 500
+/// processing_mode:
+///   request_header_mode: send
+///   response_header_mode: send
+///   request_body_mode: none
+///   response_body_mode: none
+/// ```
+pub struct ExtProcFilter {
```
twghu, line 468: derive debug trait could be useful

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+
+/// Controls which parts of the HTTP request and response are
+/// forwarded to the external processor.
+///
+/// Mirrors Envoy's [`ProcessingMode`] proto.
+///
+/// [`ProcessingMode`]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/processing_mode.proto
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct ProcessingModeConfig {
```
twghu, line 213: Would Clone + Copy be useful here as members have Copy.

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+#[async_trait]
+impl HttpFilter for ExtProcFilter {
+    fn name(&self) -> &'static str {
+        "ext_proc"
+    }
+
+    async fn on_request(&self, _ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
+        tracing::trace!(
+            target = %self.target,
+            "ext_proc on_request (skeleton)"
```
twghu: Does (skeleton) mean WIP?  AKA TODO?

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+/// Validate core numeric fields.
+fn validate_core_fields(cfg: &ExtProcConfig) -> Result<(), FilterError> {
+    if !(100..=599).contains(&cfg.status_on_error) {
+        return Err(format!(
+            "ext_proc: status_on_error {code} is not a valid HTTP status code (must be 100..=599)",
+            code = cfg.status_on_error,
+        )
+        .into());
+    }
+    if cfg.message_timeout_ms == 0 {
```
twghu, line 402: Should this have an upper bound validation?

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+/// processing_mode:
+///   request_header_mode: send
+///   response_header_mode: send
+///   request_body_mode: none
+///   response_body_mode: none
+/// ```
+pub struct ExtProcFilter {
+    /// Lazily-connecting gRPC channel to the external processor.
+    #[allow(dead_code, reason = "used by gRPC callout in subsequent PRs")]
+    channel: Channel,
```
twghu, line 474: A comment on praxis-proto dependency?

 The ExternalProcessorClient generated from praxis-proto is what will actually drive channel. Without a note here, the next PR author has to rediscover that praxis-proto needs to be added to Cargo.toml. A // TODO: ExternalProcessorClient from praxis-proto drives this channel comment on the field (or in from_config) makes the dependency explicit.

filter/ext-proc/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "praxis-proxy-ext-proc"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+description = "Envoy-compatible ext_proc filter for Praxis"
+license.workspace = true
+repository.workspace = true
+readme = "../../README.md"
+publish = true
```
twghu: This is premature until all approved

filter/ext-proc/src/tests.rs
```diff
@@ -0,0 +1,638 @@
...
+    assert_eq!(filter.name(), "ext_proc");
+}
+
+#[tokio::test]
+async fn parse_full_config_with_processing_mode() {
+    let yaml: serde_yaml::Value = serde_yaml::from_str(
+        r#"
+target: "http://127.0.0.1:50051"
+message_timeout_ms: 500
+max_message_timeout_ms: 5000
```
twghu, line 47: With the change suggested for from_config() in lib.rs this should validate timeout.

filter/ext-proc/src/tests.rs
```diff
@@ -0,0 +1,638 @@
...
+    let filter = ExtProcFilter::from_config(&yaml).unwrap();
+    assert_eq!(filter.name(), "ext_proc");
+}
+
+#[tokio::test]
+async fn accepts_custom_deferred_close_timeout() {
+    let yaml: serde_yaml::Value = serde_yaml::from_str(
+        r#"
+target: "http://127.0.0.1:50051"
+deferred_close_timeout_ms: 10000
```
twghu, line 383: accepts_custom_status_on_error and accepts_custom_deferred_close_timeout only assert filter.name() 

  These tests verify that certain configs are accepted, but the only assertion is filter.name() == "ext_proc", which is true for any successfully built filter. They're really just "doesn't error" tests. That's a limitation of the private field / Box<dyn HttpFilter> interface rather than a test bug, but it's worth noting the tests don't actually confirm the values were stored.

twghu (changes requested):
Some nits and suggestions

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+    /// Per-message timeout for gRPC calls.
+    #[allow(dead_code, reason = "used by gRPC callout in subsequent PRs")]
+    message_timeout: Duration,
+
+    /// HTTP status code returned on processor errors.
+    #[allow(dead_code, reason = "used by gRPC callout in subsequent PRs")]
+    status_on_error: u16,
+
+    /// gRPC endpoint URI (retained for diagnostics).
+    target: String,
```
twghu, line 485: Is this field in the right location?

filter/ext-proc/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+}
+
+#[async_trait]
+impl HttpFilter for ExtProcFilter {
+    fn name(&self) -> &'static str {
+        "ext_proc"
+    }
+
+    async fn on_request(&self, _ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
+        tracing::trace!(
```
shaneutt: Should be `warn!` instead, so if it ever gets triggered (for some reason) it's seen

```suggestion
        tracing::warn!(
```

filter/ext-proc/src/tests.rs
```diff
@@ -0,0 +1,638 @@
...
+target: "http://127.0.0.1:50051"
+message_timeout_ms: 0
+"#,
+    )
+    .unwrap();
+
+    let err = ExtProcFilter::from_config(&yaml).err().expect("should error");
+    assert!(
+        err.to_string().contains("message_timeout_ms"),
+        "error should mention message_timeout_ms: {err}"
```
shaneutt:
```suggestion
        "error should reject message_timeout_ms set to 0: {err}"
```

filter/ext-proc/Cargo.toml
shaneutt:
```suggestion
publish = false
```

shaneutt (changes requested):

filter/ext-proc/src/tests.rs
usize: That's true. I think that for a skeleton.. this seems okay? As the implementation proceeds, integration tests will hit the private fields. Or we can revisit and see if it's worth making it possible to inspect here.

filter/ext-proc/src/lib.rs
usize: I think it's worth having for logging purposes?

filter/ext-proc/src/lib.rs
usize: As a matter of taste, while there is definitely a reasonable limit on how long a timeout should be... it's more a matter of taste than a lower bound. Should someone have a timeout of 3600 * 1000? Well.. no. Probably not. How about 300 * 1000? Still high, but seems murkier.

shaneutt (approved):
Approving as-is for now. There's more we can do, but it's safe to assume we'll need to do follow-up PRs in _any_ case. Let's get what we have in, and iterate on it.

Thanks @usize!

PR #190, opened 2026-05-13
feat: add model alias config types and router validation

### LINKED ISSUE #77: Model Aliasing
Map model names in LLM requests to different upstream model identifiers, enabling abstraction over provider-specific model names.

Scope:
- Exact match aliases (e.g. "fast" -> "gpt-4o-mini")
- Wildcard pattern aliases with specificity ordering (e.g. "claude-*" -> route to Anthropic cluster)
- Per-route alias configuration
- Alias resolution before upstream routing

nerdalert:
## Summary

  Add model alias config types and router validation as the foundation for model aliasing (#139). This is PR 1 of 2. Its about as thin as I can get it to have PRs that build.

  - `ModelAlias` config struct with `match` (pattern) and optional `target`
  - `model_aliases: Option<Vec<ModelAlias>>` on `Route`
  - `RouterConfig` gains `model_header` and `model_alias_max_body_bytes`
  - Validation: empty aliases, empty pattern, multiple wildcards, empty
    target, invalid header name all rejected
  - Pattern matching and specificity helpers in `model_alias.rs`
  - Example config and all test literal fixups

  No runtime behavior change. Aliases parse and validate but the router ignores them at runtime.

  ## PR breakdown

  | | PR 1 (this PR) | PR 2 |
  |---|---|---|
  | **Scope** | Config types, validation, test infrastructure | Body processing, routing, Content-Length fix |
  | **Files** | 8 | 14 |
  | **Lines** | +784 | ~+600 |
  | **Runtime change** | None | Full feature activation |
  | **Core** | `ModelAlias` struct, `Route.model_aliases` field, export | — |
  | **Router config** | `model_header`, `model_alias_max_body_bytes`, defaults | — |
  | **Router construction** | `with_alias_options()`, alias validation, `resolve_routes()` | — |
  | **Alias matching** | `model_alias.rs`: pattern matching, specificity, resolution helpers | — |
  | **Body processing** | — | `on_request_body`: JSON parse, alias resolve, body rewrite |
  | **Route selection** | — | `on_request` rewrite: metadata check, alias-route gating, fallback pass |
  | **Content-Length** | — | `content_length_override` on `PingoraRequestCtx`, applied in upstream filter |
  | **Tests** | 7 (5 validation + 1 config parse + 1 pattern matching) | 14 unit + 3 schema + 5 integration |
  | **Docs** | Example config, `README.md` | `configuration.md`, `features.md`, `filters.md` |

  ## Test plan

  - [ ] `cargo test -p praxis-proxy-core route`
  - [ ] `cargo test -p praxis-proxy-filter router --features ai-inference` — 116 tests
  - [ ] `cargo clippy -p praxis-proxy-filter --features ai-inference` — clean
  - [ ] All existing router and pipeline tests pass unchanged
  - [ ] `all_example_configs_parse` validates the new example config

### DIFF
--- a/filter/src/builtins/http/traffic_management/router/config.rs
+++ b/filter/src/builtins/http/traffic_management/router/config.rs
@@ -7,6 +7,19 @@ use praxis_core::config::Route;
 use serde::Deserialize;
 
 // -----------------------------------------------------------------------------
+// Constants
+// -----------------------------------------------------------------------------
+
+/// Default header name for the resolved JSON alias value.
+pub(super) const DEFAULT_JSON_ALIAS_HEADER: &str = "X-Json-Alias";
+
+/// Default maximum body bytes to buffer for JSON alias resolution.
+pub(super) const DEFAULT_JSON_ALIAS_MAX_BODY_BYTES: usize = 10_485_760; // 10 MiB
+
+/// Hard upper bound for `json_alias_max_body_bytes`.
+pub(super) const MAX_JSON_ALIAS_BODY_BYTES: usize = 67_108_864; // 64 MiB
+
+// -----------------------------------------------------------------------------
 // RouterConfig
 // -----------------------------------------------------------------------------
 
@@ -14,7 +27,64 @@ use serde::Deserialize;
 #[derive(Deserialize)]
 #[serde(deny_unknown_fields)]
 pub(super) struct RouterConfig {
+    /// Header name for the promoted JSON field value during alias
+    /// resolution.
+    #[serde(default = "default_json_alias_header")]
+    pub json_alias_header: String,
+
+    /// Maximum body bytes to buffer when resolving JSON aliases.
+    #[serde(default = "default_json_alias_max_body_bytes")]
+    pub json_alias_max_body_bytes: usize,
+
     /// Route table entries.
     #[serde(default)]
-    pub routes: Vec<Route>,
+    pub routes: Vec<RouterRouteConfig>,
+}
+
+/// Router-owned route config so JSON body aliasing stays out of
+/// [`praxis_core::config::Route`].
+#[derive(Debug, Clone, Deserialize)]
+pub(super) struct RouterRouteConfig {
+    /// Generic path, host, header, and cluster routing fields.
+    #[serde(flatten)]
+    pub route: Route,
+
+    /// Optional JSON field aliases evaluated for this route.
+    #[serde(default)]
+    pub json_aliases: Option<Vec<JsonAlias>>,
+}
+
+impl From<Route> for RouterRouteConfig {
+    fn from(route: Route) -> Self {
+        Self {
+            route,
+            json_aliases: None,
+        }
+    }
+}
+
+/// JSON field alias rule scoped to a router route.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(super) struct JsonAlias {
+    /// Request JSON field whose string value is compared with `pattern`.
+    pub field: String,
+
+    /// Exact or single-wildcard pattern for the configured field value.
+    #[serde(rename = "match")]
+    pub pattern: String,
+
+    /// Replacement value; omitted aliases preserve the original value.
+    #[serde(default)]
+    pub target: Option<String>,
+}
+
+/// Serde default for [`RouterConfig::json_alias_header`].
+fn default_json_alias_header() -> String {
+    DEFAULT_JSON_ALIAS_HEADER.to_owned()
+}
+
+/// Serde default for [`RouterConfig::json_alias_max_body_bytes`].
+fn default_json_alias_max_body_bytes() -> usize {
+    DEFAULT_JSON_ALIAS_MAX_BODY_BYTES
 }
--- a/filter/src/builtins/http/traffic_management/router/mod.rs
+++ b/filter/src/builtins/http/traffic_management/router/mod.rs
@@ -4,6 +4,7 @@
 //! Path-prefix and host-header routing filter.
 
 mod config;
+mod json_alias;
 mod matching;
 
 #[cfg(test)]
@@ -21,12 +22,15 @@ mod tests;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use http::HeaderMap;
+use http::{HeaderMap, header::HeaderName};
 use praxis_core::config::{PathMatch, Route};
 use tracing::{debug, trace};
 
 use self::{
-    config::RouterConfig,
+    config::{
+        DEFAULT_JSON_ALIAS_HEADER, DEFAULT_JSON_ALIAS_MAX_BODY_BYTES, JsonAlias, MAX_JSON_ALIAS_BODY_BYTES,
+        RouterConfig, RouterRouteConfig,
+    },
     matching::{route_matches_request, should_stop_early, update_best_match},
 };
 use crate::{
@@ -74,6 +78,18 @@ use crate::{
 /// [`rewritten_path`]: crate::HttpFilterContext::rewritten_path
 #[derive(Debug)]
 pub struct RouterFilter {
+    /// Whether any route has JSON aliases configured.
+    #[allow(dead_code, reason = "alias config is validated before body access is wired")]
+    has_json_alias_routes: bool,
+
+    /// Maximum body bytes to buffer for JSON alias resolution.
+    #[allow(dead_code, reason = "alias config is validated before body buffering is wired")]
+    json_alias_max_body_bytes: usize,
+
+    /// Header name for the promoted JSON alias value.
+    #[allow(dead_code, reason = "alias config is validated before header promotion is wired")]
+    json_alias_header: HeaderName,
+
     /// Ordered route table with pre-computed wildcard suffixes.
     routes: Vec<ResolvedRoute>,
 }
@@ -84,6 +100,10 @@ struct ResolvedRoute {
     /// The original route configuration.
     route: Route,
 
+    /// Optional JSON aliases configured on this route.
+    #[allow(dead_code, reason = "route aliases are validated before body aliasing is wired")]
+    json_aliases: Option<Vec<JsonAlias>>,
+
     /// For wildcard hosts (e.g. `*.example.com`), the pre-lowercased
     /// suffix with leading dot: `.example.com`. `None` for exact hosts
     /// or routes without a host constraint.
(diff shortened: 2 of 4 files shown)


filter/src/builtins/http/traffic_management/router/mod.rs
```diff
@@ -222,6 +269,78 @@ fn validate_prefixes(routes: &[Route]) -> Result<(), FilterError> {
...
+/// Validates global alias options when alias routes exist.
+fn validate_alias_options(routes: &[Route], model_header: &str, max_bytes: usize) -> Result<(), FilterError> {
+    let has_aliases = routes.iter().any(|r| r.model_aliases.is_some());
+    if !has_aliases {
+        return Ok(());
+    }
+    if http::header::HeaderName::from_bytes(model_header.as_bytes()).is_err() {
+        return Err(format!("router: model_header '{model_header}' is not a valid HTTP header name").into());
+    }
+    if max_bytes == 0 {
```
shaneutt, line 334: The validation here rejects `max_bytes == 0` but accepts arbitrarily large values. This could become a vector for memory exhaustion attacks down the line, let's add a constant with some upper bound (maybe start with 64 MiB for now?) and reject values above it.

filter/src/builtins/http/traffic_management/router/model_alias.rs
```diff
@@ -0,0 +1,332 @@
...
+/// use praxis_filter::router::model_alias::pattern_matches;
+///
+/// assert!(pattern_matches("fast", "fast"));
+/// assert!(pattern_matches("claude-*", "claude-3"));
+/// assert!(pattern_matches("gpt-*-turbo", "gpt-4-turbo"));
+/// assert!(!pattern_matches("claude-*", "gpt-4"));
+/// ```
+pub fn pattern_matches(pattern: &str, model: &str) -> bool {
+    if let Some(wildcard_pos) = pattern.find('*') {
+        // Wildcard pattern: split into prefix and suffix
```
shaneutt: A note on conventions: I'm seeing a few instances of comments narrating _what_ the code does. Our [conventions on comments vs. tracing](https://github.com/praxis-proxy/praxis/blob/main/docs/conventions.md#comments-vs-tracing) ask for comments to answer "why?" and if you want to decorate "what?" use a debug/trace instead.

shaneutt (changes requested):

filter/src/builtins/http/traffic_management/router/config.rs
```diff
@@ -14,7 +24,26 @@ use serde::Deserialize;
 #[derive(Deserialize)]
 #[serde(deny_unknown_fields)]
 pub(super) struct RouterConfig {
+    /// Header name for the promoted model value during alias
+    /// resolution.
+    #[serde(default = "default_model_header")]
```
shaneutt: The existing test for RouterConfig sets both these defaulted fields explicitly, we might as well test the defaults too.

filter/src/builtins/http/traffic_management/router/mod.rs
```diff
@@ -222,6 +269,78 @@ fn validate_prefixes(routes: &[Route]) -> Result<(), FilterError> {
...
+             must not be empty (omit the key to preserve the original model)",
+            alias.pattern,
+        )
+        .into());
+    }
+    Ok(())
+}
+
+/// Validates global alias options when alias routes exist.
+fn validate_alias_options(routes: &[Route], model_header: &str, max_bytes: usize) -> Result<(), FilterError> {
```
shaneutt: This function has two error paths but I think zero test coverage?

filter/src/builtins/http/traffic_management/router/tests.rs
```diff
@@ -1183,6 +1221,119 @@ fn mixed_exact_and_prefix_ordering() {
...
+    }])
+    .unwrap_err();
+    assert!(
+        err.to_string().contains("must not be empty"),
+        "empty target should be rejected: {err}"
+    );
+}
+
+#[test]
+fn model_alias_from_config_parses_correctly() {
```
shaneutt: This test only asserts `filter.name() == "router"`, which passes for _any_ valid router configuration regardless of whether aliases were parsed correctly. We need to verify that the parsed config actually contains the expected aliases, header name or max body bytes as well.

filter/src/builtins/http/traffic_management/router/mod.rs
nerdalert: Added `MAX_MODEL_ALIAS_BODY_BYTES = 67_108_864` with the reject conditions.

filter/src/builtins/http/traffic_management/router/model_alias.rs
nerdalert: Thanks for sharing that, hadn't read that yet!

filter/src/builtins/http/traffic_management/router/config.rs
nerdalert: Added tests for default values when YAML omits `model_header` and `model_alias_max_body_bytes`

filter/src/builtins/http/traffic_management/router/tests.rs
nerdalert: test to verify parsed fields and route aliases, not just the filter name.

filter/src/builtins/http/traffic_management/router/mod.rs
nerdalert: Added tests that trigger each error branch in validate_alias_options.

examples/configs/ai/model-aliasing.yaml
```diff
@@ -0,0 +1,68 @@
+# AI Model Aliasing
```
shaneutt: Missing integration test for this example config?

shaneutt (changes requested):

core/src/config/route.rs
```diff
@@ -112,6 +112,53 @@ impl PathMatch {
     }
 }
 
+// -----------------------------------------------------------------------------
```
shaneutt: `core` is not the right place for AI-specific things. We can flatten route in with this at the filter level instead, see `PathMatch`.

core/src/config/route.rs
```diff
@@ -112,6 +112,53 @@ impl PathMatch {
...
+/// ```
+/// use praxis_core::config::ModelAlias;
+///
+/// let alias: ModelAlias = serde_yaml::from_str("match: claude-*\n").unwrap();
+/// assert_eq!(alias.pattern, "claude-*");
+/// assert!(alias.target.is_none());
+/// ```
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct ModelAlias {
```
shaneutt: This is hyper-specific to inference, but it seems like we could generalize this further and instead of assuming the `model` field, we could make the field configurable. This would allow more generic aliasing for JSON fields. LMKWYT? 🤔

shaneutt (changes requested):

core/src/config/route.rs
```diff
@@ -112,10 +112,6 @@ impl PathMatch {
     }
 }
 
-// -----------------------------------------------------------------------------
-// Route
-// -----------------------------------------------------------------------------
-
 /// A routing rule mapping requests to a cluster.
```
shaneutt, line 119: Revert?

nerdalert: Haha, you beat me to it.

core/src/config/route.rs
nerdalert: Thanks! Moved alias config out of praxis_core::config::Route and into the router filter config layer. The router now flattens the normal route fields into its own route config wrapper, and core::Route is back to generic path/host/header/cluster matching only.

core/src/config/route.rs
nerdalert: Changed the alias config from model-specific aliases to generic JSON field aliases. Each alias now declares its own field, so model is just one possible JSON field rather than a hard-coded assumption.

filter/src/builtins/http/traffic_management/router/mod.rs
```diff
@@ -4,6 +4,7 @@
 //! Path-prefix and host-header routing filter.
 
 mod config;
+pub(crate) mod json_alias;
```
shaneutt:
```suggestion
mod json_alias;
```

filter/src/builtins/http/traffic_management/router/mod.rs
```diff
@@ -74,6 +78,18 @@ use crate::{
...
+    #[allow(dead_code, reason = "alias config is validated before body access is wired")]
+    has_json_alias_routes: bool,
+
+    /// Maximum body bytes to buffer for JSON alias resolution.
+    #[allow(dead_code, reason = "alias config is validated before body buffering is wired")]
+    json_alias_max_body_bytes: usize,
+
+    /// Header name for the promoted JSON alias value.
+    #[allow(dead_code, reason = "alias config is validated before header promotion is wired")]
+    json_alias_header: Cow<'static, str>,
```
shaneutt: In validation we already parse this into `http::header::HeaderName`, so we might consider just doing that here anyway:


```suggestion
    json_alias_header: http::header::HeaderName,
```

shaneutt (changes requested):

filter/src/builtins/http/traffic_management/router/mod.rs
nerdalert: Done ty.

filter/src/builtins/http/traffic_management/router/mod.rs
nerdalert: Done, ty

filter/src/builtins/http/traffic_management/router/config.rs
```diff
@@ -14,7 +27,64 @@ use serde::Deserialize;
...
+/// [`praxis_core::config::Route`].
+#[derive(Debug, Clone, Deserialize)]
+pub(super) struct RouterRouteConfig {
+    /// Generic path, host, header, and cluster routing fields.
+    #[serde(flatten)]
+    pub route: Route,
+
+    /// Optional JSON field aliases evaluated for this route.
+    #[serde(default)]
+    pub json_aliases: Option<Vec<JsonAlias>>,
```
shaneutt, line 54: I'm still not super-comfortable with where this lives, but we've done enough work in this iteration and we've simplified enough that I think this can be reconsidered in a follow up. I've created https://github.com/praxis-proxy/praxis/issues/366 for that purpose, please feel free to consider this comment resolved at your discretion.

shaneutt (approved):

PR #195, opened 2026-05-13
feat: add MCP classifier filter for body-aware routing

### LINKED ISSUE #149: MCP Protocol Support
- [ ] Use Kuadrant MCP Gateway if feasible
- [ ] Model Context Protocol proxy and routing
- [ ] JSON-RPC 2.0 message parsing and forwarding
- [ ] MCP-specific filter hooks (tool invocation, context management)
- [ ] Integration tests + example config

nerdalert:
## Summary

  Adds an `mcp` filter that parses JSON-RPC request bodies to extract MCP protocol metadata for routing. This is MCP classification and static routing only, not full MCP gateway support. See the entire e2e PoC for agentic and how the PRs will be chunked here https://github.com/praxis-proxy/ai/issues/148#issuecomment-4368239729. Also a section on what was added for the broader end to end needs: 

  - Classifies 14 known MCP methods (`initialize`, `tools/call`, `tools/list`, `resources/read`, `prompts/get`, `ping`, etc.) with `Other` for unknown methods
  - Extracts `params.name` (tools/call, prompts/get), `params.uri` (resources/read), session ID from `MCP-Session-Id`, protocol version from initialize body or
  `MCP-Protocol-Version` header
  - Promotes durable metadata (`mcp.method`, `mcp.name`, `mcp.session_id`, `mcp.protocol_version`) and filter results under `"mcp"` key
  - Promotes internal routing headers (`x-praxis-mcp-method`, `x-praxis-mcp-name`, `x-praxis-mcp-kind`, `x-praxis-mcp-session-present`) for router matching —
  stripped before upstream by PR 2 header hygiene
  - Validates `Mcp-Method`/`Mcp-Name` headers against body-derived values; mismatches rejected with JSON-RPC error `-32001 HeaderMismatch`
  - Rejects JSON-RPC batches regardless of `on_invalid` config
  - Enforces required `params.name`/`params.uri` for methods that need them
  - Handles partial StreamBuffer chunks correctly (continues before EOS, processes at EOS)
  - Reuses the shared `json_rpc` parser — no duplicate envelope parsing

  ### Config

  ```yaml
  filter: mcp
  max_body_bytes: 65536
  on_invalid: reject        # or continue (pass-through for old SSE transport, I validated both with FastMCP which you can see in the epic issue)
  header_validation:
    mismatch: reject        # or ignore
    missing: reject         # reject, synthesize, or ignore
```

 ### Follow-up: header validation hardening

  A second PR will add on top of this one. I genuinely worked hard to try and get this smoll but at 1500 LOC apparently failed 😿 The tests ballooned it a lot even with trying to split hardening/testing out:
  - `missing: synthesize` actually injecting standard `mcp-method` / `mcp-name` headers
  - Required `params.name` / `params.uri` enforcement for:
    - `tools/call`
    - `prompts/get`
    - `resources/read`
  - Batch request rejection regardless of `on_invalid`
  - Safer partial chunk handling for StreamBuffer parse failures before EOS
  - Spurious `Mcp-Name` rejection for nameless methods such as `tools/list`
  - HeaderMismatch responses preserving the original JSON-RPC id
  - Additional unit and integration tests for the above hardening cases

### Expanded scope from the issue

  The original issue focused on MCP-aware routing, but this PR also lays down several reusable pieces needed by the broader MCP/A2A plan:

  - Uses the shared JSON-RPC parser instead of introducing MCP-specific parsing logic.
  - Writes durable per-request metadata so later filters, response handling, and logging can access MCP context beyond the body phase.
  - Extracts MCP session presence and protocol version, even though full MCP session management is deferred to later gateway/session PRs.
  - Promotes bounded internal routing headers for compatibility with the existing `router` filter, allowing useful static MCP routing before full `mcp_gateway` behavior exists.
  - Adds strict header/body validation around `Mcp-Method` and `Mcp-Name` to prevent clients from spoofing routing decisions with headers.
  - Includes compatibility behavior for legacy/old-SSE deployments through `on_invalid: continue` and configurable missing-header handling.
  - Adds docs, examples, and integration coverage so the classifier can be exercised independently before catalog aggregation, backend session mapping, and tool-prefix rewriting land later.
 ### Test plan

  - 34 unit tests: config parsing, method classification, name/URI extraction, protocol version, session detection, metadata promotion, header promotion, filter
   results, mismatch rejection, missing-header rejection, required-name enforcement, batch rejection, partial-chunk handling, compatibility mode
  - 12 integration tests: routing by tool name, default fallback, header/body mismatch 400, missing-name 400, batch rejection, non-JSON pass-through, standard
  MCP headers preserved upstream, internal headers stripped, example config routing
  - Full suite: 407 pass, 0 regressions, 0 clippy errors

  Refs praxis-proxy/ai#149

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -487,6 +487,7 @@ A filter can have both `conditions` (request phase) and
 | `credential_injection` | Security | HTTP | Per-cluster API key injection with client credential stripping. Literal `value` fields are redacted in `--dump` output. |
 | `json_body_field` | Payload Processing | HTTP | Extract a JSON body field and promote to header |
 | `json_rpc` | Payload Processing | HTTP | Parse JSON-RPC 2.0 envelopes and extract method/id/kind for routing |
+| `mcp` | Payload Processing | HTTP | MCP protocol classifier: extract method, tool/resource/prompt name, and session metadata for routing |
 | `compression` | Payload Processing | HTTP | Gzip, brotli, and zstd response compression |
 | `cors` | Security | HTTP | CORS preflight handling, origin validation, credential support |
 | `csrf` | Security | HTTP | Origin-based CSRF protection with gradual rollout and Sec-Fetch-Site support |
--- a/examples/configs/payload-processing/mcp-classifier-routing.yaml
+++ b/examples/configs/payload-processing/mcp-classifier-routing.yaml
@@ -1,0 +1,59 @@
+# MCP Classifier Routing
+#
+# Routes MCP requests by body-derived method and tool name.
+# The `mcp` filter parses JSON-RPC bodies and promotes internal
+# routing headers (x-praxis-mcp-*) for router matching. These
+# internal headers are stripped before upstream by header hygiene.
+#
+# This example uses `missing: synthesize` so it works with real
+# MCP clients that do not send Mcp-Method / Mcp-Name headers.
+# Standard MCP headers are synthesized from body-derived values
+# and forwarded to backends. Mismatched headers are still rejected.
+#
+# For non-MCP traffic (e.g. old SSE transport), set on_invalid:
+# continue and header_validation.missing: ignore.
+
+listeners:
+  - name: mcp
+    address: "127.0.0.1:8080"
+    filter_chains: [mcp-routing]
+
+filter_chains:
+  - name: mcp-routing
+    filters:
+      - filter: mcp
+        max_body_bytes: 65536
+        on_invalid: reject
+        header_validation:
+          mismatch: reject
+          missing: synthesize
+        headers:
+          method: x-praxis-mcp-method
+          name: x-praxis-mcp-name
+          kind: x-praxis-mcp-kind
+          session_present: x-praxis-mcp-session-present
+
+      - filter: router
+        routes:
+          - path_prefix: "/"
+            headers:
+              x-praxis-mcp-name: "get_weather"
+            cluster: "weather-tools"
+          - path_prefix: "/"
+            headers:
+              x-praxis-mcp-name: "get_calendar"
+            cluster: "calendar-tools"
+          - path_prefix: "/"
+            cluster: "default-mcp"
+
+      - filter: load_balancer
+        clusters:
+          - name: "weather-tools"
+            endpoints:
+              - "127.0.0.1:9001"
+          - name: "calendar-tools"
+            endpoints:
+              - "127.0.0.1:9002"
+          - name: "default-mcp"
+            endpoints:
+              - "127.0.0.1:9003"
--- a/filter/src/builtins/http/payload_processing/mcp/config.rs
+++ b/filter/src/builtins/http/payload_processing/mcp/config.rs
@@ -1,0 +1,199 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Configuration types for the MCP filter.
+
+use serde::Deserialize;
+
+use crate::FilterError;
+
+// -----------------------------------------------------------------------------
+// Constants
+// -----------------------------------------------------------------------------
+
+/// Default maximum request body size for `StreamBuffer` mode (64 `KiB`).
+pub(crate) const DEFAULT_MAX_BODY_BYTES: usize = 65_536;
+
+// -----------------------------------------------------------------------------
+// Behavior Enums
+// -----------------------------------------------------------------------------
+
+/// Header validation mismatch behavior.
+#[derive(Debug, Clone, Copy, Default, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum MismatchBehavior {
+    /// Reject the request with a JSON-RPC error response.
+    #[default]
+    Reject,
+    /// Ignore the mismatch and continue processing.
+    Ignore,
+}
+
+/// Header validation missing behavior.
+#[derive(Debug, Clone, Copy, Default, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum MissingHeaderBehavior {
+    /// Ignore the missing headers and continue processing.
+    #[default]
+    Ignore,
+    /// Synthesize the missing headers from body-derived values.
+    Synthesize,
+    /// Reject the request when expected MCP headers are absent.
+    Reject,
+}
+
+/// Invalid MCP message handling.
+#[derive(Debug, Clone, Copy, Default, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum InvalidMcpBehavior {
+    /// Reject non-MCP input with HTTP 400.
+    #[default]
+    Reject,
+    /// Continue processing without MCP metadata.
+    Continue,
+}
+
+// -----------------------------------------------------------------------------
+// HeaderValidation
+// -----------------------------------------------------------------------------
+
+/// Header validation config controlling how header/body mismatches are handled.
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct HeaderValidation {
+    /// Behavior when header value conflicts with body-derived value.
+    #[serde(default)]
+    pub mismatch: MismatchBehavior,
+    /// Behavior when expected MCP headers are absent.
+    #[serde(default)]
+    pub missing: MissingHeaderBehavior,
+}
+
+// -----------------------------------------------------------------------------
+// McpHeaders
+// -----------------------------------------------------------------------------
+
+/// Promoted header names for MCP metadata.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct McpHeaders {
+    /// Header name for the MCP method (e.g. `x-praxis-mcp-method`).
+    #[serde(default = "default_method_header")]
+    pub method: Option<String>,
+    /// Header name for the tool/resource/prompt name (e.g. `x-praxis-mcp-name`).
+    #[serde(default = "default_name_header")]
+    pub name: Option<String>,
+    /// Header name for the JSON-RPC kind (e.g. `x-praxis-mcp-kind`).
+    #[serde(default = "default_kind_header")]
+    pub kind: Option<String>,
+    /// Header name for MCP session presence (e.g. `x-praxis-mcp-session-present`).
+    #[serde(default = "default_session_present_header")]
+    pub session_present: Option<String>,
+}
+
+impl Default for McpHeaders {
+    fn default() -> Self {
+        Self {
+            method: default_method_header(),
+            name: default_name_header(),
+            kind: default_kind_header(),
+            session_present: default_session_present_header(),
+        }
+    }
+}
+
+/// Default method header name.
+#[allow(
+    clippy::unnecessary_wraps,
+    reason = "serde default functions require Option return type"
+)]
+fn default_method_header() -> Option<String> {
+    Some("x-praxis-mcp-method".to_owned())
+}
+
+/// Default name header name.
+#[allow(
+    clippy::unnecessary_wraps,
+    reason = "serde default functions require Option return type"
+)]
+fn default_name_header() -> Option<String> {
+    Some("x-praxis-mcp-name".to_owned())
+}
+
+/// Default kind header name.
+#[allow(
+    clippy::unnecessary_wraps,
+    reason = "serde default functions require Option return type"
+)]
+fn default_kind_header() -> Option<String> {
+    Some("x-praxis-mcp-kind".to_owned())
+}
+
+/// Default session-present header name.
+#[allow(
+    clippy::unnecessary_wraps,
+    reason = "serde default functions require Option return type"
+)]
+fn default_session_present_header() -> Option<String> {
+    Some("x-praxis-mcp-session-present".to_owned())
+}
+
+// -----------------------------------------------------------------------------
+// McpConfig
+// -----------------------------------------------------------------------------
+
+/// YAML configuration for the MCP filter.
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct McpConfig {
+    /// Maximum body size in bytes for `StreamBuffer`.
+    #[serde(default = "default_max_body_bytes")]
+    pub max_body_bytes: usize,
+
+    /// Invalid input handling behavior.
+    #[serde(default)]
+    pub on_invalid: InvalidMcpBehavior,
+
+    /// Header validation settings.
+    #[serde(default)]
+    pub header_validation: HeaderValidation,
+
+    /// Header names for MCP metadata promotion.
+    #[serde(default)]
+    pub headers: McpHeaders,
+}
+
+/// Default max body bytes.
+fn default_max_body_bytes() -> usize {
+    DEFAULT_MAX_BODY_BYTES
+}
+
+// -----------------------------------------------------------------------------
+// Config Validation
+// -----------------------------------------------------------------------------
+
+/// Validate and build the final configuration.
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: filter/src/builtins/http/payload_processing/mcp/envelope.rs [hunk 1]
```diff
@@ -0,0 +1,172 @@
...
+    /// Protocol version from initialize params or `Mcp-Protocol-Version` header.
+    pub protocol_version: Option<String>,
+}
+
+// -----------------------------------------------------------------------------
+// Extraction
+// -----------------------------------------------------------------------------
+
+/// Extract MCP-specific metadata from parsed JSON body and request headers.
+pub(crate) fn extract_mcp_envelope(body: &[u8], method_str: &str, request_headers: &http::HeaderMap) -> McpEnvelope {
```
shaneutt: With the `extract_name` and `extract_protocol_version` helpers, it looks like we're parsing the body 3 times per request.

It looks like we could parse the body once, e.g.:

```rust
let parsed = serde_json::from_slice::<Value>(body).ok();
```

And then pass the parsed `&Value` to the helpers instead to reduce it

filter/src/builtins/http/payload_processing/mcp/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// Promote MCP metadata to filter results for router branch conditions.
+fn promote_filter_results(
+    ctx: &mut HttpFilterContext<'_>,
+    envelope: &crate::builtins::http::payload_processing::json_rpc::envelope::JsonRpcEnvelope,
+    mcp: &McpEnvelope,
+) -> Result<(), FilterError> {
+    let results = ctx.filter_results.entry("mcp").or_default();
+    results.set("method", mcp.method.as_str().to_owned())?;
```
shaneutt: Would we potentially want to check for control characters here as well? 🤔

filter/src/builtins/http/payload_processing/mcp/envelope.rs
nerdalert: Ty, good catch. To clarify a bit, not every MCP method hit all of those paths. Here is an example, tools/list does not need params.name or protocolVersion, while tools/call needs name extraction, and initialize needs protocol-version extraction. So the exact number of parses depended on the method, but it was still unnecessary extra parsing. Changed it so the PR now avoids that whole class of reparse by sharing the already parsed JSON value.

filter/src/builtins/http/payload_processing/mcp/mod.rs
nerdalert: Header promotion already skipped unsafe control-character values, but the method filter result did not, so a malformed method could cause a filter error. I made filter results and metadata use the same guard and (and test)t for a method containing an escaped newline.

docs/filters.md
twghu, line 1: session is not the correct qualifier: from envelope.rs

let session_id = request_headers
.get("mcp-session-id")
.and_then(|v| v.to_str().ok())
.map(str::to_owned);

Maybe "MCP protocol classifier: extract method, tool name, and session metadata for routing" which makes no false claim about the source

shaneutt (dismissed):
Waiting on @twghu's review. When he's :+1: I'm :+1:.

examples/configs/payload-processing/mcp-classifier-routing.yaml
twghu, line 1: The config below the comment "strict mode (default)" uses 
missing: synthesize,

which neither matches strict mode nor the compatibility mode the comment describes. 

The comment presents two modes — strict and compatibility — but the config implements a third hybrid (mismatch: reject, missing:  synthesize) without the comment acknowledging it.

The comment implies the config is demonstrating strict mode, but the config is actually a hybrid that isn't described  anywhere in the comments. A reader would look at the comment, expect missing: reject, then see missing: synthesize and not understand why.

Update the comment to accurately describe what the config is actually doing and why synthesize is the right choice for a working  example with real MCP clients?

Integration tests always inject headers manually, never testing missing: synthesize.

filter/src/builtins/http/payload_processing/mcp/config.rs
twghu, line 1: - Not wanting to be picky :( iirc struct field ordering is best alphabetical: McpHeaders, McpConfig
- File ordering - private functions interspersed with public types convention requires public types then private 
- not sure if this is a problem but lint picked up 

pub(crate) fn build_config(cfg: McpConfig) -> Result<(usize, McpConfig), FilterError>

  The usize in the tuple is cfg.max_body_bytes, which is already inside the returned McpConfig. The caller unpacks both and stores them  separately, but validated_config.max_body_bytes already has the same value. The return type could be Result<McpConfig, FilterError> and the  caller reads config.max_body_bytes directly.

- MissingHeaderBehavior default is Reject (line 37–38) (see general review comment for details)

-   http::HeaderName::from_bytes rejects empty input already. The explicit empty check only exists to produce a different error message. That's a  reasonable justification but worth noting it's defensive rather than necessary (line 192-194)

filter/src/builtins/http/payload_processing/mcp/envelope.rs
twghu, line 1: - Module doc comment (line 4) "MCP-specific extraction from JSON-RPC request bodies."  session comes from the request header, protocol version can come from either.
- from_method_str / as_str roundtrip   Every pair is a correct inverse from_method_str(m.as_str()) == m holds for all variants including Other. Not tested?   A roundtrip property test could save glitches from later code mods where edits are not matched.
- requires_name  (line 90) maybe comment as "whether extraction of params.name is attempted for this method."
- McpEnvelope field ordering
-  extract_protocol_version (lines 156–170) The fallback at lines 166–169 reads mcp-protocol-version from the request header for every method  unconditionally — including tools/call, ping, anything. The function doc says "from initialize params or Mcp-Protocol-Version header" which  accurately describes the behaviour, but there is no explanation of why non-initialize methods are allowed to set protocol version metadata via
  header.  Some detailed comment about handling would help future understanding :)

filter/src/builtins/http/payload_processing/mcp/mod.rs
twghu, line 1: - Line 19 - mod tests placement after uses? 
- Line 64 - relates to comment on the filters.md about reject?
- Line 73-78 McpFilter field order.
-   validate_single_header treats invalid UTF-8 header as missing (line 229)
  ctx.request.headers.get(header_name).and_then(|v| v.to_str().ok())
  If Mcp-Method is present but contains invalid UTF-8, to_str() returns None and the else branch fires — the header is treated as missing rather  than as a mismatch. With missing: synthesize the body value gets injected, silently overriding a present but unreadable header. With missing:  reject it gets a bare 400 with no indication why. 
- contains_control_chars not applied to kind (lines 283, 324, 353) thinking that even thought values come from enum, check for safe assignment of meta data generally.
-  mcp_header_mismatch_rejection hardcodes "id":null (line 264)
    Previously identified. envelope is available in on_request_body at the call site but not threaded.

 The JSON-RPC 2.0 spec requires that error responses echo the id from the request. If a client sends:

  {"jsonrpc":"2.0","id":42,"method":"tools/call","params":{"name":"get_weather"}}

  with a mismatched Mcp-Method header, they get back:

  {"jsonrpc":"2.0","error":{"code":-32001,"message":"HeaderMismatch"},"id":null}

  The id should be 42. A client correlating async responses by id gets the wrong answer.

  The fix requires threading. In on_request_body, envelope is in scope at line 141 and envelope.id is Some("42"). But validate_mcp_headers at line 151 doesn't receive envelope — it only gets ctx, mcp_envelope, and config. By the time mcp_header_mismatch_rejection is called inside validate_single_header, the parsed id is gone. To fix it, envelope.id (and envelope.id_kind to know whether to serialize it as string, number, or null) would need to be passed down through validate_mcp_headers validate_single_header → mcp_header_mismatch_rejection.

filter/src/builtins/http/payload_processing/mcp/tests.rs
twghu, line 1: - Missing test: batch with on_invalid: continue

No test sends a JSON-RPC batch array with on_invalid: continue to document the pass-through behaviour. The batch finding from earlier has no  test coverage at all.

- Test for case
where Mcp-Method and Mcp-Name headers are present, match the body, and processing succeeds. The matching branch of   validate_single_header (where header_value == body_value) is untested.
There is  no test where Mcp-Method and Mcp-Name headers are present, match the body, and processing succeeds. The matching branch of  validate_single_header (where header_value == body_value) is untested.

tests/integration/tests/suite/examples/payload_processing.rs
twghu, line 1: - calendar_guard but never send a request that routes to it.. seems untested
- mcp_json_post duplicates json_post_with_mcp_headers from mcp.rs ?

tests/integration/tests/suite/mcp.rs
twghu, line 1: Missing tests?

  - mismatch: ignore path — no test anywhere in this file
  - Batch with on_invalid: continue — no test
  - id field in HeaderMismatch response — not asserted

twghu:
The implementation is well-structured. Parser reuse via parse_json_rpc_value
is good. StreamBuffer + EOS gating is the established pattern in this codebase and is used correctly. McpMethod::Other(String)  handles forward-compatibility cleanly. Control character guards are applied
consistently across all promotion paths. The three-mode design (pass-through /  classifier / gateway) is clearly articulated and the PR stays within its  stated scope. Test count (34 unit, 12 integration) is good.
  
Worth checking chunk boundaries are handled (i think they are).

Two behavioral bugs?

  1. Default missing: reject rejects all real MCP traffic

  MissingHeaderBehavior defaults to Reject. Real MCP clients — including FastMCP
  cited in the PR — do not send Mcp-Method/Mcp-Name headers; those are
  Praxis-internal conventions, not part of the MCP spec. A proxy deployed with
  default config will 400 every legitimate MCP request. The PR's own example
  config, integration test helpers, and the issue praxis-proxy/ai#148 comment all use synthesize
  or ignore to work around this — which confirms the default is wrong. Strict
  header enforcement must be opt-in, not opt-out.

  2. PR summary misrepresents batch rejection behavior

  BatchPolicy::Reject causes parse_json_rpc_value to return
  Err(UnsupportedBatch). The let Ok(Some(envelope)) pattern treats that error
  identically to non-MCP input — it falls through to handle_non_mcp. With
  on_invalid: continue, batches pass through to the backend. The PR summary
  states "Rejects JSON-RPC batches regardless of on_invalid config" — this is
  only true when on_invalid: reject (the default). The claim is inaccurate and
  needs correcting.
  
Summary
  1. Change default missing behavior from Reject to Ignore
  2. Correct the batch rejection claim in the PR summary
  3. Add an integration test sending a fragmented body and verifying correct
  extraction — a tools/call body split across at least two chunks, asserting
  method and tool name are correctly promoted. A second test should document (or
  enforce) the behavior when the body exceeds max_body_bytes.

docs/filters.md
nerdalert: Agreed. I updated the filter description so it does not imply the session id is body-derived; method/name come from the JSON-RPC body, while session metadata comes from MCP headers. ty!

examples/configs/payload-processing/mcp-classifier-routing.yaml
nerdalert: Good catch. I changed the example comments to describe the actual mode: strict mismatch rejection with missing-header synthesis, which makes the example work with real MCP clients that do not send Mcp-Method or Mcp-Name.

filter/src/builtins/http/payload_processing/mcp/config.rs
nerdalert: Not picky at all! 

Addressed the substantive pieces: missing now defaults to ignore so default classifier config works with real MCP clients, and build_config now returns only the validated config since max_body_bytes is already part of it. Also cleaned up the ordering where it was low-risk and kept the explicit empty-name check for a clearer error.

filter/src/builtins/http/payload_processing/mcp/envelope.rs
nerdalert: Updated docs to state that MCP metadata comes from both the parsed JSON-RPC value and MCP headers, added stronger method roundtrip coverage including Other, and documented why Mcp-Protocol-Version can be read for non-initialize methods.

filter/src/builtins/http/payload_processing/mcp/mod.rs
nerdalert: Fixed the behavioral issues here: invalid UTF-8 MCP headers are now treated as malformed present headers instead of missing headers, and HeaderMismatch responses now echo the parsed JSON-RPC id with the correct JSON type. Also updated the docs/default example to match the new default missing-header behavior.

filter/src/builtins/http/payload_processing/mcp/tests.rs
nerdalert: Added unit coverage for batch handling under on_invalid: continue, the successful matching-header branch, the new default missing-header behavior, and HeaderMismatch id preservation.

tests/integration/tests/suite/examples/payload_processing.rs
nerdalert: Added calendar-route coverage for the MCP classifier example, using a request without client-supplied Mcp-Method / Mcp-Name headers so the test exercises missing: synthesize with real-client behavior. Left the small local helper duplication in place because the helpers live in separate integration modules, and extracting a shared helper would broaden the test utility API beyond this review fix.

tests/integration/tests/suite/mcp.rs
nerdalert: Added integration coverage for mismatch: ignore, batch behavior under on_invalid: continue, and the id field in HeaderMismatch responses.

nerdalert:
> The implementation is well-structured. Parser reuse via parse_json_rpc_value is good. StreamBuffer + EOS gating is the established pattern in this codebase and is used correctly. McpMethod::Other(String) handles forward-compatibility cleanly. Control character guards are applied consistently across all promotion paths. The three-mode design (pass-through / classifier / gateway) is clearly articulated and the PR stays within its stated scope. Test count (34 unit, 12 integration) is good.
> 
> Worth checking chunk boundaries are handled (i think they are).
> 
> Two behavioral bugs?
> 
> 1. Default missing: reject rejects all real MCP traffic
> 
> MissingHeaderBehavior defaults to Reject. Real MCP clients — including FastMCP cited in the PR — do not send Mcp-Method/Mcp-Name headers; those are Praxis-internal conventions, not part of the MCP spec. A proxy deployed with default config will 400 every legitimate MCP request. The PR's own example config, integration test helpers, and the issue praxis-proxy/ai#148 comment all use synthesize or ignore to work around this — which confirms the default is wrong. Strict header enforcement must be opt-in, not opt-out.
> 
> 2. PR summary misrepresents batch rejection behavior
> 
> BatchPolicy::Reject causes parse_json_rpc_value to return Err(UnsupportedBatch). The let Ok(Some(envelope)) pattern treats that error identically to non-MCP input — it falls through to handle_non_mcp. With on_invalid: continue, batches pass through to the backend. The PR summary states "Rejects JSON-RPC batches regardless of on_invalid config" — this is only true when on_invalid: reject (the default). The claim is inaccurate and needs correcting.
> 
> Summary
> 
> 1. Change default missing behavior from Reject to Ignore
> 2. Correct the batch rejection claim in the PR summary
> 3. Add an integration test sending a fragmented body and verifying correct
>    extraction — a tools/call body split across at least two chunks, asserting
>    method and tool name are correctly promoted. A second test should document (or
>    enforce) the behavior when the body exceeds max_body_bytes.

@twghu ty for the review! I think I addressed everything. Some of what I added to address the review was slated for a followup PR to try and keep remotely close to LOC cap. Updating the followup punch list with the following:

Second PR will add on top of this one to finish out `MCP Request Classification And Static Routing`:
  - Required `params.name`/`params.uri` enforcement for `tools/call`, `prompts/get`, and `resources/read`
  - Spurious `Mcp-Name` rejection for nameless methods such as `tools/list`
  - Additional unit and integration tests for the above hardening cases

Thanks!

shaneutt (approved):
There's more we could do here, but this is a great start and we can do more in upcoming iterations. Thank you @nerdalert!

PR #207, opened 2026-05-14
feat: add MCP header validation hardening

### LINKED ISSUE #149: MCP Protocol Support

nerdalert:
This is the second PR for the MCP classifier filter to enforce stricter validation (was supposed to be closer to half but the first PR scope creeped):

- Reject `tools/call`, `prompts/get`, and `resources/read` when the required `params.name` or `params.uri` field is missing from the body (respects `on_invalid: continue` for pass-through pipelines)
- Reject spurious `Mcp-Name` headers on methods that have no body-derived name (e.g. `tools/list` with `Mcp-Name: evil`) under default `mismatch: reject`

  Refs praxis-proxy/ai#149

### DIFF
--- a/filter/src/builtins/http/payload_processing/mcp/mod.rs
+++ b/filter/src/builtins/http/payload_processing/mcp/mod.rs
@@ -153,6 +153,10 @@ impl HttpFilter for McpFilter {
 
         let mcp_envelope = extract_mcp_envelope(&value, method_str, &ctx.request.headers);
 
+        if let Some(action) = reject_missing_required_selector(&mcp_envelope, &envelope, &self.config) {
+            return Ok(action);
+        }
+
         if let Err(action) = validate_mcp_headers(ctx, &mcp_envelope, &envelope, &self.config) {
             return Ok(action);
         }
@@ -221,6 +225,23 @@ fn handle_non_mcp(config: &McpConfig) -> Result<FilterAction, FilterError> {
     }
 }
 
+/// Selector-bearing methods cannot be trusted when the selector is absent or malformed.
+fn reject_missing_required_selector(
+    mcp: &McpEnvelope,
+    envelope: &crate::builtins::http::payload_processing::json_rpc::envelope::JsonRpcEnvelope,
+    config: &McpConfig,
+) -> Option<FilterAction> {
+    let requires = mcp.method.requires_name() || mcp.method.requires_uri();
+    if requires && mcp.name.is_none() {
+        match config.on_invalid {
+            InvalidMcpBehavior::Reject => Some(mcp_invalid_params_rejection(envelope)),
+            InvalidMcpBehavior::Continue => Some(FilterAction::Continue),
+        }
+    } else {
+        None
+    }
+}
+
 /// Validate `Mcp-Method` and `Mcp-Name` headers against body-derived values.
 ///
 /// When `missing: synthesize` is configured and a standard MCP header is
@@ -235,6 +256,14 @@ fn validate_mcp_headers(
 
     if let Some(name) = &mcp.name {
         validate_single_header(ctx, "mcp-name", name, envelope, config)?;
+    } else if ctx.request.headers.get("mcp-name").is_some() {
+        match config.header_validation.mismatch {
+            MismatchBehavior::Reject => {
+                warn!("client sent Mcp-Name header but body method has no name");
+                return Err(mcp_header_mismatch_rejection(envelope));
+            },
+            MismatchBehavior::Ignore => {},
+        }
     }
 
     Ok(())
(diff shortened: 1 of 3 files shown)


filter/src/builtins/http/payload_processing/mcp/tests.rs
twghu, line 1: - missing message on line 708 compare for example line 626

filter/src/builtins/http/payload_processing/mcp/mod.rs
twghu, line 1: Line 228, should the comment be skip requests with methods .. ?

tests/integration/tests/suite/mcp.rs
twghu, line 1: Is this enough tests?

twghu (changes requested):
just some  questions,  see file comments.

filter/src/builtins/http/payload_processing/mcp/mod.rs
```diff
@@ -221,6 +225,19 @@ fn handle_non_mcp(config: &McpConfig) -> Result<FilterAction, FilterError> {
     }
 }
 
+/// Reject or skip methods that require a name/URI when extraction failed.
+fn reject_missing_required_selector(mcp: &McpEnvelope, config: &McpConfig) -> Option<FilterAction> {
+    let requires = mcp.method.requires_name() || mcp.method.requires_uri();
+    if requires && mcp.name.is_none() {
+        match config.on_invalid {
+            InvalidMcpBehavior::Reject => Some(FilterAction::Reject(Rejection::status(400))),
```
shaneutt: On rejection we return a flat 400, no body, but sibling functions like `mcp_header_mismatch_rejection` returns a structured JSON-RPC error body. We need to consistently provide errors, here we should probably return a JSON-RPC `InvalidParams`.

filter/src/builtins/http/payload_processing/mcp/tests.rs
```diff
@@ -607,6 +607,107 @@ async fn synthesize_injects_standard_mcp_headers() {
     );
 }
 
+// -----------------------------------------------------------------------------
+// Required Name/URI Tests
+// -----------------------------------------------------------------------------
+
```
shaneutt, line 613: Need a test for `tools/call` with _no params key at all_ as opposed to empty params.

Also need a negative test for non-string name values.

filter/src/builtins/http/payload_processing/mcp/tests.rs
```diff
@@ -607,6 +607,107 @@ async fn synthesize_injects_standard_mcp_headers() {
...
+async fn spurious_mcp_name_ignored_when_mismatch_ignore() {
+    let filter = make_filter(r#"{"header_validation": {"mismatch": "ignore"}}"#);
+    let body_str = r#"{"jsonrpc":"2.0","id":1,"method":"tools/list"}"#;
+    let req = make_mcp_request(&[("mcp-name", "evil")]);
+    let mut ctx = crate::test_utils::make_filter_context(&req);
+    let mut body = Some(Bytes::from(body_str));
+
+    let action = filter.on_request_body(&mut ctx, &mut body, true).await.unwrap();
+
+    assert!(matches!(action, FilterAction::Release));
```
shaneutt: Project conventions ask for assertion messages, this would be a good place to add one.

tests/integration/tests/suite/mcp.rs
```diff
@@ -225,6 +225,54 @@ fn mcp_standard_headers_preserved_internal_stripped() {
...
+    assert_eq!(
+        parse_status(&raw),
+        400,
+        "spurious Mcp-Name on nameless method should be rejected in default mode"
+    );
+}
+
 // -----------------------------------------------------------------------------
 // Fragmented / Oversized Body Tests
 // -----------------------------------------------------------------------------
```
shaneutt, line 343: We could use more integration tests here that show that the params are getting to the backend properly, with a mock MCP server, etc.

shaneutt (changes requested):

filter/src/builtins/http/payload_processing/mcp/tests.rs
nerdalert: Done ty. Added the missing assertion message on the mismatch-ignore test.

filter/src/builtins/http/payload_processing/mcp/mod.rs
nerdalert: Reworded the selector comment to explain why selector-bearing methods cannot be trusted when the selector is absent or malformed.

tests/integration/tests/suite/mcp.rs
nerdalert: Done ty. Expanded the integration coverage with backend-forwarding proof plus missing/no params and non-string name rejection cases.

filter/src/builtins/http/payload_processing/mcp/mod.rs
nerdalert: Added Missing required MCP selectors now return a InvalidParams error body with the original request ID, instead of a flat 400.

filter/src/builtins/http/payload_processing/mcp/tests.rs
nerdalert: Added coverage for tools/call with no params key and for non-string params.name, the non-string case also exercises string ID escaping.

filter/src/builtins/http/payload_processing/mcp/tests.rs
nerdalert: Added, ty.

tests/integration/tests/suite/mcp.rs
nerdalert: Added an integration test using the echo backend to prove valid MCP params are forwarded upstream unchanged, plus integration coverage for invalid missing/non-string params.

PR #353, opened 2026-05-15
feat: add mcp_gateway static broker filter

nerdalert:
## Summary

  Adds the new `mcp_gateway` filter for MCP gateway mode. This PR implements the static broker/catalog phase and the supporting pipeline infrastructure needed for `mcp_gateway -> load_balancer` pipelines without requiring a `router` filter. Adding a static catalog makes sense for this first gateway PR because it lets us validate mcp_gateway behavior without adding backend discovery, caching, or session initialization yet.

> ⚠️ (yellow alert for attn 😛) Adding a specific callout for brevity. This PR uses a static MCP tool catalog by design: Praxis is not discovering backend tools yet, so configured tools must be normalized into valid MCP tools/list responses. Along with enabling smol PRs, a static catalog lets us validate the gateway contract without depending on live MCP servers (IT should do that as well when the epic is complete).

> - static: configured catalog only
>  - dynamic: discover from backend tools/list
> - hybrid: discover backend tools, then apply config overrides/filtering/prefixes

> MCP requires every tools/list tool to include inputSchema. Since this PR uses a static configured catalog, omitted [inputSchema](https://raw.githubusercontent.com/modelcontextprotocol/modelcontextprotocol/main/schema/2025-03-26/schema.ts) values are treated as no-argument tools and emitted as {"type":"object","additionalProperties":false}. That keeps simple dev/test configs concise while still returning spec-shaped MCP tool objects; we can tighten validation later if we decide production configs should require explicit schemas.
  
  This PR adds:

  - `mcp_gateway` filter registration and config parsing
  - Static configured MCP tool catalog support
  - Short-circuit handling for `initialize`, `tools/list`, `ping`, `notifications/initialized`, and `DELETE`
  - Controlled JSON-RPC `-32601` responses for unsupported MCP methods and `tools/call` until backend routing lands
  - Config validation for public/backend paths, duplicate exposed tool names, invalid schemas, and malformed tool definitions
  - Pipeline validation for `mcp_gateway` as a cluster-selecting filter
  - Conflict detection for `mcp_gateway` and `router` in the same pipeline
  - Cluster cross-reference validation from `mcp_gateway.servers[].cluster` to `load_balancer.clusters[].name`
  - StreamBuffer writeback for body-phase `ctx.cluster`, `ctx.rewritten_path`, and durable metadata
  - Mutated request body forwarding support, including upstream `Content-Length` repair
  - Retry safety for StreamBuffer bodies using the larger of original and mutated body sizes
  - Reusable integration-test body mutator coverage for ReadWrite + StreamBuffer body/path/cluster handoff
  - Example config for static MCP gateway broker/catalog mode

  ## Scope

  This implements static MCP gateway broker behavior only. It does not forward `tools/call` to backend MCP servers yet.

  ## Follow-Up

  A follow-up PR will complete end-to-end MCP gateway routing:

  - `tools/call` catalog lookup by prefixed tool name
  - Backend cluster selection from the selected tool
  - Prefix stripping from `params.name` before forwarding
  - Backend path rewrite to each configured server path
  - Multi-backend routing proof with two echo backends
  - End-to-end body mutation proof that the backend receives the stripped tool name and repaired `Content-Length`
  - Full MCP flow coverage: initialize -> initialized notification -> tools/list -> tools/call -> delete
  - Later in the PR chain for the epic https://github.com/praxis-proxy/ai/issues/148#issuecomment-4368239729 you will see a dynamic discovery, where the gateway calls backend tools/list at startup or on notifications/tools/list_changed, merges the results into the live catalog, and refreshes when backends notify

  ## Test Plan

  - MCP gateway unit tests for config validation and broker responses
  - MCP gateway integration tests for initialize, tools/list, ping, notifications, DELETE, wrong-path 404, unsupported methods, and PR-limited tools/call
  behavior
  - StreamBuffer adapter integration tests for mutated body forwarding, path rewrite writeback, and Content-Length repair
  - Pipeline validation tests for `mcp_gateway -> load_balancer`, router conflict detection, and cluster cross-reference validation

  Refs praxis-proxy/ai#149, praxis-proxy/ai#213, praxis-proxy/ai#212


nerdalert:
@shaneutt Not sure how to get this one coherent and much smaller but not sure how to reopen anymore. Can chunk it post review maybe. The last one for this section (and subsequent agent tasks after this are more chunkable) Ty!

filter/src/builtins/http/ai/agentic/mcp/broker/mod.rs
twghu, line 1: HTTP status codes for JSON-RPC errors

The spec implies a clean three-tier taxonomy. 

  Tier 1 — HTTP transport failures → 4xx correct

  The request never formed a valid JSON-RPC message. No JSON-RPC error response is possible.
  - Body is not valid JSON → 400 ✓
  - Wrong path → 404 ✓
  - Wrong HTTP method → 405 ✓
  - DELETE without Mcp-Session-Id → 400 ✓ (missing required HTTP-level header)

  Tier 2 — JSON-RPC protocol errors → should be HTTP 200

  The JSON parsed fine but the envelope is malformed.
  A JSON-RPC error response can be formed (using null for id when the id is missing/unparseable).
  - parse_json_rpc_value fails (line 653): valid JSON, bad JSON-RPC structure → -32600 in HTTP 200
  - Missing method field (line 657): valid envelope, incomplete request → -32600 in HTTP 200
  - Null/missing id on a request (the invalid_request_action path) → -32600 in HTTP 200 with "id":null

  Tier 3 — Application errors → should be HTTP 200

  Valid JSON-RPC request, method-level outcome.
  - Method not found → -32601 in HTTP 200
  - tools/call stub → -32601 in HTTP 200
  
The three return-400 sites in on_request_body and json_rpc_error_action_with_id conflate two different error tiers. The fix is one-line only at json_rpc_error_action_with_id — the other two should stay as-is.

Line 157 — correct as 400. The body is not valid JSON. No JSON-RPC envelope can be formed, so there is no id to echo back. HTTP 400 is the right transport-level response here.

Lines 161 and 165 — correct as 400. Same reasoning: parse_json_rpc_value failed (not a valid JSON-RPC object) or there is no method field. These are broken at the envelope level before any protocol semantics apply.

Line 375 — should be 200. json_rpc_error_action_with_id is only reached when we have a well-formed JSON-RPC envelope and a recoverable id. At that point the transport succeeded; the error is a protocol-layer result (-32600 invalid request, -32601 method not found, tools/call stub). Per JSON-RPC 2.0, which MCP inherits, those travel inside HTTP 200 responses. Change Rejection::status(400) to Rejection::status(200) on line 375. 

The integration tests asserting 400 for those paths would need updating too.

tests/integration/tests/suite/mcp_broker.rs
twghu, line 1: Missing test?

  #[test]
  fn mcp_gateway_example_serves_prefixed_catalog() {
      // backends are placeholder cluster targets; gateway short-circuits
      // all current methods before forwarding
      let weather_guard = start_backend_with_shutdown("weather");
      let calendar_guard = start_backend_with_shutdown("calendar");
      let proxy_port = free_port();

      let config = load_example_config(
          "payload-processing/mcp-gateway.yaml",
          proxy_port,
          HashMap::from([
              ("127.0.0.1:3001", weather_guard.port()),
              ("127.0.0.1:3002", calendar_guard.port()),
          ]),
      );
      let proxy = start_proxy(&config);

      let raw = http_send(proxy.addr(), &json_post("/mcp",
          r#"{"jsonrpc":"2.0","id":1,"method":"tools/list"}"#));
      assert_eq!(parse_status(&raw), 200, "tools/list should return 200");
      let body = parse_body(&raw);
      assert!(body.contains("weather_get_weather"), "weather tool with prefix");
      assert!(body.contains("weather_forecast"),    "weather forecast with prefix");
      assert!(body.contains("cal_create_event"),    "calendar tool with prefix");
      assert!(body.contains("cal_list_events"),      "calendar list with prefix");
      assert!(body.contains("\"city\""),             "inputSchema from example config");
  }

twghu:
mcp_gateway is placed at filter/src/agentic/ rather than under filter/src/builtins/http/ai/agentic/ where the existing mcp and json_rpc modules live ? 
  
It seems that the full path qualification of x::y::z would be sufficient for identifying module coordinates,  is the there value in naming change or at least more qualified comment.
  
  From the tree:
  
  filter/src/agentic/                                     ← crate root level
  filter/src/builtins/http/ai/agentic/         ← inside builtins

  In Rust module paths:
  - crate::agentic
  - crate::builtins::http::ai::agentic

Both have nearly identical doc comments ("Agentic protocol filters: …") and both contain MCP-related code. A contributor grepping for agentic or navigating the tree hits two results with no obvious signal for which one to look in or add to.

examples/README.md
twghu, line 1: Missing links?

  What the PR added (lines 45–47 under Payload Processing):
  configs/payload-processing/ai-inference-body-based-routing.yaml  ← doesn't exist
  configs/payload-processing/json-rpc-routing.yaml                  ← doesn't exist
  configs/payload-processing/mcp-classifier-routing.yaml            ← doesn't exist

  What already existed (lines 125–128 under AI / Inference):
  configs/ai/ai-inference-body-based-routing.yaml  ✓ exists
  configs/ai/json-rpc-routing.yaml                 ✓ exists
  configs/ai/mcp-classifier-routing.yaml           ✓ exists
  
  The mcp-gateway.yaml entry at line 48 is the only one that's both new and correct — the file exists at
  configs/payload-processing/mcp-gateway.yaml.

  The other three are duplicates with broken paths.

filter/src/agentic/mcp_gateway/config.rs
```diff
@@ -0,0 +1,361 @@
...
+
+/// Matches literal `..` and percent-encoded variants (`%2e%2e`, `.%2e`,
+/// `%2e.`). Mirrors the runtime check in `apply_rewritten_path` so bad
+/// backend paths fail at config load, not as runtime errors.
+#[allow(clippy::indexing_slicing, reason = "bounds checked by i + 2 < b.len()")]
+fn is_traversal_segment(seg: &str) -> bool {
+    if seg == ".." {
+        return true;
+    }
+    let mut dots = 0u8;
```
twghu: Theoretical overflow
 
A segment with exactly 256 encoded dots wraps to dots == 0 → not detected (false negative, not a security issue since 256 dots ≠ ..). A segment of exactly 258 dots wraps to dots == 2 → flagged as traversal (false positive). Use u16 or add an early if dots > 2 { return false;}:

  let mut dots = 0u16;
  // or:
  if dots > 2 { return false; }

filter/src/builtins/http/ai/agentic/mcp/broker/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        let Some(ref method_str) = envelope.method else {
+            return Ok(FilterAction::Reject(Rejection::status(400)));
+        };
+
+        if !contains_control_chars(method_str) {
+            ctx.set_metadata("json_rpc.method", method_str.clone());
+            ctx.set_metadata("mcp.method", method_str.clone());
+        }
+
+        dispatch_method(ctx, &self.catalog, &value, &envelope, method_str)
```
twghu, line 189: Can accept tools/list, ping, etc. without requiring Mcp-Session-Id?

on_request()        → validates path + method, handles DELETE
on_request_body()   → parses JSON, extracts envelope + method string
dispatch_method()   → routes to initialize / tools_list / ping / notification / etc.

If session enforcement "should be" added in a follow-up to dispatch_method() after the envelope is parsed so you  know the method, and you can skip the check for initialize itself (which is what creates the session) but require it for everything else.

filter/src/builtins/http/ai/agentic/mcp/broker/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+/// Does not initialize backends — that belongs to follow-up backend session work.
+fn handle_initialize(
+    ctx: &mut HttpFilterContext<'_>,
+    value: &serde_json::Value,
+    envelope: &JsonRpcEnvelope,
+) -> FilterAction {
+    if let Some(version) = value
+        .get("params")
+        .and_then(|p| p.get("protocolVersion"))
+        .and_then(|v| v.as_str())
```
twghu, line 300: Not a rfc just a drive by comment:

Do we ever want to support versioning of protocols ?

twghu (changes requested):
This PR adds the first phase of an MCP gateway filter: a static tool catalog broker that short-circuits initialize, tools/list, ping, notifications/initialized, and DELETE. tools/call returns a controlled -32601 stub.

The supporting infrastructure includes StreamBuffer body-mutation writeback (rewritten_path, Content-Length repair), pipeline validation for mcp_gateway → load_balancer without a router, and cluster cross-reference validation extended to cover mcp_gateway.servers[].cluster.
  
The scope is clearly declared and the PR is internally consistent. Issues below are organized by severity.

Some nits, some functional questions.

filter/src/builtins/http/ai/agentic/mcp/broker/mod.rs
nerdalert: Fixed and ty! JSON-RPC protocol/application errors now return HTTP 200; malformed JSON/envelope, wrong path/method, and DELETE without session remain transport-level 4xx.

tests/integration/tests/suite/mcp_broker.rs
nerdalert: oof, good catch! Added the example-config integration test and verified the prefixed catalog plus city schema.

examples/README.md
nerdalert: Removed the broken duplicate payload-processing links; kept only the new mcp-gateway.yaml entry there.

filter/src/agentic/mcp_gateway/config.rs
nerdalert: Fixed by widening the dot counter to u16, and mirrored that in the runtime path rewrite helper.

filter/src/builtins/http/ai/agentic/mcp/broker/mod.rs
nerdalert: Agreed. Session enforcement should happen after envelope parsing in dispatch_method, skipping initialize. Adding that to the PR stack for a followup, appreciate it.

filter/src/builtins/http/ai/agentic/mcp/broker/mod.rs
nerdalert: There is a spike opened for it! https://github.com/praxis-proxy/praxis/issues/366

filter/src/agentic/mcp_gateway/config.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+    Ok(())
+}
+
+/// Matches literal `..` and percent-encoded variants (`%2e%2e`, `.%2e`,
+/// `%2e.`). Mirrors the runtime check in `apply_rewritten_path` so bad
+/// backend paths fail at config load, not as runtime errors.
+#[allow(clippy::indexing_slicing, reason = "bounds checked by i + 2 < b.len()")]
+fn is_traversal_segment(seg: &str) -> bool {
+    if seg == ".." {
```
shaneutt: Wait I think we're doing traversal check in two places

filter/src/builtins/http/ai/agentic/mcp/broker/config.rs
```diff
@@ -0,0 +1,361 @@
+// SPDX-License-Identifier: MIT
```
shaneutt, line 1: We shouldn't call this `mcp_gateway` probably just `mcp`? Also this needs to be moved under `builtins/ai/agentic`

filter/src/agentic/mcp_gateway/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        .is_some()
+    {
+        FilterAction::Reject(Rejection::status(204))
+    } else {
+        FilterAction::Reject(Rejection::status(400))
+    }
+}
+
+/// Generates a new gateway session and returns MCP capabilities.
+/// Does not initialize backends — that belongs to follow-up backend session work.
```
shaneutt:
```suggestion
/// Does not initialize backends, that belongs to follow-up backend session work.
```

filter/src/agentic/mcp_gateway/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    let id_json = format_id_json(envelope);
+    let session_id = generate_session_id();
+
+    debug!(session_id_len = session_id.len(), "gateway initialize");
+    ctx.set_metadata("mcp.session_id", session_id.clone());
+
+    let response_body = format!(
+        r#"{{"jsonrpc":"2.0","id":{id_json},"result":{{"protocolVersion":"2025-03-26","capabilities":{{"tools":{{"listChanged":false}}}},"serverInfo":{{"name":"praxis-mcp-gateway","version":"0.1.0"}}}}}}"#,
+    );
```
shaneutt: A few problems here:

We've added this "mcp-gateway" language everywhere, but that's a misnomer with Praxis. We need to change this framing throughout the PR: just refer to Praxis as Praxis. Yes it has MCP protocol support, but that is a feature not a fundamental to its identity.

Also, this hardcodes the version, among other information.

We should clean this up: don't do it as one big string, and use dynamic values where applicable.

filter/src/agentic/mcp_gateway/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+// -----------------------------------------------------------------------------
+// Session ID
+// -----------------------------------------------------------------------------
+
+/// Generate a cryptographically random gateway session ID.
+fn generate_session_id() -> String {
+    let bytes: [u8; 16] = rand::random();
+    let hex: String = bytes.iter().map(|b| format!("{b:02x}")).collect();
+    format!("gw-{hex}")
+}
```
shaneutt: For now let's assume we want these to be cryptographically _secure_ as well as random. You could use `OsRng` here to make that more explicit.

filter/src/pipeline/checks.rs
```diff
@@ -106,6 +111,21 @@ pub(super) fn check_open_security_filters(
     }
 }
 
+/// `router` cannot coexist with body-phase cluster selectors because it
+/// runs later and unconditionally overwrites `ctx.cluster`.
```
shaneutt: I don't think this should work quite like this.

It may be reasonable in the future to have router run _prior_ to mcp routing, to reduce the selectable clusters or something like that, in addition to other variations.

Trying to deepen how we validate pipelines in this way may lead to trouble later, perhaps we should consider moving the decision point into the filters themselves, and give them a way to pass messages to modify their results? Something I'm going to think on. 🤔

filter/src/pipeline/checks.rs
```diff
@@ -106,6 +111,21 @@ pub(super) fn check_open_security_filters(
...
+pub(super) fn check_conflicting_cluster_selectors(names: &[&str], errors: &mut Vec<String>) {
```
shaneutt: This is broken in its current state, because it doesn't account for conditions which might actually result in no conflict.

protocol/src/http/pingora/handler/request_filter/stream_buffer.rs
```diff
@@ -153,8 +155,8 @@ pub(super) async fn pre_read_body(
             .execute_http_request_body(&mut filter_ctx, &mut body, end_of_stream)
             .await;
 
-        ctx.request_body_bytes = filter_ctx.request_body_bytes;
```
shaneutt, line 156: This removes accumulation in pre-read, wont this break things?

shaneutt (changes requested):

filter/src/pipeline/checks.rs
twghu: I think that the MCP-idiomatic answer is that it should never need to set ctx.cluster in the first place.

Should a pipeline have a way for a filter to say "set cluster to X only if I handled this   request."

The check doesn't verify ordering and ordering implies a code path we don't need to process.

The current design coordinates inter-filter behavior through shared mutable context (ctx.cluster), and then appears to catch broken combinations statically. But static name-based checks can't capture intent — they can ban or log  configurations.

In short, we probably don't want to act on  routing info inter filter for it to be possibly actionable.  So  does route make sense in this content?

Unless I've misunderstood this, a filter shouldn't be in the routing business and probably actions on routing based on context could lead to unwelcome complications.

Then if a filter isn't in the routing business why add the check.

filter/src/pipeline/checks.rs
twghu: Agree.

filter/src/agentic/mcp_gateway/config.rs
nerdalert: Consolidated traversal detection into a shared has_dot_dot_traversal helper, and now both MCP path validation and upstream rewritten-path validation use the same implementation.

filter/src/builtins/http/ai/agentic/mcp/broker/config.rs
nerdalert: Done, thanks, I should have flagged that in the PR notes.

filter/src/agentic/mcp_gateway/mod.rs
nerdalert: Cleaned up the MCP-gateway framing so this refers to Praxis and MCP static catalog behavior instead. The initialize response is now built as structured JSON and uses dynamic crate version data where applicable.

filter/src/agentic/mcp_gateway/mod.rs
nerdalert: Done. Session IDs now use OsRng with TryRngCore

filter/src/pipeline/checks.rs
nerdalert: Thanks guys. Context on the thoughts. Did an e2e prototype where the MCP gateway was already forwarding tools/call to real backends. In that version, the MCP code picked the backend itself, then handed that choice to the load balancer. A router in the same pipeline could overwrite that backend.

This PR no longer does that forwarding. It only answers static catalog/broker methods locally and returns controlled unsupported for tools/call, so it does not currently choose a backend or set ctx.cluster. 

Will take the feedback and rethink the e2e validation, for reference: https://github.com/praxis-proxy/praxis/issues/24#issuecomment-4368239729

filter/src/pipeline/checks.rs
nerdalert: Backed out the change, it no longer does backend forwarding or sets ctx.cluster. Ty.

protocol/src/http/pingora/handler/request_filter/stream_buffer.rs
nerdalert: Ty! Checked the Pingora body path as well. Pingora still has a fixed retry buffer, but it also has a specific path for bodies that were already consumed before forwarding: it gives request_body_filter() one terminal callback so the proxy implementation can emit its buffered body upstream.

In Praxis StreamBuffer mode, Praxis pre-read can run the body filters while reading chunks, then run them again at EOS with one full combined body. If we copy filter_ctx.request_body_bytes back directly, we can accidentally count those bytes twice.

Changed this so pre-read tracks the original bytes from the client separately as they arrive. At the end, ctx.request_body_bytes still reflects the real original body size, while ctx.mutated_request_body_len separately tracks the final forwarded body size if a body-writing filter changed the payload. That keeps retry/body-size logic correct without double-counting.

  Changes:

- Added an explicit original_body_bytes counter in stream_buffer.rs.
- Count original downstream chunks before creating the synthetic full-body EOS pass.
- Seed/write ctx.request_body_bytes from that original counter during pre-read.
- Keep ctx.mutated_request_body_len for the final forwarded body length when body-writing filters mutate the payload.
- Did not restore the old direct ctx.request_body_bytes = filter_ctx.request_body_bytes line because that can double-count in this pre-read path.

Example Flow: 

 1. Client sends request body in chunks.
2. Praxis StreamBuffer pre-reads those chunks before forwarding upstream.
3. While reading, Praxis counts the real client bytes into original_body_bytes.
4. At EOS, Praxis combines the chunks into one full body.
5. Praxis runs body filters on that full body so they can inspect or rewrite it.
6. Pingora later calls request_body_filter() once because the body was already consumed before forwarding.
7. Praxis uses that callback to forward the buffered body from ctx.pre_read_body.
8. ctx.request_body_bytes stays as the original client body size.
9. If filters changed the body, ctx.mutated_request_body_len stores the final forwarded size.

shaneutt (approved):
Approved once open comments are resolved and closed, and once @twghu's review is complete. Most of these changes are very isolated, and it's OK to expecte we'll need to do some additional iterations before it's complete.

Thanks @nerdalert!

PR #367, opened 2026-05-18
docs: add stateful proxy state management proposal

### LINKED ISSUE #146: Spike: Stateful Proxy Analysis
The purpose of this spike is to analyze what state management we will need for Praxis given our current scope and goals.

The result of this spike should be a proposal discussion created for us to build consensus on before we put the plan into motion.

The proposal should provide a table of all the state and related features that we need to support here, and both local and external mechanisms for storing that state.

### LINKED ISSUE #215: Preserve stateful filter state across hot-reload pipeline rebuilds
## Summary

  Pipeline hot-reload currently rebuilds stateful filter instances, which resets
  local state (rate limiter buckets, circuit breaker state, etc.) while the
  process continues serving traffic. This is acceptable for local protection
  mechanisms, but not for correctness-critical quota, session, or task state.

  ## Current Behavior

  When the config file changes and a pipeline reload is triggered, all filter
  instances are rebuilt from scratch. Any in-memory state held by those filters
  (e.g. rate limiter counters, circuit breaker trip counts) is lost. Traffic
  continues to be served with reset state, which can allow bursts past rate
  limits or re-enable tripped circuit breakers.

  ## Desired Behavior

  Stateful filters should be able to preserve or recover their state across
  hot-reload events. Possible approaches include:

  - Migrating local state from old filter instances to new ones during swap
  - Externalizing state that must survive reloads (e.g. to a shared backend)
  - Providing a lifecycle hook for filters to snapshot and restore state

  ## Context

  Identified during the [stateful proxy state management spike](https://github.com/praxis-proxy/ai/issues/146).
  See proposal [#367](https://github.com/praxis-proxy/praxis/pull/367) for
  the broader state model discussion. As shared hot-path state moves to
  external backends (e.g. Valkey), this becomes less critical for those
  features — but local-only state (circuit breakers, health snapshots,
  local caches) will still benefit from reload-safe handling.

nerdalert:
## Summary

  This PR adds proposal `00099_stateful-proxy-state-management.md` for issue praxis-proxy/ai#146.

  The proposal summarizes the stateful proxy spike into the initial proposal format:

  - Defines the state classes Praxis should use consistently.
  - Separates request-local facts, local runtime state, shared hot-path state, durable business state, configuration state, and externalized decisions.
  - Positions Redis/Valkey as the first shared hot-path backend, not as a universal store.
  - Recommends typed state APIs over exposing a raw global KV API as the primary filter-facing abstraction.
  - Calls out local-state limits, hot-reload reset behavior, failure modes, TTLs, key rules, and bounded observability as first-class concerns.

  ## Spike Research

  The background research and implementation notes are here:

  https://github.com/nerdalert/praxis-research-spikes/tree/main/stateful-proxy

  That spike includes the detailed state inventory, local vs external storage analysis, comparable proxy patterns, and proposed implementation slices.

  ## Process Note

  This is intentionally a first proposal PR with only `What?` and `Why?`.

  It does not include a `How?` section. The detailed requirements and design should be added in a follow-up proposal update after the direction is accepted.

  ## Follow-Up Direction

  The spike recommends starting with:

  - State model and key convention docs.
  - Bounded local state utilities.
  - Typed state traits, starting with rate limiting.
  - State operation metrics.
  - Redis/Valkey backend support after the trait and local behavior are reviewed.
  - Later typed stores for token ledgers, protocol sessions, task ownership, policy caches, routing snapshots, cache indexes, and usage event export.


docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+---
+
+# Stateful Proxy State Management
+
+## What?
+
+Praxis should define a project-wide state model before
+stateful AI gateway, routing, quota, protocol, and
+observability features each grow their own storage
+patterns. The model should distinguish request-local
```
shaneutt, line 16: 💯

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+facts, local runtime state, shared hot-path state,
+durable business state, configuration state, and
+externalized decision state.
```
shaneutt:
```suggestion
externalized decision state. It must provide optionality
for storage (e.g. local in-memory or disk, vs external kv-store, etc).
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+The spike for issue 99 found that Praxis already has
+several local state patterns: request metadata, local
+load-balancer counters, circuit breaker state, health
+snapshots, hot-reloaded config snapshots, and filter-owned
+maps. The per-IP rate limiter is the clearest existing
+example: it uses a local `DashMap` with a `100_000` entry
+soft cap and `200_000` entry hard cap to prevent unbounded
+growth. These patterns are useful, but they are not enough
+for features that must remain correct across multiple
+Praxis replicas.
```
shaneutt, line 31: 👍

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+
+This proposal establishes the direction that stateful
+features should use explicit storage classes and typed
```
shaneutt:
```suggestion
features must use explicit storage classes and typed
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+domain APIs instead of exposing a raw global key-value API
+as the primary filter-facing abstraction. Local in-memory
```
shaneutt, line 36: 👍

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+implementations should be available for tests, development,
+demos, and explicitly single-replica deployments. Shared
+hot-path implementations should use a Redis/Valkey
+compatible backend first, with strict timeouts, TTLs, key
+conventions, failure-mode defaults, and bounded metrics
+labels.
```
shaneutt: There are situations where local storage could in fact be sufficient for HA deployments (e.g. optional optimization data where latency is extremely high priority, per-site metrics, etc). There's no need to _exclude_ them from production scenarios. However I think your point is that they **should not be the default**, if so I strongly agree.

We should provide guidance for what those scenarios are in documentation later. For this proposal I suggest we shift the language some:


```suggestion
implementations should be available for tests, development,
demos, explicitly single-replica deployments, and some niche
HA scenarios. The default production guidance will be that
implementations must use an external (e.g. Valkey) storage
backend first, with strict timeouts, TTLs, key conventions,
failure-mode defaults, and bounded metrics labels.
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+- Require every shared hot-path state operation to define
+  timeout behavior, fail-open or fail-closed semantics,
+  key schema, TTL, cardinality bounds, and metrics.
+- Keep durable billing, audit, certificate, object,
+  vector, and control-plane state out of synchronous
+  request processing unless a feature explicitly justifies
+  that cost.
+- Provide an incremental implementation path that starts
+  with documentation and bounded local primitives before
+  adding shared backends.
```
shaneutt, line 71:
```suggestion
  adding shared backends.
- Enforce that no built-in filters outright fail without internal storage. Filters
  must tolerate lack of storage whenever possible. An explicit opt-out will be
  needed for any filter that is purely bound on external storage solutions.
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+
```
shaneutt, line 75: We need to emphasize metrics further here. Praxis metrics themselves are already worth their weight in gold, but because of the architecture of Praxis end-users are quite able to construct and manage their own metrics (e.g. for their custom filters/probes).


```suggestion
- Provide a complete experience for the storage and retrieval of metrics per-site, with
  optional and customizable storage solutions. This covers built-in metrics (core proxy,
  built-in filters, etc) but also extensions (custom filters).
```

There's probably a better way to word that out, consider it a starting point if nothing else.

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+| Intelligent routing and llm-d/GIE | Routing depends on fresh backend pressure, readiness, cost, and scheduler signals. | Local validated snapshots with freshness limits and safe fallback behavior. |
+| Response and semantic caching | Cache entries, fill locks, vector refs, and purge state have cache-specific semantics. | Separate cache APIs; do not overload quota/session stores as a generic cache. |
+| Auth, RBAC, guardrails, and policy | Security decisions may be expensive, cached, audited, or externally delegated. | Request-local decision facts, bounded caches, policy-version keys, and fail-closed defaults for enforcement. |
+| Retry, hedging, and failover | Attempts and budgets must avoid loops, amplification, and confusing usage records. | Keep attempt state request-local first; add shared fleet budgets only when needed. |
+| Observability | Metrics, logs, traces, and exporters aggregate state and can create cardinality risk. | Keep exporter state bounded and never label on raw request, user, session, task, or prompt values. |
+
+### Non-Goals
+
+- Do not make Praxis a database.
+- Do not require every stateful feature to use Redis.
```
shaneutt: Not that we wouldn't integrate with Redis, but due to the community friendliness of Valkey let's use it as our common example.


```suggestion
- Do not require every stateful feature to use Valkey.
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+- Do not expose a raw global mutable key-value API as the
+  primary filter-facing abstraction.
```
shaneutt, line 106: </img>

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+- Do not put SQL, Kubernetes API, object-store, vector
+  database, or other slow control-plane calls directly on
+  every request path.
```
shaneutt, line 109: This one is kinda interesting, because _yes_ it's a good general goal. But if someone deploys semantic cache and marks it as required for every request...

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+- Do not store prompts, API keys, tokens, tool arguments,
+  or PII in shared-state keys.
+- Do not block MCP, A2A, MaaS, routing scorer, or cache
+  work on implementing every state class at once.
+
+## Why?
+
+### Motivation
+
+Praxis is becoming an AI-native proxy framework, not just
```
shaneutt:
```suggestion
Praxis is an AI-native proxy framework, not just
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+a stateless reverse proxy. Planned features need decisions
+based on facts produced before, during, and after a
+request: parsed request bodies, streaming response usage,
+tenant identity, token budgets, backend pressure, protocol
+session IDs, task IDs, guardrail findings, retry attempts,
+and dynamic configuration versions.
+
+Without a shared state model, each feature will be tempted
+to solve this locally. That creates predictable production
```
shaneutt, line 129: 💯

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+risks: unbounded memory maps, duplicated backend clients,
+incompatible Redis key formats, hidden fail-open security
+paths, hot-reload state loss, high-cardinality metrics, and
+test coverage that passes in a single process but fails
+across multiple replicas.
```
shaneutt, line 134: 👍

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+
+Hot reload is a concrete example of the problem. Today,
+pipeline reload rebuilds stateful filter instances, so
+local rate limiter and circuit breaker state resets while
+the process continues serving traffic. That behavior is
+acceptable for local protection, but not for
+correctness-critical quota, session, or task state.
```
shaneutt, line 141: 👍 

However, we should actually get this reported as a request for improvement, I thought I had an issue for this somewhere but I can not find it.

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+into one storage layer. Request facts should stay local to
+the request. Local runtime state should be fast,
+bounded, and disposable. Shared hot-path state should be
+reserved for correctness-sensitive counters, ledgers,
+sessions, and ownership maps. Durable business records
+should flow through async sinks or product systems.
+Configuration should remain a validated snapshot from the
+config or control plane.
+
+Redis/Valkey is a good first shared hot-path backend
```
shaneutt:
```suggestion
Valkey is a good first shared hot-path backend
```

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+  and metric behavior so that every filter does not create
+  a different operational surface.
+- As a Praxis developer, I want typed state traits for
+  rate limits, token ledgers, sessions, task ownership, and
+  usage events so that filters do not hand-roll storage and
+  key semantics.
+- As an AI gateway operator, I want token counts and usage
+  facts to flow from request and response processing into
+  quota enforcement and billing export without turning
+  request metadata into the ledger of record.
```
shaneutt, line 182: 👍

shaneutt (dismissed):
An excellent first pass at the goals and motivation describing what we want to do, and why we want to do it 👍 

Most of my comments are minor, once resolved LGTM

docs/proposals/00099_stateful-proxy-state-management.md
```diff
@@ -0,0 +1,179 @@
...
+- Centralize shared backend configuration so filters do
+  not create independent Redis clients, key formats,
+  timeouts, or failure behavior.
+- Require every shared hot-path state operation to define
+  timeout behavior, fail-open or fail-closed semantics,
+  key schema, TTL, cardinality bounds, and metrics.
+- Keep durable billing, audit, certificate, object,
+  vector, and control-plane state out of synchronous
+  request processing unless a feature explicitly justifies
+  that cost.
```
twghu, line 65: Generally, and likely this is covered somewhere else, I'll mention here for completeness.

Monitoring degradation of filter performance relating to storage or network etc and provide supportable metrics.

If we can say "X" is operating below acceptable configurable guards then we have the capacity to control graceful degradation.   While degradation affects latency RTT identifying trends can lead to faster isolation of cause.

Ability to fail gracefully while attempting to recover.

Support on this is going to be complex, any metrics and data routes from point to point will save time in determining performance issues.  These metrics and route info through for example A2A all help to inform the identification of problems.

In the case of increasing storage degradation as an example, a separate administration point from Praxis, the monitoring and raising of alerts for attention and possible automatic rectification will be critical.

twghu:
Generally this is very sound.  

"Praxis should define a project-wide state model before
stateful AI gateway, routing, quota, protocol, and
observability features each grow their own storage
patterns"  

This is probably a good time to call out the value of a data flow chart with links and nuances that can an aid to referencing.

A non-goal "do not put slow calls on every request path" — what if an operator marks semantic cache as required per-request, that's exactly  what happens. 

The proposal could identify an exception for caching (which has different latency characteristics than SQL or  Kubernetes API calls) or acknowledge this tension explicitly. 

While not completely orthogonal to the proposal it is crucial that visibility is available for degraded or suspected evolving restrictions.   Support folks can be chasing figments and spend hours in the wrong location if we don't have patterns in place to not only surface "degradation" but "sub-optimal" configuration points.

twghu (dismissed):
Solid work.

docs/proposals/00099_stateful-proxy-state-management.md
nerdalert: Opened! https://github.com/praxis-proxy/praxis/issues/371

docs/proposals/00099_stateful-proxy-state-management.md
nerdalert: Thanks! Added a user story covering configurable performance guards, trend detection, and graceful degradation with recovery.

docs/proposals/00099_stateful-proxy-state-management.md
nerdalert: LOOOOOOOL 🐐

docs/proposals/00099_stateful-proxy-state-management.md
nerdalert: Truth.

shaneutt (approved):

shaneutt (commented):
<deleted>

shaneutt (approved):

PR #372, opened 2026-05-20
Add responses_format classifier filter for AI API body-aware routing

nerdalert:
## Summary

  This is the first in a stacked implementation for Epic praxis-proxy/ai#93, “Responses API and Agentic Loop Orchestration.” It establishes the body-aware routing foundation that later PRs will build on for Responses orchestration, state, tool calls, and streaming. This is long but want to add some framing here.

  - Adds `responses_format` built-in HTTP filter that classifies request bodies as Responses API (`input`), Chat Completions (`messages`), unknown JSON, invalid JSON, or non-JSON
  - Promotes classification facts (`format`, `model`, `stream`, `store`, `previous_response_id` presence, `conversation` presence) to internal routing headers
  (`x-praxis-ai-*`), durable metadata, and filter results
  - Carries body-derived `filter_results` from pre-read body inspection into request-phase branch evaluation so `on_result` can route using classifier output
  - Does not mutate the request body: uses `ReadOnly` access with `StreamBuffer`
  - Bounds body-derived promoted values and suppresses control-character values before header/result promotion
  - `on_invalid` defaults to `continue` for mixed-traffic listeners; configurable to `reject` for AI-only listeners

 ## Architectural callouts worth mentioning:

- This PR lays the pre-upstream classification and policy-routing groundwork for a future agentic Responses orchestrator, while keeping the actual model/tool.
  loop out of branch re-entry because that lifecycle needs response-aware orchestration.
  - This PR is the routing/classification foundation, not the agent loop itself. It intentionally stops at request-body classification and promotion so later stacked PRs can build orchestration, state, tool execution, and streaming on top.
  - Body-derived filter_results now cross from the pre-read body phase into request-phase branch evaluation. That is an architectural change to the Pingora request context, not just a filter-local change.
  - The filter uses StreamBuffer + ReadOnly: Praxis buffers enough body to classify it, but does not mutate or rewrite the body. That keeps this PR safe for pass-through routing.
  - on_invalid: continue is the default to support mixed-traffic listeners. on_invalid: reject is intended for AI-only listeners and now rejects valid JSON that is not Responses or Chat Completions.
  - Promotion is deliberately split across three surfaces:
    x-praxis-ai-* internal headers for router matching, durable metadata for later lifecycle phases, and filter_results for branch-chain decisions.
  - Body-derived promoted values are bounded/sanitized before header/result promotion. This avoids turning model names or other request fields into oversized or unsafe synthetic headers.
  - This PR relies on existing internal header hygiene: x-praxis-ai-* headers are usable inside Praxis for routing but stripped before upstream.
 
  ## Test plan

  - [x] Unit: Responses with string input classified as `responses`
  - [x] Unit: Responses with item-array input classified as `responses`
  - [x] Unit: Chat Completions with `messages` classified as `chat_completions`
  - [x] Unit: Unknown JSON classified as `unknown`
  - [x] Unit/integration: Unknown JSON rejects when `on_invalid: reject`
  - [x] Unit: Non-JSON continues/rejects according to `on_invalid`
  - [x] Unit/integration: Invalid JSON rejects when `on_invalid: reject`
  - [x] Unit: All extracted facts promoted to headers, durable metadata, and filter results
  - [x] Unit: Missing optional facts are not promoted
  - [x] Unit/integration: Oversized and control-character `model` values are not promoted
  - [x] Integration: Route by classifier result (`responses` vs `chat_completions` vs default)
  - [x] Integration: Route by promoted model header
  - [x] Integration: Route by promoted stream header
  - [x] Integration: Route by body-derived `filter_results` via `on_result`
  - [x] Integration: Reserved `x-praxis-ai-*` headers stripped before upstream
  - [x] Integration: Body byte-for-byte unchanged after classification
  - [x] Integration: Large body over 64 KiB classified and forwarded
  - [x] Schema: Example config parses (`all_example_configs_parse`)
  - [x] Functional: Example config routes Responses, Chat Completions, and unknown traffic
  - [x] Full suites: filter, protocol, integration, schema, and nightly fmt pass

  Issues: praxis-proxy/ai#220, partial praxis-proxy/ai#94, part of praxis-proxy/ai#93

### DIFF
--- a/examples/configs/ai/openai/responses/format-routing.yaml
+++ b/examples/configs/ai/openai/responses/format-routing.yaml
@@ -1,0 +1,56 @@
+# Responses Format Routing
+#
+# Routes AI API traffic by detected body format. The `responses_format`
+# filter parses JSON request bodies and classifies them as `responses`
+# (Responses API), `chat_completions` (Chat Completions API), or
+# `unknown` (valid JSON but unrecognized format).
+#
+# Classification facts are promoted to internal routing headers
+# (x-praxis-ai-*) that the router matches for cluster selection.
+# These internal headers are stripped before upstream by header hygiene.
+#
+# Non-JSON, invalid JSON, and unrecognized JSON bodies continue by
+# default so mixed-traffic listeners do not reject non-AI requests.
+# Set `on_invalid: reject` for listeners that only accept AI API
+# traffic (rejects anything without `input` or `messages`).
+
+listeners:
+  - name: ai-gateway
+    address: "127.0.0.1:8080"
+    filter_chains: [ai-routing]
+
+filter_chains:
+  - name: ai-routing
+    filters:
+      - filter: responses_format
+        on_invalid: continue
+        max_body_bytes: 67108864
+        headers:
+          format: x-praxis-ai-format
+          model: x-praxis-ai-model
+          stream: x-praxis-ai-stream
+
+      - filter: router
+        routes:
+          - path: "/v1/responses"
+            headers:
+              x-praxis-ai-format: "responses"
+            cluster: "responses-backend"
+          - path: "/v1/chat/completions"
+            headers:
+              x-praxis-ai-format: "chat_completions"
+            cluster: "chat-backend"
+          - path_prefix: "/"
+            cluster: "default-backend"
+
+      - filter: load_balancer
+        clusters:
+          - name: "responses-backend"
+            endpoints:
+              - "127.0.0.1:3001"
+          - name: "chat-backend"
+            endpoints:
+              - "127.0.0.1:3002"
+          - name: "default-backend"
+            endpoints:
+              - "127.0.0.1:3003"
--- a/filter/src/builtins/http/ai/openai/responses/classify.rs
+++ b/filter/src/builtins/http/ai/openai/responses/classify.rs
@@ -1,0 +1,429 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Pure request body classifier for Responses API versus Chat Completions.
+
+// -----------------------------------------------------------------------------
+// AiRequestFormat
+// -----------------------------------------------------------------------------
+
+/// Classified request body format.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AiRequestFormat {
+    /// `OpenAI` Responses API (has `input` field).
+    Responses,
+    /// Chat Completions API (has `messages` field).
+    ChatCompletions,
+    /// Valid JSON but neither recognized format.
+    UnknownJson,
+    /// Body is not valid JSON.
+    InvalidJson,
+    /// Body is empty or absent.
+    NonJson,
+}
+
+impl AiRequestFormat {
+    /// Stable string representation for headers, metadata, and filter results.
+    pub(crate) fn as_str(self) -> &'static str {
+        match self {
+            Self::Responses => "responses",
+            Self::ChatCompletions => "chat_completions",
+            Self::UnknownJson => "unknown",
+            Self::InvalidJson => "invalid_json",
+            Self::NonJson => "non_json",
+        }
+    }
+}
+
+// -----------------------------------------------------------------------------
+// ClassifiedRequest
+// -----------------------------------------------------------------------------
+
+/// Extracted facts from a classified request body.
+#[derive(Debug)]
+pub(crate) struct ClassifiedRequest {
+    /// Detected body format.
+    pub format: AiRequestFormat,
+    /// Extracted `model` field value, if present.
+    pub model: Option<String>,
+    /// Extracted `stream` field value, if present.
+    pub stream: Option<bool>,
+    /// Extracted `store` field value, if present.
+    pub store: Option<bool>,
+    /// Extracted `background` field value, if present.
+    pub background: Option<bool>,
+    /// Whether `previous_response_id` is present and non-null.
+    pub has_previous_response_id: bool,
+    /// Whether `conversation` is present and non-null.
+    pub has_conversation: bool,
+}
+
+// -----------------------------------------------------------------------------
+// Classification
+// -----------------------------------------------------------------------------
+
+/// Classify a request body and extract routing facts.
+///
+/// This function is pure: no I/O, no side effects, no mutation of
+/// the input bytes.
+///
+/// ```ignore
+/// let body = br#"{"model":"gpt-4.1","input":"Hello"}"#;
+/// let result = classify_request_body(body);
+/// assert_eq!(result.format.as_str(), "responses");
+/// assert_eq!(result.model.as_deref(), Some("gpt-4.1"));
+/// ```
+pub(crate) fn classify_request_body(body: &[u8]) -> ClassifiedRequest {
+    if body.is_empty() {
+        return empty_result(AiRequestFormat::NonJson);
+    }
+
+    let Ok(value) = serde_json::from_slice::<serde_json::Value>(body) else {
+        return empty_result(AiRequestFormat::InvalidJson);
+    };
+
+    let Some(obj) = value.as_object() else {
+        return empty_result(AiRequestFormat::InvalidJson);
+    };
+
+    let format = classify_format(obj);
+
+    ClassifiedRequest {
+        format,
+        model: extract_string(obj, "model"),
+        stream: obj.get("stream").and_then(serde_json::Value::as_bool),
+        store: obj.get("store").and_then(serde_json::Value::as_bool),
+        background: obj.get("background").and_then(serde_json::Value::as_bool),
+        has_previous_response_id: obj.get("previous_response_id").is_some_and(|v| !v.is_null()),
+        has_conversation: obj.get("conversation").is_some_and(|v| !v.is_null()),
+    }
+}
+
+/// Determine format from top-level keys. When both `input` and
+/// `messages` are present, `input` takes precedence (Responses API).
+fn classify_format(obj: &serde_json::Map<String, serde_json::Value>) -> AiRequestFormat {
+    if obj.contains_key("input") {
+        AiRequestFormat::Responses
+    } else if obj.contains_key("messages") {
+        AiRequestFormat::ChatCompletions
+    } else {
+        AiRequestFormat::UnknownJson
+    }
+}
+
+// -----------------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------------
+
+/// Build a result with no extracted facts.
+fn empty_result(format: AiRequestFormat) -> ClassifiedRequest {
+    ClassifiedRequest {
+        format,
+        model: None,
+        stream: None,
+        store: None,
+        background: None,
+        has_previous_response_id: false,
+        has_conversation: false,
+    }
+}
+
+/// Extract a string field from a JSON object, converting numbers/booleans
+/// to their string representation.
+fn extract_string(obj: &serde_json::Map<String, serde_json::Value>, key: &str) -> Option<String> {
+    obj.get(key).and_then(|v| match v {
+        serde_json::Value::String(s) => Some(s.clone()),
+        serde_json::Value::Number(n) => Some(n.to_string()),
+        serde_json::Value::Bool(b) => Some(b.to_string()),
+        _ => None,
+    })
+}
+
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+#[cfg(test)]
+#[allow(
+    clippy::unwrap_used,
+    clippy::expect_used,
+    clippy::indexing_slicing,
+    clippy::panic,
+    clippy::needless_raw_strings,
+    clippy::needless_raw_string_hashes,
+    reason = "tests"
+)]
+mod tests {
+    use super::*;
+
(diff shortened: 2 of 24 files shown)


filter/src/builtins/http/ai/openai/responses/classify.rs
```diff
@@ -0,0 +1,352 @@
...
+        model: extract_string(obj, "model"),
+        stream: obj.get("stream").and_then(serde_json::Value::as_bool),
+        store: obj.get("store").and_then(serde_json::Value::as_bool),
+        has_previous_response_id: obj.get("previous_response_id").is_some_and(|v| !v.is_null()),
+        has_conversation: obj.get("conversation").is_some_and(|v| !v.is_null()),
+    }
+}
+
+/// Determine format from top-level keys.
+fn classify_format(obj: &serde_json::Map<String, serde_json::Value>) -> AiRequestFormat {
```
shaneutt, line 104: Technically we didn't end up with a test (or rustdocs) for bodies containing both `input` and `messages`

filter/src/builtins/http/ai/responses_format/config.rs
```diff
@@ -0,0 +1,152 @@
...
+/// Behavior when the request body is not a recognized AI API format.
+///
+/// When set to `reject`, bodies that are not valid Responses or Chat
+/// Completions requests are rejected with HTTP 400. This includes
+/// invalid JSON, non-JSON bodies, and valid JSON that does not
+/// contain `input` or `messages`.
+#[derive(Debug, Clone, Copy, Default, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum OnInvalidBehavior {
+    /// Continue processing without classification metadata.
```
shaneutt: I'm not sure this comment is capturing what's happening on `Continue` becuse we always write metadata? 🤔

filter/src/builtins/http/ai/openai/responses/config.rs
```diff
@@ -0,0 +1,152 @@
...
+/// Completions requests are rejected with HTTP 400. This includes
+/// invalid JSON, non-JSON bodies, and valid JSON that does not
+/// contain `input` or `messages`.
+#[derive(Debug, Clone, Copy, Default, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum OnInvalidBehavior {
+    /// Continue processing without classification metadata.
+    #[default]
+    Continue,
+    /// Reject the request with HTTP 400.
```
shaneutt, line 38:
```suggestion

    /// Reject the request with HTTP 400.
```

filter/src/builtins/http/ai/openai/responses/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+/// Maximum length of a body-derived value promoted to a header.
+const MAX_PROMOTED_VALUE_LEN: usize = 256;
+use crate::{
+    FilterAction, FilterError, Rejection,
+    body::{BodyAccess, BodyMode},
+    builtins::http::ai::agentic::json_rpc::contains_control_chars,
+    factory::parse_filter_config,
+    filter::{HttpFilter, HttpFilterContext},
+};
```
shaneutt, line 50: </img>

filter/src/builtins/http/ai/responses_format/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+    ctx: &mut HttpFilterContext<'_>,
+    classified: &classify::ClassifiedRequest,
+) -> Result<(), FilterError> {
+    let results = ctx.filter_results.entry("responses_format").or_default();
+
+    results.set("format", classified.format.as_str())?;
+
+    if let Some(model) = &classified.model
+        && !contains_control_chars(model)
+        && model.len() <= 256
```
shaneutt:
```suggestion
        && model.len() <= MAX_PROMOTED_VALUE_LEN
```

filter/src/builtins/http/ai/openai/responses/tests.rs
```diff
@@ -0,0 +1,393 @@
...
+        ctx.filter_metadata
+            .get("responses_format.has_conversation")
+            .map(String::as_str),
+        Some("true")
+    );
+}
+
+#[tokio::test]
+async fn promotes_filter_results_for_full_responses_request() {
+    let ctx = run_filter("{}", FULL_RESPONSES_BODY).await;
```
shaneutt, line 293: Technically no unit test exercises the full on_request_body flow through the reject path. Only handle_invalid_format is tested in isolation.

shaneutt (changes requested):

filter/src/builtins/http/ai/openai/responses/mod.rs
nerdalert: tommy-boy-chris-farley-sd


😆  Was trying to dedup with responses_format but that wasn't great. Moved the helper to a neutral shared HTTP utility.

filter/src/builtins/http/ai/responses_format/config.rs
nerdalert: Yeah, that was a less than useful comment 😬 Updated it to reflect the actual behavior: `continue` does not reject the request, and classification facts are still  recorded when the body can be classified.

filter/src/builtins/http/ai/openai/responses/classify.rs
nerdalert: Ty! Added explicit coverage for this case and documented the precedence near the classifier logic. If both fields are present, `input` wins and the request is classified as Responses API, matching the current classifier behavior intentionally.

filter/src/builtins/http/ai/responses_format/mod.rs
nerdalert: Replaced the hard-coded `256` with `MAX_PROMOTED_VALUE_LEN` so header and filter-result promotion use the same bound.

filter/src/builtins/http/ai/openai/responses/tests.rs
nerdalert: Added full `on_request_body` reject-path coverage. Now tests the actual filter with `on_invalid: reject` and asserts that it returns`FilterAction::Reject`, instead of only testing `handle_invalid_format()`.

filter/src/builtins/http/ai/openai/responses/classify.rs
```diff
@@ -0,0 +1,365 @@
+// SPDX-License-Identifier: MIT
```
leseb, line 1: should `filter/src/builtins/http/ai/responses_format/` be `filter/src/builtins/http/ai/openai/responses/ instead? i think it'd be cleaner, and gives a natural home for the follow-on prs in the epic (#354) - also this gives rooms if we implement anthropic messages and google interactions support at one point, so we could have:

```
filter/src/builtins/http/ai/openai/responses/
filter/src/builtins/http/ai/anthropic/messages/
filter/src/builtins/http/ai/google/interactions/
```

filter/src/builtins/http/ai/openai/responses/classify.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Extracted `model` field value, if present.
+    pub model: Option<String>,
+    /// Extracted `stream` field value, if present.
+    pub stream: Option<bool>,
+    /// Extracted `store` field value, if present.
+    pub store: Option<bool>,
+    /// Whether `previous_response_id` is present and non-null.
+    pub has_previous_response_id: bool,
+    /// Whether `conversation` is present and non-null.
+    pub has_conversation: bool,
```
leseb, line 58: we can add `background` too?

nerdalert: Done!

 - Added top-level boolean background classification; absent, null, non-boolean, and nested fields are ignored.
  - Promoted background to metadata and filter results as responses_format.background, with no default header.
  - Kept this as classification only: no validation of background:true with store:false, and no rejection of background:true with stream:true.
  - Added tests for extraction, promotion, no default header, and branch routing on responses_format.background == true.

filter/src/builtins/http/ai/openai/responses/classify.rs
nerdalert: Great suggestion, done!

filter/src/builtins/http/ai/openai/responses/classify.rs
leseb: Perfect, the validation will be done by another filter.

filter/src/builtins/http/ai/openai/responses/classify.rs
```diff
@@ -0,0 +1,416 @@
...
+#[allow(
+    clippy::unwrap_used,
+    clippy::expect_used,
+    clippy::indexing_slicing,
+    clippy::panic,
+    clippy::needless_raw_strings,
+    clippy::needless_raw_string_hashes,
+    reason = "tests"
+)]
+mod tests {
```
leseb, line 156: do we need a test for responses.input being null? this should classify has responses since the key is present.

filter/src/builtins/http/ai/openai/responses/config.rs
```diff
@@ -0,0 +1,154 @@
...
+use serde::Deserialize;
+
+use crate::FilterError;
+
+// -----------------------------------------------------------------------------
+// Constants
+// -----------------------------------------------------------------------------
+
+/// Default maximum request body size for `StreamBuffer` mode (10 MiB).
+const DEFAULT_MAX_BODY_BYTES: usize = 10_485_760; // 10 MiB
```
leseb: this won't be enough if someones sends a responses request with an inline file_data (base64-encoded file) or a large input_image with a base64 data URL, the body can easily exceed 10 MiB. see the openai spec defaults:

* InputFileContentParam.file_data: maxLength 32 MiB
* InputImageContentParamAutoParam.image_url: maxLength 20 MiB
* InputTextContentParam.text: maxLength 10 MiB

a reasonable default would be 64 MiB so it covers a request with one 32 MiB file_data + one 20 MiB image_url + the JSON envelope

filter/src/builtins/http/value_safety.rs
```diff
@@ -0,0 +1,17 @@
...
+// -----------------------------------------------------------------------------
+// Control Character Detection
+// -----------------------------------------------------------------------------
+
+/// Returns `true` if `s` contains ASCII control characters (0x00-0x1F
+/// except 0x09 horizontal tab) or DEL (0x7F).
+///
+/// Used to reject header injection attempts when promoting body-derived
+/// values to request headers, metadata, or filter results.
+pub(crate) fn contains_control_chars(s: &str) -> bool {
```
leseb, line 19: can't we use http::HeaderValue::from_str(s).is_err() instead? unless we want to support utf-8 for metadata / filter results, but that seems unlikely?

filter/src/builtins/http/ai/openai/responses/config.rs
leseb: or perhaps just let it resolve to usize::MAX when max_bytes is None?

examples/configs/ai/openai/responses/format-routing.yaml
```diff
@@ -0,0 +1,56 @@
+# Responses Format Routing
```
leseb, line 1: put the config into `examples/configs/ai/openai/responses` just like the rust files?

leseb: and just call the file `format-routing.yaml`?

filter/src/builtins/http/ai/openai/responses/classify.rs
nerdalert: Done: Added explicit coverage for `{"input": null}`. Classification is key-presence based here, so this still classifies as Responses; semantic validation of whether null input is accepted belongs to a later validation/orchestration filter.

filter/src/builtins/http/ai/openai/responses/config.rs
nerdalert: Done: Raised the default `max_body_bytes` to 64 MiB to cover common inline Responses payloads while keeping request buffering bounded. Did not default to `usize::MAX` because this filter uses `StreamBuffer`, and an unbounded default would make memory behavior less predictable. The value remains configurable.

filter/src/builtins/http/value_safety.rs
nerdalert: Done: Switched promoted-value safety to `http::HeaderValue::from_str` and added a positive helper, `is_safe_promoted_value`, for new code. Left the existing  inverse helper as a compatibility wrapper for current call sites. Tests cover visible ASCII, newline rejection, and the actual HTTP parser behavior for tab.

examples/configs/ai/openai/responses/format-routing.yaml
nerdalert: Done!

shaneutt (approved):

filter/src/builtins/http/ai/openai/responses/classify.rs
leseb: Thanks!

filter/src/builtins/http/ai/openai/responses/config.rs
leseb: yup, thanks

PR #375, opened 2026-05-21
fix: fail early when build prerequisites are missing or broken

leseb:
## Summary

Adds prerequisite checks to the Makefile that run before any build/test/lint target. Catches missing or broken tools (cargo, cmake, nightly toolchain) at `make` time instead of letting them fail deep inside a Cargo build script with a cryptic error.

**Why:** cmake on macOS can appear installed (`command -v cmake` succeeds) but be broken — e.g. a Homebrew cmake whose Python shebang points to a removed interpreter. The build script for `libz-ng-sys` then panics with "is cmake not installed?" after minutes of compilation. Same for a missing Rust nightly toolchain failing only at the `cargo +nightly fmt` step at the end of `make lint`.

**What:**
- `check-prereqs` validates `cargo` using `--version` (catches broken installs, not just missing binaries)
- `check-prereqs-cmake` validates `cmake --version` — only wired as a dependency for targets that compile Rust (build, test, bench, lint, etc.)
- `check-prereqs-nightly` validates `cargo +nightly --version` — only wired for targets that need nightly (lint, fmt, fuzz)
- Targets that don't need these tools (container, help, tools) are unaffected

## Test plan
- [x] `make check-prereqs` passes when cargo is available
- [x] `make check-prereqs` fails with clear message when cargo is missing from PATH
- [x] `make check-prereqs-cmake` fails with clear message when cmake is broken (dead shebang)
- [x] `make check-prereqs-nightly` fails with clear message when nightly toolchain is not installed
- [x] Container and help targets are not gated by prerequisite checks


Makefile
```diff
@@ -9,6 +9,30 @@ V       ?=
 UNAME_S := $(shell uname -s | tr A-Z a-z)
 UNAME_M := $(shell uname -m)
 
+# -------------------------------------------------------------------
+# Prerequisite checks
+# -------------------------------------------------------------------
+
+REQUIRED_CMDS := cargo
+RUST_TARGETS := all build release check clean \
```
shaneutt:
```suggestion
RUST_TARGETS := all build release check \
```

shaneutt (dismissed):
One comment, when that's resolved 👍

Makefile
leseb: Done in https://github.com/praxis-proxy/praxis/pull/375/changes/8d13e8e58ad7798417c08f2a5d72390149244a24

shaneutt (approved):
:+1:

PR #378, opened 2026-05-21
fix(makefile): add missing xtask lint-deps to lint target

leseb:
## Summary

The `make lint` target was documented in CLAUDE.md as running `clippy + nightly fmt check + xtask lint-deps`, but the actual Makefile only ran clippy and fmt. This adds the missing `cargo xtask lint-deps` step so the target matches its documentation.


leseb:
I guess, we either remove it from CLAUDE.md if we feel we don't need it or we add it to the Make target, up to the maintainers.

shaneutt (approved):
</img>

PR #379, opened 2026-05-21
fix(tls): zeroize private key PEM bytes after parsing

leseb:
## Summary

- Wrap intermediate `Vec<u8>` holding raw PEM key material in `Zeroizing::new()` in both `loader.rs` and `cached.rs`
- Ensures private key bytes are overwritten with zeros on drop, preventing them from lingering in freed heap memory
- The crate already used `Zeroizing` for the final DER-encoded key; this closes the gap on the intermediate PEM stage

## Test plan

- [x] `cargo check -p praxis-proxy-tls` passes
- [x] `cargo test -p praxis-proxy-tls` — all 124 unit tests and 22 doc-tests pass
- [x] `make fmt` clean


tls/src/cached.rs
```diff
@@ -245,7 +245,7 @@ fn parse_cert_pem(cert_path: &str) -> Result<Vec<Vec<u8>>, TlsError> {
 
 /// Read a PEM private key file and return the DER-encoded key bytes.
 fn parse_key_pem(key_path: &str) -> Result<Vec<u8>, TlsError> {
```
shaneutt: While we're in here, we can make the type system guarantee the zeroize on return of this function:

```suggestion
fn parse_key_pem(key_path: &str) -> Result<Zeroizing<Vec<u8>>, TlsError> {
```

We'll need to wrap the map below.

shaneutt (dismissed):
Good change, only one comment. Once it's resolved LGTM.

tls/src/cached.rs
leseb: Good call, addressed in https://github.com/praxis-proxy/praxis/pull/379/changes/467a39ca3d4ee666fd20fd2d23d5cd1510496f35.

shaneutt (approved):

PR #380, opened 2026-05-21
fix: make PRAXIS_LOG_FORMAT comparison case-insensitive

leseb:
## Summary

- Makes the `PRAXIS_LOG_FORMAT` environment variable comparison case-insensitive by using `eq_ignore_ascii_case("json")` instead of `== "json"` in `xtask/src/main.rs`.
- Previously, only the exact lowercase `json` value was recognized; now `JSON`, `Json`, etc. all work as expected.

## Test plan

- [x] `make fmt` passes
- [x] `make lint` passes
- [x] `cargo check -p xtask` passes


shaneutt (approved):
Good catch, there's one other place we'll wanna do this too I'll grab that.

PR #382, opened 2026-05-21
fix(router): enforce segment-boundary path prefix matching

aslakknutsen:
Path prefix /api now only matches /api, /api/, and /api/... — not /apikeys. Aligns with Gateway API HTTPPathMatch semantics (element-wise prefix matching). Also applies to the condition evaluator's path_prefix.

### DIFF
--- a/filter/src/builtins/http/traffic_management/router/matching.rs
+++ b/filter/src/builtins/http/traffic_management/router/matching.rs
@@ -75,7 +75,10 @@ pub(super) fn update_best_match<'a>(
 
 /// Return `true` if shorter prefixes cannot improve on the current best.
 pub(super) fn should_stop_early(best: Option<(Specificity, &Route)>, route: &Route) -> bool {
-    let route_len = route.path_match.len();
+    let route_len = match &route.path_match {
+        PathMatch::Exact { path } => path.len(),
+        PathMatch::Prefix { path_prefix } => crate::path_match::path_prefix_specificity(path_prefix),
+    };
     best.is_some_and(|((_, bp, _), _)| route_len < bp)
 }
 
--- a/filter/src/builtins/http/traffic_management/router/matching.rs
+++ b/filter/src/builtins/http/traffic_management/router/matching.rs
@@ -30,7 +30,7 @@ pub(super) fn route_matches_request(
             }
         },
         PathMatch::Prefix { path_prefix } => {
-            if !path.starts_with(path_prefix.as_str()) {
+            if !crate::path_match::path_prefix_matches(path, path_prefix) {
                 return false;
             }
         },
@@ -52,7 +52,10 @@ pub(super) type Specificity = (bool, usize, usize);
 /// Computes the specificity of a route for comparison.
 fn route_specificity(route: &Route) -> Specificity {
     let is_exact = route.path_match.is_exact();
-    let path_len = route.path_match.len();
+    let path_len = match &route.path_match {
+        PathMatch::Exact { path } => path.len(),
+        PathMatch::Prefix { path_prefix } => crate::path_match::path_prefix_specificity(path_prefix),
+    };
     let constraints = usize::from(route.host.is_some()) + route.headers.as_ref().map_or(0, HashMap::len);
     (is_exact, path_len, constraints)
 }
--- a/filter/src/builtins/http/traffic_management/router/mod.rs
+++ b/filter/src/builtins/http/traffic_management/router/mod.rs
@@ -23,7 +23,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use http::{HeaderMap, header::HeaderName};
-use praxis_core::config::{PathMatch, Route};
+use praxis_core::config::Route;
 use tracing::{debug, trace};
 
 use self::{
@@ -116,8 +116,9 @@ struct ResolvedRoute {
 impl RouterFilter {
     /// Create a router from a list of routes with default alias options.
     ///
-    /// Returns an error if any `path_prefix` (other than `"/"`)
-    /// does not end with `'/'`.
+    /// Path prefix matching uses segment-boundary semantics per Gateway API:
+    /// `/api` matches `/api`, `/api/`, `/api/v1` but NOT `/apikeys`.
+    /// Trailing slashes on prefixes are ignored (`/api` ≡ `/api/`).
     ///
     /// ```
     /// use praxis_core::config::{PathMatch, Route};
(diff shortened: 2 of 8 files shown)


filter/src/builtins/http/traffic_management/router/matching.rs
```diff
@@ -52,7 +52,10 @@ pub(super) type Specificity = (bool, usize, usize);
 /// Computes the specificity of a route for comparison.
 fn route_specificity(route: &Route) -> Specificity {
     let is_exact = route.path_match.is_exact();
-    let path_len = route.path_match.len();
+    let path_len = match &route.path_match {
+        PathMatch::Exact { path } => path.len(),
+        PathMatch::Prefix { path_prefix } => crate::path_match::path_prefix_specificity(path_prefix),
+    };
```
usize, line 58: `route_specificity` now uses the trimmed length (via `path_prefix_specificity`), which is correct. But two callers in this file still use raw `path_match.len()`:

1. **`sort_routes`** (below, line ~259) sorts by `route.path_match.len()`, so `/api/` (5) sorts ahead of `/api` (4) even though they're semantically equivalent under the new matching.
2. **`should_stop_early`** (line 75) compares the candidate's raw `path_match.len()` against the best's specificity (which is now trimmed). This is conservative (never incorrect) but inconsistent.

Neither is a correctness bug today — `update_best_match` uses the correct trimmed specificity regardless of iteration order — but the inconsistency is surprising and could bite future contributors. Consider updating `sort_routes` and `should_stop_early` to use `path_prefix_specificity` as well.

filter/src/condition/request.rs
```diff
@@ -76,7 +76,7 @@ fn matches_request(m: &ConditionMatch, req: &Request) -> bool {
     }
 
     if let Some(ref prefix) = m.path_prefix
-        && !req.uri.path().starts_with(prefix)
+        && !crate::path_match::path_prefix_matches(req.uri.path(), prefix)
```
usize, line 79: We have unit tests, but could we also add an integration test like:

```rust
#[test]
fn when_path_prefix_rejects_non_segment_boundary() {
    let req = make_request(Method::GET, "/apikeys", HeaderMap::new());
    assert!(!should_execute(&[when(path_match("/api"))], &req));
}
```

filter/src/path_match.rs
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Shane Utt
```
usize: nit: Copyright should be new style e.g., praxis contributors

shaneutt (commented):
It is my understanding that Gateway API is the only standard that forces this. I'm not opposed to making it the default here, as it is generally safer. However: all other proxies do not do this, and only Envoy allows an opt-in, so I suppose the question is: should we provide an opt-out?

Not a question that needs to block this PR, just food for thought.

filter/src/path_match.rs
shaneutt:
```suggestion
// Copyright (c) 2026 Praxis Contributors
```

shaneutt (approved):
Thanks @aslakknutsen 👍

We may revisit having an opt-out later, but we can allow necessity to be the mother of invention. Once all comments are fully resolved +1

aslakknutsen:
Sorry, missed the feedback here. I'll have a look

shaneutt (dismissed):
Once the review items from @praxis-bot are addressed, LGTM lets merge it and anything else we want to do we can handle as a follow up.

filter/src/builtins/http/traffic_management/router/mod.rs
```diff
@@ -116,8 +116,9 @@ struct ResolvedRoute {
 impl RouterFilter {
     /// Create a router from a list of routes with default alias options.
     ///
```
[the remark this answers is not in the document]
aslakknutsen: fixed in https://github.com/praxis-proxy/praxis/pull/382/commits/e715ca82110b5a1211627fc8265b8d8287352bc4

filter/src/condition/request.rs
aslakknutsen: fixed in https://github.com/praxis-proxy/praxis/pull/382/commits/e715ca82110b5a1211627fc8265b8d8287352bc4

filter/src/builtins/http/traffic_management/router/matching.rs
aslakknutsen: fixed in https://github.com/praxis-proxy/praxis/pull/382/commits/e715ca82110b5a1211627fc8265b8d8287352bc4

filter/src/builtins/http/traffic_management/router/tests.rs
```diff
@@ -721,20 +721,46 @@ fn should_stop_early_false_when_no_best() {
...
-        host: None,
-        headers: None,
-        cluster: "api".into(),
-    }])
-    .unwrap_err();
-    assert!(
-        err.to_string().contains("must end with '/'"),
-        "path_prefix without trailing slash should be rejected: {err}"
-    );
+fn prefix_without_trailing_slash_accepted() {
```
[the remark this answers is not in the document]
aslakknutsen: fixed in https://github.com/praxis-proxy/praxis/pull/382/commits/ee5125ee8df1e29c165596c95450707766348be4

filter/src/path_match.rs
```diff
@@ -0,0 +1,93 @@
...
+/// or continue with a `/` separator.
+pub(crate) fn path_prefix_matches(path: &str, prefix: &str) -> bool {
+    let trimmed = prefix.strip_suffix('/').unwrap_or(prefix);
+    if trimmed.is_empty() {
+        return true;
+    }
+    if path == trimmed {
+        return true;
+    }
+    path.starts_with(trimmed) && path.as_bytes().get(trimmed.len()) == Some(&b'/')
```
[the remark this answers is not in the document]
aslakknutsen: fixed in https://github.com/praxis-proxy/praxis/pull/382/commits/ee5125ee8df1e29c165596c95450707766348be4

filter/src/builtins/http/traffic_management/router/matching.rs
```diff
@@ -52,7 +52,10 @@ pub(super) type Specificity = (bool, usize, usize);
 /// Computes the specificity of a route for comparison.
 fn route_specificity(route: &Route) -> Specificity {
     let is_exact = route.path_match.is_exact();
-    let path_len = route.path_match.len();
+    let path_len = match &route.path_match {
```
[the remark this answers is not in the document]
aslakknutsen: fixed in https://github.com/praxis-proxy/praxis/pull/382/commits/e715ca82110b5a1211627fc8265b8d8287352bc4

PR #388, opened 2026-05-26
fix(test-utils): use unique filenames for generated client certificates

leseb:
## Summary

- Add an `AtomicUsize` counter to `TestCertificates` so each `generate_client_cert()` call produces uniquely named files (`client-{n}.pem` / `client-{n}-key.pem`)
- Prevents silent file overwrite if the method is called multiple times on the same instance (e.g. testing two distinct client identities in an mTLS scenario)
- Counter is initialized in both `generate()` and `generate_for_san()` constructors

## Test plan

- [ ] Verify `cargo check -p praxis-test-utils` passes
- [ ] Verify existing mTLS integration tests still pass
- [ ] Confirm calling `generate_client_cert()` twice on the same instance produces two distinct file pairs

### DIFF
--- a/tests/utils/src/net/tls.rs
+++ b/tests/utils/src/net/tls.rs
@@ -7,7 +7,10 @@ use std::{
     io::{Read, Write},
     net::TcpStream,
     path::PathBuf,
-    sync::Arc,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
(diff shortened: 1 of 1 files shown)


leseb:
gh hick up

shaneutt (approved):

PR #390, opened 2026-05-26
proposal: HTTP callout filter for inline sub-requests

usize:
## Summary

- Adds proposal `docs/proposals/00358_http-callout-filter.md` for an `http_callout` filter that makes outbound HTTP requests to external services during request processing
- Covers inline guardrail callouts (e.g. Lakera Guard), content classification, authorization checks, and dynamic enrichment — without requiring ext-proc
- Proposes solutions for HTTP client lifecycle, timeout budget, cancellation, recursion guard, circuit breaker state, inter-filter state, and memory bounding

Refs: praxis-proxy/ai#97
Discussion: https://github.com/praxis-proxy/praxis/discussions/87

Assisted by opus 4.6


mariusdanciu:
Just out of curiosity. Since praxis is essentially a lib crate, why would this filter be preferred as opposed of registering a custom filter that does the same with more flexibility ? Since the filter calls out to other services, a custom filter would manage specific protocol details perhaps in a more optimal manner than describing the behavior of API handling via the config props.

docs/proposals/00358_http-callout-filter.md
```diff
@@ -0,0 +1,425 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/358
+discussion: https://github.com/praxis-proxy/praxis/discussions/87
+status: proposed
+authors:
+  - usize
```
shaneutt, line 6:
```suggestion
  - usize
stakeholders:
  - shaneutt
  - twghu
  - nerdalert
```

(recently added field)

docs/proposals/00358_http-callout-filter.md
```diff
@@ -0,0 +1,425 @@
...
+external services during request processing. The filter sends
+a request to a configured target, extracts fields from the
+response, and writes them into filter results for downstream
+branch-chain evaluation.
+
+This is the first concrete deliverable from the sub-request
+orchestration primitive described in
+[discussion #87](https://github.com/praxis-proxy/praxis/discussions/87),
+scoped to inline HTTP callouts with fail-open/closed
+semantics — without requiring ext-proc.
```
shaneutt:
```suggestion
semantics without requiring ext-proc.
```

docs/proposals/00358_http-callout-filter.md
```diff
@@ -0,0 +1,425 @@
...
+        failure_mode: closed
+        circuit_breaker:
+          failure_threshold: 5
+          recovery_timeout: 30s
+
+      - filter: guardrails
+        rules:
+          - target: header
+            name: "x-praxis-callout-flagged"
+            contains: "true"
```
shaneutt: The design says we're extracting results to `FilterResultSet`, but here we're checking for a header. Is this example broken, or am I missing something? 🤔

docs/proposals/00358_http-callout-filter.md
```diff
@@ -0,0 +1,425 @@
...
+- **NGINX auth_request** — sub-request to an authorization
+  endpoint; response status controls access.
+- **Lakera Guard** — content-safety HTTP API for prompt
+  injection, PII, and harmful content detection
+  ([docs](https://docs.lakera.ai/docs/api/guard)).
+- **NVIDIA NeMo Guardrails** — self-hosted guardrails
+  server with configurable safety rails
+  ([docs](https://docs.nvidia.com/nemo/guardrails/latest/user-guides/server-guide.html)).
+
+## How?
```
shaneutt: The proposal conventions ask that the first PR for a proposal only include the "What?" and "Why?" but wait on the "How?" until those are agreed on and merged.  Can this "How?" section wait?

shaneutt (changes requested):

docs/proposals/00358_http-callout-filter.md
usize: Good catch. Earlier example was based on header-promotion and this lingered. I'll change it to use FilterResultSet.

docs/proposals/00358_http-callout-filter.md
usize: Yes. Can remove for now if we'd like.

usize:
> Just out of curiosity. Since praxis is essentially a lib crate, why would this filter be preferred as opposed of registering a custom filter that does the same with more flexibility ? Since the filter calls out to other services, a custom filter would manage specific protocol details perhaps in a more optimal manner than describing the behavior of API handling via the config props.

Such a custom filter would still be an option. The idea is to build support for inline configuration of straightforward cases, like the Lakera example given. There are a number of vendors selling guardrails services. Having a simple out-of-the-box way of configuring the proxy to use them seems valuable.

It also gives us the scaffolding for use cases that might require deeper changes, like the disaggregated inference example in #87 -- where we might need to be able to overwrite the response with a callout.

shaneutt (approved):
I don't see any blockers, we can merge and iterate forward from here.

If the custom filters question is something we need to revisit, make a separate PR that adds resolving that question as a graduation criteria (e.g. we can't move out of proposed until stakeholders consider it resolved)

PR #391, opened 2026-05-27
Draft: feat(filter): add ConversationBackend trait and types (#359)

jlojosnegros:
Following up on the phased approach proposed in praxis-proxy/ai#98. This is step 1a:
the `ConversationBackend` trait and its supporting types.

Opening this as a draft to collect early feedback on the API shape
before the in-memory implementation goes in.

## What's in this PR

A new module at `filter/src/builtins/http/ai/state/` containing:

- **`ConversationBackend`**: pluggable trait with four operations:
  `create`, `get`, `append`, `delete`
- **`ConversationScope`**: four-level hierarchy (`Global`, `Tenant`,
  `User`, `Session`) encoded as an enum so invalid combinations are
  impossible at compile time
- **`ConversationMessage`**: immutable, constructed via `new()`;
  content is `serde_json::Value` to stay agnostic to the wire format
  (Chat Completions vs Responses API)
- **`ConversationEntry`**: returned by the backend; private fields
  with read-only getters so only backend implementations can construct
  entries
- **`ConversationError`**: `ScopeMismatch` is a distinct variant from
  `NotFound` to avoid masking access control violations (403 vs 404)

## Design decisions worth calling out

**TTL is required in `create`.** I made `ttl: Duration` non-optional
following the direction in praxis-proxy/ai#146, which requires every hot-path state
operation to define TTL and cardinality bounds. If a very long lifetime
is needed it can be expressed as `Duration::MAX`.

**`get` returns `Option`, `append` returns `Err(NotFound)`.** Not
finding a conversation in `get` is routine: the filter creates a new
one. Not finding it in `append` means the caller sent a bad
`previous_response_id`, which is an error worth surfacing explicitly.

**`ConversationScope` is for access control, not key uniqueness.**
Conversation IDs are UUID v4 and globally unique. The scope is stored
in the entry and backends are expected to check it on retrieval,
returning `ScopeMismatch` on a mismatch. Whether and how strictly
backends enforce this is their responsibility.

**The backend generates IDs as UUID v4 without hyphens (`simple()`),
producing 32 lowercase hex characters.** This matches the suffix format
of a real OpenAI `resp_` ID, so the filter layer can translate between
the backend key and the wire format with a single prefix add/strip and
no re-encoding:

```text
backend stores : 550e8400e29b41d4a716446655440000
sent to client : resp_550e8400e29b41d4a716446655440000
received back  : strip "resp_" -> 550e8400e29b41d4a716446655440000
```

## Open questions

1. Is `serde_json::Value` for message content acceptable, or would you
   prefer a typed `ContentPart` enum? The tradeoff is flexibility vs
   compile-time validation.

2. Module location: `filter/src/builtins/http/ai/state/` works for now,
   but if the plan is to share state backends across crates a separate
   workspace crate would make more sense. Happy to move it if you have
   a preference.

Relates to praxis-proxy/ai#98.


filter/src/builtins/http/ai/state/memory.rs
```diff
@@ -0,0 +1,324 @@
...
+
+use std::{sync::Arc, time::Duration};
+
+use async_trait::async_trait;
+use dashmap::DashMap;
+use uuid::Uuid;
+
+use super::{ConversationBackend, ConversationEntry, ConversationError, ConversationMessage, ConversationScope};
+
+/// In-memory [`ConversationBackend`] backed by a [`DashMap`]
```
leseb, line 18: why not use sqlite::memory to validate both trait contracts and sql semantic? also this will be the default backend for developers. using this feels like a waste of time/effort since we know (i think) a sql backend is required anyway.

filter/src/builtins/http/ai/state/memory.rs
```diff
@@ -0,0 +1,324 @@
...
+        };
+
+        self.store.insert(Arc::clone(&id), entry);
+
+        Ok(id)
+    }
+
+    async fn get(&self, scope: &ConversationScope, id: &str) -> Result<Option<ConversationEntry>, ConversationError> {
+        let now = super::now_ms();
+        let Some(entry_ref) = self.store.get(id) else {
```
leseb, line 83: get is an impossible api to serve; we need cursor-based pagination (cursor, offset, limit, or streaming iterator)

leseb:
The trait should be renamed from `ConversationBackend` to `ResponseStore` and rebuilt around Responses API semantics.

**Missing methods:**
- `upsert_response` / `get_response` / `delete_response` / `list_responses` ... response-level CRUD
- `list_input_items(response_id, cursor, limit)` ... cursor-based pagination
- `get_by_previous_response_id` ... chain lookup for rehydrate (Filter 2)
- `get_by_conversation_id` ... context lookup by API conversation ID
- Streaming lifecycle: `begin` / `update` / `finalize` / `fail` / `cancel`

**Externally-supplied response IDs:** The store must accept IDs from `request_validate` (Filter 0), not generate them internally.

**New types needed:**
- `ResponseEntry` ... id, status, model, usage, metadata, response_object JSON, previous_response_id, conversation_id, timestamps
- `ResponseItem` ... stable IDs, sequence numbers, type, hidden/internal flag
- `PagedResult<T>` ... items + next_cursor

**Additional error variants:** `AlreadyExists`, `Conflict`, `InvalidState`

**SQLite in-memory backend instead of DashMap.** Production is SQLite/Postgres, so dev/test should use `sqlite::memory:` to validate trait contract and SQL semantics together.

**What carries over:** `ConversationScope` (maybe renamed to `Scope`), error type structure, module layout.

jlojosnegros:
Hi @leseb, thanks for the detailed review — this is exactly the kind of feedback I needed.

I want to be transparent about the context in which this PR was written. When I picked up issue praxis-proxy/ai#98, the GitHub issue was the only design information I had access to. I wasn't aware of the decomposition document or the proxy conformance guidelines that informed your design. Working from the issue description alone, I modeled a generic conversation CRUD abstraction, which made sense given what was visible — but clearly misses what the project actually needs.

Having now read the decomposition doc, I understand the gap. `response_store` is not a generic message store: it's a SQL-backed component (SQLite/Postgres, two tables) that persists full `ResponseResource` objects plus hidden messages, receives IDs from `request_validate` rather than generating them, handles incremental upserts driven by SSE lifecycle events, exposes its own HTTP CRUD routes, and is shared across multiple filters via `Arc`. None of that is derivable from the issue description.

Given that, this PR doesn't make sense to keep open. I'll close it now. If there's a good entry point for a new contributor once the design is more settled — or if there's a smaller piece of `response_store` where a PR would actually be useful — I'm happy to pick it up.

Thanks again for taking the time to review it.

leseb:
> Hi @leseb, thanks for the detailed review — this is exactly the kind of feedback I needed.
> 
> I want to be transparent about the context in which this PR was written. When I picked up issue praxis-proxy/ai#98, the GitHub issue was the only design information I had access to. I wasn't aware of the decomposition document or the proxy conformance guidelines that informed your design. Working from the issue description alone, I modeled a generic conversation CRUD abstraction, which made sense given what was visible — but clearly misses what the project actually needs.
> 
> Having now read the decomposition doc, I understand the gap. `response_store` is not a generic message store: it's a SQL-backed component (SQLite/Postgres, two tables) that persists full `ResponseResource` objects plus hidden messages, receives IDs from `request_validate` rather than generating them, handles incremental upserts driven by SSE lifecycle events, exposes its own HTTP CRUD routes, and is shared across multiple filters via `Arc`. None of that is derivable from the issue description.
> 
> Given that, this PR doesn't make sense to keep open. I'll close it now. If there's a good entry point for a new contributor once the design is more settled — or if there's a smaller piece of `response_store` where a PR would actually be useful — I'm happy to pick it up.
> 
> Thanks again for taking the time to review it.

Sure, looking fw to the followup! @jlojosnegros do you want to picj filter responses store from https://github.com/praxis-proxy/praxis/pull/418?

Issue #210, opened 2026-05-15
Response-based token counting from provider JSON


shaneutt:
Extract token usage fields from provider response bodies. Parse `usage.prompt_tokens` and `usage.completion_tokens` from OpenAI-format JSON responses. Map equivalent fields from Anthropic, Google, Bedrock, and Azure response schemas.

PR #407, opened 2026-05-28
proposal: add proposal for provider token usage mapping

yehuditkerido:
Propose a library to extract token usage from AI provider responses
(OpenAI, Anthropic, Google, Bedrock, Azure) into a unified TokenUsage
struct. This enables downstream features like token-based rate limiting.

Refs: praxis-proxy/ai#87

### DIFF
--- a/docs/proposals/00216_provider-token-mapping.md
+++ b/docs/proposals/00216_provider-token-mapping.md
@@ -1,0 +1,118 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/216
+discussion: https://github.com/praxis-proxy/praxis/issues/216#issuecomment-4563373566
+status: proposed
+authors:
+  - yehuditkerido
+stakeholders:
+  - shaneutt
+  - twhgu
+---
+
+# Provider-Specific Token Usage Mapping
+
+## What?
+
+An internal module to extract token usage from AI provider responses
+and convert them to a unified format.
+
+Each provider returns token counts in different JSON structures:
+
+| Provider  | Field Path |
+|-----------|------------|
+| OpenAI    | `usage.prompt_tokens`, `usage.completion_tokens` |
+| Anthropic | `usage.input_tokens`, `usage.output_tokens` |
+| Google    | `usageMetadata.promptTokenCount`, `usageMetadata.candidatesTokenCount` |
+| Bedrock (InvokeModel) | `inputTokenCount`, `outputTokenCount` (root level) |
+| Bedrock (Converse)    | `usage.inputTokens`, `usage.outputTokens` |
+| Azure     | Same as OpenAI |
+
+This proposal adds a mapping library that:
+
+1. Takes a provider identifier and response body
+2. Parses the provider-specific JSON structure
+3. Returns a unified `TokenUsage` struct
+
+```rust
+pub struct TokenUsage { /* fields private */ }
+
+impl TokenUsage {
+    pub fn input_tokens(&self) -> u64;
+    pub fn output_tokens(&self) -> u64;
+    pub fn total_tokens(&self) -> u64;  // from response, or computed: input + output
+}
+
+pub enum Provider {
+    OpenAi,
+    Anthropic,
+    Google,
+    Bedrock,
+    Azure,
+}
+
+/// Extracts token usage from a provider's JSON response body.
+/// Returns `None` if usage data is missing or malformed.
+pub fn extract_token_usage(provider: Provider, body: &[u8]) -> Option<TokenUsage>;
+```
+
+### Goals
+
+- Extract token usage from 5 AI provider response formats
+- Provide a unified `TokenUsage` representation
+- Enable reuse by other filters ([#210], [#212]) and systems
+
+### Non-Goals
+
+- Rate limiting logic (separate issue)
+- Streaming-specific logic: SSE parsing, chunk identification ([#211])
+- Exposing tokens in headers or metrics ([#214])
+
+**Note on streaming:** This module parses JSON containing `usage` data,
+regardless of whether it came from a non-streaming response or the
+final chunk of a streaming response. The JSON structure is the same
+in both cases. Streaming-specific handling (SSE format, identifying
+which chunk has usage) is #211's scope.
+
+## Why?
+
+### Motivation
+
+Other filters and systems need token counts in a consistent
+format. Without this library, each consumer would need to
+implement provider-specific parsing separately.
+
+This is foundational work for the Token Counting epic ([#20])
+that enables:
+- Token-based rate limiting
+- Usage logging and metrics
+- Cost tracking
+
+### User Stories
+
+- As a **rate limiting filter**, I need consistent token counts
+  regardless of which AI provider was used, so that I can enforce
+  limits without provider-specific code.
+
+- As the **token counting filter ([#210])**, I need a mapping
+  from each provider's field names to a unified format, so that I
+  can parse any provider's response.
+
+- As a **logging/metrics system**, I need token data in a
+  predictable structure, so that I can track usage across all
+  providers consistently.
+
+## Open Question
+
+This proposal focuses on extracting token usage only. Should
+we consider a broader "provider response translator" that
+normalizes the entire response to a common format (e.g., OpenAI)?
+
+If there are future requirements that would benefit from full
+response translation, it may be worth designing the library
+with that in mind.
+
+[#20]: https://github.com/praxis-proxy/praxis/issues/20
+[#210]: https://github.com/praxis-proxy/praxis/issues/210
+[#211]: https://github.com/praxis-proxy/praxis/issues/211
+[#212]: https://github.com/praxis-proxy/praxis/issues/212
+[#214]: https://github.com/praxis-proxy/praxis/issues/214


docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,65 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/216
+discussion: https://github.com/praxis-proxy/praxis/issues/216#issuecomment-4563373566
+status: proposed
+authors:
+  - yehuditkerido
```
shaneutt, line 6:
```suggestion
  - yehuditkerido
  - shaneutt 
  - twhgu
```

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,65 @@
...
+
+```rust
+pub struct TokenUsage {
+    pub input_tokens: u64,
+    pub output_tokens: u64,
+    pub total_tokens: u64,
+}
+```
+
+### Non-Goals
```
shaneutt, line 64: We've explicitly called out our non-goals, let's please capture our goals as well (at a high level, bullet pointed is fine)

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,65 @@
...
+## Why?
+
+Other filters and systems need token counts in a consistent
+format. Without this library, each consumer would need to
+implement provider-specific parsing.
+
+This is foundational work that enables:
+- Token-based rate limiting
+- Usage logging and metrics
+- Cost tracking
```
shaneutt, line 88: Could we go into a bit more detail here? The template has motivation, and user stories can we add user stories here to make the motivation extra clear?

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. Takes a provider identifier and response body
+2. Parses the provider-specific JSON structure
+3. Returns a unified `TokenUsage` struct
+
+```rust
+pub struct TokenUsage {
+    pub input_tokens: u64,
+    pub output_tokens: u64,
+    pub total_tokens: u64,
```
shaneutt: Total tokens will always be the sum of the other two, so the invariant here will need to be maintained. Perhaps we just provide a method?

shaneutt (changes requested):
Thank you @yehuditkerido!

Mostly minor comments, and an ask about expanding the "Why?" and motivation.

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,65 @@
...
+This proposal adds a mapping library that:
+
+1. Takes a provider identifier and response body
+2. Parses the provider-specific JSON structure
+3. Returns a unified `TokenUsage` struct
+
+```rust
+pub struct TokenUsage {
+    pub input_tokens: u64,
+    pub output_tokens: u64,
```
alexsnaps: This means `output_tokens` would be `0` until we got to the response? Or can I only get to a `TokenUsage` instance once _all_ token usages have been resolved? 
I'm generally not in favor of "magic values". So wondering if these should be `Option<u64>`s? Only `output_tokens`, and `TokenUsage` itself would be made available as an `Option`, so that until at least `input_tokens` has been resolved, you can't get to a `TokenUsage` at all... Do we need to concern ourselves with never resolving `input_tokens`, yet getting to a response?

docs/proposals/00216_provider-token-mapping.md
alexsnaps: I would advocate for methods for all field accesses, and keep fields private. In this particular case it probably doesn't matter too much, "only 16 (or 24) bytes" per connection, but should we want to to store these `u64` differently, we couldn't "easily" anymore. All those calls would get inline'd away anyways. wdyt?

docs/proposals/00216_provider-token-mapping.md
yehuditkerido: If I understand correctly, you're thinking about the streaming scenario where values might arrive incrementally? For #216, the library expects a JSON body that already contains the complete `usage` data - whether that's a non-streaming response or the final chunk of a streaming response (with `stream_options.include_usage: true`). The JSON structure is the same in both cases ([ref](https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156)).

So the function returns `Option<TokenUsage>`:
- `Some(TokenUsage)` if the body has valid usage data
- `None` if it doesn't (error response, missing field, etc.)

There's no partial state - we either have all values or none. The logic of *which* chunk to parse in streaming (finding the one with `usage`) would be handled by #211, which can then call this library.

alexsnaps: Right, so we either got it all, or we expose nothing (ie a `None` is returned until the data is complete). Probably is fine for most, in the case of Bedrock (going by your doc here) they could be "far apart", is it acceptable? Also, do we have the guarantee these are always both provided (I see them doc'ed mostly, but couldn't figure out if they really are always both provided)...

yehuditkerido: From what I found in the [Bedrock TokenUsage docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_TokenUsage.html), when `usage` is present, all fields are marked as required. Anthropic seems similar.

About "far apart" in Bedrock - those fields are just at a different JSON path but still in the same response body. Let me know if I'm missing something there.

My understanding on partial data: JSON parsing should be all-or-nothing, so if parsing succeeds we'd have all values, and if something breaks we'd get `None`. Same for streaming chunks. But happy to hear if there are edge cases I'm not considering.

Based on this, `Option<TokenUsage>` seemed like the right fit.

alexsnaps: Cool! If making that available atomically & knowing they must always both eventually resolve, then that API makes sense to me too. Wanted to make sure tho. Thanks 🙏

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,98 @@
...
+  - yehuditkerido
+  - shaneutt
+  - twhgu
+---
+
+# Provider-Specific Token Usage Mapping
+
+## What?
+
+A library to extract token usage from AI provider responses
```
alexsnaps: One last nitpick here, also for my understanding tho:

Do we really mean library here? i.e. a `crate`? I guess no... do we? Or is that "an internal API" to extract ... 
I think I read that we might want to be able to support provider behind a feature flag eventually, should that be part of the goals of this too? Or are we fine postponing that, if ever needed (plus I don't expect that code to bloat the binary, but...  ¯\_(ツ)_/¯ )

yehuditkerido: You're right, "library" was the wrong term - I meant an internal module within the `filter` crate, not a separate crate. Should reword that in the proposal.

On feature flags - I think we can postpone that for now. The code is just JSON parsing for each provider, nothing heavy. If binary size becomes a concern later we can revisit, but seems like premature optimization at this point.

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,98 @@
+---
```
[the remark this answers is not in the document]
yehuditkerido: Will move reviewers to stakeholders field, keeping just myself as author.

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,98 @@
...
+| Google    | `usageMetadata.promptTokenCount`, `usageMetadata.candidatesTokenCount` |
+| Bedrock   | `inputTokenCount`, `outputTokenCount` (root level) |
+| Azure     | Same as OpenAI |
+
+This proposal adds a mapping library that:
+
+1. Takes a provider identifier and response body
+2. Parses the provider-specific JSON structure
+3. Returns a unified `TokenUsage` struct
+
```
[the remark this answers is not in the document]
yehuditkerido: Extended the comment to cover cases where total is provided in the response.

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,98 @@
...
+A library to extract token usage from AI provider responses
+and convert them to a unified format.
+
+Each provider returns token counts in different JSON structures:
+
+| Provider  | Field Path |
+|-----------|------------|
+| OpenAI    | `usage.prompt_tokens`, `usage.completion_tokens` |
+| Anthropic | `usage.input_tokens`, `usage.output_tokens` |
```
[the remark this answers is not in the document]
yehuditkerido: Extended the table to include those options.

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,98 @@
...
+
+1. Takes a provider identifier and response body
+2. Parses the provider-specific JSON structure
+3. Returns a unified `TokenUsage` struct
+
+```rust
+pub struct TokenUsage { /* fields private */ }
+
+impl TokenUsage {
+    pub fn input_tokens(&self) -> u64;
```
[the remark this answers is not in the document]
yehuditkerido: The parse function returns `Option<TokenUsage>` - either we successfully extract all values, or we return `None`. Individual fields are plain `u64`, not `Option<u64>`. This was discussed above - JSON parsing is all-or-nothing, so partial states don't occur in practice.

docs/proposals/00216_provider-token-mapping.md
```diff
@@ -0,0 +1,98 @@
...
+    pub fn output_tokens(&self) -> u64;
+    pub fn total_tokens(&self) -> u64;  // computed: input + output
+}
+```
+
+### Goals
+
+- Extract token usage from 5 AI provider response formats
+- Provide a unified `TokenUsage` representation
```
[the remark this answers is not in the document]
yehuditkerido: Added a note in the proposal clarifying the boundary with #211.

docs/proposals/00216_provider-token-mapping.md
alexsnaps: >On feature flags 

Right, didn't want these to be introduced yet, just maybe keep it in mind when layout'ing the code, so that _if we ever want to gate it_ it'd be straight forward.

docs/proposals/00216_provider-token-mapping.md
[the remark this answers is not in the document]
alexsnaps: The point being that the proposal itself isn't capturing that decision. Which makes me also wonder whether "where that `parse` function" is and its signature... but I guess that's supposed to be in #210 ? IMHO maybe these issues where then slices slightly "too thin", wdyt @shaneutt ? That being said, I've had to jump around issues within this epic #20 quite a bit anyways  ¯\_(ツ)_/¯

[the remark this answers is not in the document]
yehuditkerido: Added the function signature - that's part of #216's scope. 
As for the issue slicing, I'll defer to @shaneutt / @szedan-rh  on that one.

docs/proposals/00216_provider-token-mapping.md
shaneutt: Sounds good, will capture this as a follow up 👍

shaneutt (approved):

PR #408, opened 2026-05-28
fix(filter): merge StreamBuffer limits with largest-wins

leseb:
## Summary

- **Largest-wins merge**: When multiple filters declare `StreamBuffer`, the pipeline now takes the largest `max_bytes` value instead of the smallest, so every filter gets enough buffer to do its job. `None` (unbounded) is treated as larger than any finite limit.
- **Runtime body mode clamping**: A new `clamp_body_mode_to_ceiling` helper ensures that runtime `set_*_body_mode` calls from filter hooks cannot widen limits beyond the baseline established at pipeline build time + global body limits. This preserves the safety ceiling even with the largest-wins merge.
- **Test updates**: All unit tests updated to reflect the new semantics; no integration test changes needed (496 integration tests pass).

The pipeline-level body ceiling (`apply_body_limits`) remains the hard safety cap for memory protection — individual filters express need, the pipeline enforces the bound.

## Test plan

- [x] All 1332 filter unit tests pass
- [x] All 496 integration tests pass
- [x] Full workspace build succeeds

### DIFF
--- a/filter/src/pipeline/tests.rs
+++ b/filter/src/pipeline/tests.rs
@@ -538,7 +538,7 @@ fn body_capabilities_buffer_overrides_stream_buffer() {
 }
 
 #[test]
-fn body_capabilities_multiple_stream_buffer_takes_min() {
+fn body_capabilities_multiple_stream_buffer_merges() {
     let pipeline = make_pipeline(vec![
         Box::new(StreamBufferReleaseFilter { marker: b"A" }),
         Box::new(StreamBufferReleaseFilter { marker: b"B" }),
--- a/filter/src/pipeline/tests.rs
+++ b/filter/src/pipeline/tests.rs
@@ -552,6 +552,23 @@ fn body_capabilities_multiple_stream_buffer_merges() {
     );
 }
 
+#[test]
+fn body_capabilities_multiple_stream_buffer_largest_wins() {
+    let pipeline = make_pipeline(vec![
+        Box::new(BoundedStreamBufferFilter { max_bytes: 1024 }),
+        Box::new(BoundedStreamBufferFilter { max_bytes: 65_536 }),
+    ]);
+    let caps = pipeline.body_capabilities();
+
+    assert_eq!(
+        caps.request_body_mode,
+        BodyMode::StreamBuffer {
+            max_bytes: Some(65_536)
+        },
+        "largest StreamBuffer limit should win when merging finite limits"
+    );
+}
+
 #[tokio::test]
 async fn execute_request_body_release_propagates() {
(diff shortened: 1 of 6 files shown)


filter/src/pipeline/tests.rs
```diff
@@ -538,7 +538,7 @@ fn body_capabilities_buffer_overrides_stream_buffer() {
 }
 
 #[test]
-fn body_capabilities_multiple_stream_buffer_takes_min() {
+fn body_capabilities_multiple_stream_buffer_merges() {
     let pipeline = make_pipeline(vec![
         Box::new(StreamBufferReleaseFilter { marker: b"A" }),
         Box::new(StreamBufferReleaseFilter { marker: b"B" }),
```
shaneutt, line 544: We need a test asserting the largest-wins

protocol/src/http/pingora/handler/mod.rs
```diff
@@ -148,6 +148,29 @@ where
...
+    };
+
+    match (mode, ceiling) {
+        (BodyMode::StreamBuffer { max_bytes }, Some(limit)) => BodyMode::StreamBuffer {
+            max_bytes: Some(max_bytes.map_or(limit, |v| v.min(limit))),
+        },
+        (BodyMode::SizeLimit { max_bytes }, Some(limit)) => BodyMode::SizeLimit {
+            max_bytes: max_bytes.min(limit),
+        },
+        (m, None | Some(_)) => m,
```
shaneutt, line 179: This means a downgrade to Stream from a StreamBuffer sort of silently passes through. This is effectively correct, in that Stream mode doesn't clamp, but we should consider documenting the behavior further.

protocol/src/http/pingora/handler/request_filter/mod.rs
```diff
@@ -413,6 +414,32 @@ mod tests {
         );
     }
 
+    #[test]
```
shaneutt, line 418: We appear to be missing several test cases?

shaneutt (changes requested):
<deleted>

protocol/src/http/pingora/handler/request_filter/mod.rs
leseb: Added in 131dca4ab887cd506fe4af060589e396d43153c8

protocol/src/http/pingora/handler/mod.rs
leseb: Documented in 131dca4ab887cd506fe4af060589e396d43153c8

filter/src/pipeline/tests.rs
leseb: Added in 131dca4ab887cd506fe4af060589e396d43153c8

shaneutt (approved):

PR #410, opened 2026-05-28
feat(filter): add openai_responses_validate filter for Responses API

### LINKED ISSUE #20: Filter 0: request_validate
## Purpose

Parse and validate the incoming Responses API request. Reject invalid parameter combinations. Normalize fields. Set up shared metadata for downstream filters.

## Praxis trait methods

on_request_body — parse JSON body, validate, write metadata
request_body_mode → StreamBuffer

## Behavior

- Parse POST /v1/responses body into ResponsesRequest
- Reject: stream=true && background=true
- Reject: background=true && store=false
- If stream=false: run the full filter chain synchronously, collect all events internally, return the final ResponseResource as a single JSON response (not SSE). The filters operate the same way — stream_events still processes chunks — but output is buffered and returned as application/json instead of streamed as text/event-stream.
- Preserve instructions in request metadata (do NOT inject as system message here — in pass-through mode the inference server handles it; in conversion mode responses_proxy injects it during Responses → Chat Completions transformation)
- Validate include values against spec enum: reasoning.encrypted_content, message.output_text.logprobs. Unrecognized values will be passed through.
- Validate model (spec allows null/omitted — handle gracefully, use default if configured)
- Generate response_id (UUID)
- Generate conversation_id from request or create new
- Do NOT normalize text.format → response_format (handle text, json_object, json_schema with name, description, schema, strict)
- Validate and preserve text.verbosity (low|medium|high)
- Preserve stream_options if provided (obfuscation handling is deferred — see Deferred Items)
- Do NOT validate tool_choice against spec union:
  - String values: "none", "auto", "required"
  - Specific function: {"type": "function", "name": "..."}
  - Allowed tools: {"type": "allowed_tools", "tools": [...], "mode": "auto |required|none"}
- Preserve metadata (arbitrary key-value dict, max 16 keys, max 64-char keys, max 512-char values)
- Preserve safety_identifier (opaque string, passed to inference)
- Preserve prompt_cache_key (opaque string, passed to inference)
- Set responses.response_id, responses.conversation_id, responses.store, responses.background, responses.stream, responses.tenant_id, responses.status, responses.iteration in metadata.

## Config

```yaml
filter: request_validate
max_body_bytes: 67108864  # 64 MiB
```

## Dependencies

None.

## Reference

- [Proposal: Responses API Filters](docs/proposals/00354_responses-api-filters.md) — Filter 0
- Part of praxis-proxy/ai#93

leseb:
## Summary

Adds the `openai_responses_validate` filter — the first filter in the OpenAI Responses API pipeline (Filter 0 from the [decomposition design doc](https://docs.google.com/document/d/1vnviJdWbro571LhiDv2wtdWtGpjGU2RSDnVQSQFglXY)). Runs after the `openai_responses_format` classifier (#372), reading its metadata for parameter-combination validation. Does targeted `serde_json::Value` extraction for fields the classifier doesn't cover (`instructions`, `conversation.id`, `include`), generates response/conversation IDs, and writes `responses.*` metadata for downstream pipeline filters.

**What it validates** (parameters the proxy needs for its own operation):
- `stream=true + background=true` → reject (proxy needs to know streaming mode)
- `background=true + store=false` → reject (background processing requires persistence)

**What it skips:**
- Non-responses requests (format guard checks `openai_responses_format.format`)
- All unknown fields forwarded unchanged to the inference server

**Design decisions:**
- No typed deserialization — body parsed as `serde_json::Value` for targeted extraction only
- Reads `stream`/`store`/`background` from classifier metadata, not from body re-parse
- Body buffering handled by the upstream classifier's `StreamBuffer`; this filter declares `ReadOnly` access

Closes: praxis-proxy/ai#20
Part of: praxis-proxy/ai#93

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/examples/configs/ai/openai/responses/request-validate.yaml
+++ b/examples/configs/ai/openai/responses/request-validate.yaml
@@ -50,4 +50,4 @@ filter_chains:
         clusters:
           - name: inference
             endpoints:
-              - "127.0.0.1:3001"
+              - "127.0.0.1:8000"
--- a/examples/configs/ai/openai/responses/request-validate.yaml
+++ b/examples/configs/ai/openai/responses/request-validate.yaml
@@ -3,16 +3,15 @@
 # Requires the ai-inference feature:
 #   cargo build -p praxis --features ai-inference
 #
-# The responses_format classifier must run before openai_responses_validate.
+# The openai_responses_format classifier must run before openai_responses_validate.
 # It buffers the body, classifies the request as Responses API, and
-# promotes model/stream/store/background to responses_format.*
+# promotes model/stream/store/background to openai_responses_format.*
 # metadata that openai_responses_validate reads for parameter-combination
 # validation.
 #
-# openai_responses_validate then does targeted field extraction for fields
-# the classifier does not cover (instructions, metadata constraints,
-# conversation.id, include), generates response and conversation IDs,
-# and writes responses.* metadata for downstream pipeline filters.
+# openai_responses_validate then extracts conversation.id when present,
+# generates response and conversation IDs, and writes responses.*
+# metadata for downstream pipeline filters.
 #
 # Valid requests are forwarded to the upstream backend with the
 # original body unchanged. Invalid requests receive a 400 JSON
--- a/examples/configs/ai/openai/responses/request-validate.yaml
+++ b/examples/configs/ai/openai/responses/request-validate.yaml
@@ -1,0 +1,53 @@
+# Responses API Request Validation
+#
+# Requires the ai-inference feature:
+#   cargo build -p praxis --features ai-inference
+#
+# The responses_format classifier must run before request_validate.
+# It buffers the body, classifies the request as Responses API, and
+# promotes model/stream/store/background to responses_format.*
+# metadata that request_validate reads for parameter-combination
+# validation.
+#
+# request_validate then does targeted field extraction for fields
+# the classifier does not cover (instructions, metadata constraints,
+# conversation.id, include), generates response and conversation IDs,
+# and writes responses.* metadata for downstream pipeline filters.
+#
+# Valid requests are forwarded to the upstream backend with the
+# original body unchanged. Invalid requests receive a 400 JSON
+# error response.
+#
+# Example valid request:
+#
+#   curl -X POST http://localhost:8080/v1/responses \
+#     -H "Content-Type: application/json" \
+#     -d '{"model":"gpt-4.1","input":"Hello, world!"}'
+#
+# Example rejected request (stream + background conflict):
+#
+#   curl -X POST http://localhost:8080/v1/responses \
+#     -H "Content-Type: application/json" \
+#     -d '{"model":"gpt-4.1","input":"test","stream":true,"background":true}'
+
+listeners:
+  - name: responses-gateway
+    address: "127.0.0.1:8080"
+    filter_chains:
+      - validate-and-route
+
+filter_chains:
+  - name: validate-and-route
+    filters:
+      - filter: openai_responses_format
+        on_invalid: reject
+      - filter: request_validate
+      - filter: router
+        routes:
+          - path: "/v1/responses"
+            cluster: inference
+      - filter: load_balancer
+        clusters:
+          - name: inference
+            endpoints:
+              - "127.0.0.1:3001"
--- a/examples/configs/ai/openai/responses/request-validate.yaml
+++ b/examples/configs/ai/openai/responses/request-validate.yaml
@@ -3,13 +3,13 @@
 # Requires the ai-inference feature:
 #   cargo build -p praxis --features ai-inference
 #
-# The responses_format classifier must run before request_validate.
+# The responses_format classifier must run before openai_responses_validate.
 # It buffers the body, classifies the request as Responses API, and
 # promotes model/stream/store/background to responses_format.*
-# metadata that request_validate reads for parameter-combination
+# metadata that openai_responses_validate reads for parameter-combination
 # validation.
 #
-# request_validate then does targeted field extraction for fields
+# openai_responses_validate then does targeted field extraction for fields
 # the classifier does not cover (instructions, metadata constraints,
 # conversation.id, include), generates response and conversation IDs,
 # and writes responses.* metadata for downstream pipeline filters.
@@ -41,7 +41,7 @@ filter_chains:
     filters:
       - filter: openai_responses_format
         on_invalid: reject
-      - filter: request_validate
+      - filter: openai_responses_validate
       - filter: router
         routes:
           - path: "/v1/responses"
--- a/filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
+++ b/filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
@@ -136,7 +136,7 @@ impl HttpFilter for RequestValidateFilter {
             },
         };
 
-        if let Err(e) = validate_request(ctx, &parsed) {
+        if let Err(e) = validate_request(ctx) {
             debug!(error = %e, "request validation failed");
             return Ok(FilterAction::Reject(
                 Rejection::status(400)
--- a/filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
+++ b/filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
@@ -296,7 +296,7 @@ mod tests {
             "response_id should be set with resp_ prefix"
         );
         assert!(
-            ctx.filter_metadata.get("responses.conversation_id").is_some(),
+            ctx.filter_metadata.contains_key("responses.conversation_id"),
             "conversation_id should be set"
         );
         assert_eq!(
@@ -525,7 +525,7 @@ mod tests {
         let ctx = run_filter(r#"{"input": "Hello"}"#, &[]).await;
 
         assert!(
-            ctx.filter_metadata.get("responses.response_id").is_some(),
+            ctx.filter_metadata.contains_key("responses.response_id"),
             "response_id should still be generated"
         );
     }
--- a/filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
+++ b/filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
@@ -1,0 +1,582 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! `request_validate` filter: validate and enrich incoming Responses
+//! API requests.
+//!
+//! Expects the upstream `responses_format` classifier to have already
+//! identified this request as a Responses API request and promoted
+//! routing facts (`model`, `stream`, `store`, `background`) to
+//! `responses_format.*` metadata.
+//!
+//! This filter reads classifier metadata for parameter-combination
+//! validation, then does targeted JSON field extraction for fields
+//! the classifier does not cover (`instructions`, `metadata`,
+//! `conversation.id`, `include`). It does **not** deserialize the
+//! full body into a typed struct.
+//!
+//! # YAML
+//!
+//! ```yaml
+//! filter: request_validate
+//! ```
+
+mod validate;
+
+use std::{
+    sync::atomic::{AtomicU64, Ordering},
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+use async_trait::async_trait;
+use bytes::Bytes;
+use tracing::debug;
+
+use self::validate::validate_request;
+use crate::{
+    FilterAction, FilterError, Rejection,
+    body::{BodyAccess, BodyMode},
+    filter::{HttpFilter, HttpFilterContext},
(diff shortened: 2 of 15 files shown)


leseb:
Added skip/pr-hygiene label since this adds a new filter; self-contained change.

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// Create a filter from YAML config.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`FilterError`] if the YAML config contains unknown fields.
+    #[allow(clippy::unnecessary_wraps, reason = "signature required by FilterFactory")]
+    pub fn from_config(_config: &serde_yaml::Value) -> Result<Box<dyn HttpFilter>, FilterError> {
+        Ok(Box::new(Self {
+            counter: AtomicU64::new(0),
+        }))
```
alexsnaps: `impl Default for RequestValidateFilter` ?

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+        let id = self.generate_raw_id();
+        format!("conv_{id}")
+    }
+
+    /// Generate a raw hex ID from timestamp and counter.
+    fn generate_raw_id(&self) -> String {
+        #[allow(clippy::cast_possible_truncation, reason = "micros fit u64")]
+        let micros = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
```
alexsnaps, line 109: Not for this PR per se, but having a consistent and centralized way to deal with time across the code base might be a good thing. 
This `unwrap_or_default` makes me uneasy... This will eventually run on a system with an incorrect system clock. Assumptions like these are usually fine, but are just a matter of time (pun not intended) "until someones breaks such invariant".

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+        let seq = self.counter.fetch_add(1, Ordering::Relaxed);
+
+        format!("{micros:016x}{seq:04x}")
+    }
+}
+
+#[async_trait]
+impl HttpFilter for RequestValidateFilter {
+    fn name(&self) -> &'static str {
+        "request_validate"
```
alexsnaps: extract in `const` ? this is reused in the `write_metadata` helper if I read this right.

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+        &self,
+        ctx: &mut HttpFilterContext<'_>,
+        body: &mut Option<Bytes>,
+        end_of_stream: bool,
+    ) -> Result<FilterAction, FilterError> {
+        if !end_of_stream {
+            return Ok(FilterAction::Continue);
+        }
+
+        let chunk = body.as_deref().unwrap_or_default();
```
alexsnaps: We do expect a `None` here? Should we then not just early return in that case? The `Bytes::new()` is then supposed to fail the deserialization below? And then be rejected?

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+        };
+
+        if let Err(e) = validate_request(ctx) {
+            debug!(error = %e, "request validation failed");
+            return Ok(FilterAction::Reject(
+                Rejection::status(400)
+                    .with_header("content-type", "application/json")
+                    .with_body(Bytes::from(e.to_json_body())),
+            ));
+        }
```
alexsnaps: Stylistic, so feel free to ignore, but this could be a single expression dealing with returning a single `Ok(FilterAction::Reject)` "unifying" these two branches which are effectively dealing with the same error, but are currently disjoint (with an helper indirection).

filter/src/builtins/http/ai/openai/responses/request_validate/validate.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// Format as a JSON error response body.
+    pub(crate) fn to_json_body(&self) -> String {
+        serde_json::json!({
+            "error": {
+                "message": &self.message,
+                "type": "invalid_request_error"
+            }
+        })
+        .to_string()
```
alexsnaps: Call into the helper `reject_invalid` and avoid dups?

alexsnaps (changes requested):

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: done in https://github.com/praxis-proxy/praxis/pull/410/changes/f9c33a1d276ab6fee518dfce8d63f4935946f026

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: done in https://github.com/praxis-proxy/praxis/pull/410/changes/f9c33a1d276ab6fee518dfce8d63f4935946f026

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: done in https://github.com/praxis-proxy/praxis/pull/410/changes/f9c33a1d276ab6fee518dfce8d63f4935946f026

filter/src/builtins/http/ai/openai/responses/request_validate/validate.rs
leseb: done in https://github.com/praxis-proxy/praxis/pull/410/changes/f9c33a1d276ab6fee518dfce8d63f4935946f026

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: that's a good point, i've opened https://github.com/praxis-proxy/praxis/issues/460 to track this, thanks

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,614 @@
...
+        &self,
+        ctx: &mut HttpFilterContext<'_>,
+        body: &mut Option<Bytes>,
+        end_of_stream: bool,
+    ) -> Result<FilterAction, FilterError> {
+        if !end_of_stream {
+            return Ok(FilterAction::Continue);
+        }
+
+        let parsed = match parse_and_validate(ctx, body) {
```
nerdalert, line 154: Looks like validator is supposed to run only on OpenAI Responses API requests, but it currently runs on any JSON request that reaches it. So a Chat Completions request can get tagged as if it were a validated Responses request, with generated `responses.*` metadata. That can confuse later routing/filters because the proxy now thinks the wrong API was validated. Wdyt? ty!

  ```suggestion
          if ctx.get_metadata("responses_format.format") != Some("responses") {
              return Ok(FilterAction::Release);
          }

          let parsed = match parse_and_validate(ctx, body) {

filter/src/builtins/http/ai/openai/responses/request_validate/validate.rs
```diff
@@ -0,0 +1,158 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Validation logic for Responses API requests.
+//!
+//! Only validates parameters the proxy needs for its own operation
+//! (stream/background interaction, background/store dependency, model
```
nerdalert: Smol nit: remove `model for routing` from this comment, or update the comment to list only the checks this file performs. I think the functionality got removed mid-PR. ty!

examples/configs/ai/openai/responses/request-validate.yaml
```diff
@@ -0,0 +1,53 @@
+# Responses API Request Validation
```
shaneutt, line 1: This new example needs an integration test

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,614 @@
...
+//! API requests.
+//!
+//! Expects the upstream `responses_format` classifier to have already
+//! identified this request as a Responses API request and promoted
+//! routing facts (`model`, `stream`, `store`, `background`) to
+//! `responses_format.*` metadata.
+//!
+//! This filter reads classifier metadata for parameter-combination
+//! validation, then does targeted JSON field extraction for fields
+//! the classifier does not cover (`instructions`, `metadata`,
```
shaneutt: This mentions metadata but I think that's gone? 🤔

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,614 @@
...
+// RequestValidateFilter
+// -----------------------------------------------------------------------------
+
+/// Validates and enriches Responses API requests.
+///
+/// Reads classifier metadata for parameter-combination checks, then
+/// parses the body as [`serde_json::Value`] for targeted field
+/// extraction. Does not deserialize the full body into a typed struct.
+///
+/// This filter has no configuration — body buffering is handled by
```
shaneutt:
```suggestion
/// This filter has no configuration, body buffering is handled by
```

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
```diff
@@ -0,0 +1,614 @@
...
+        #[allow(clippy::cast_possible_truncation, reason = "micros fit u64")]
+        let micros = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_micros()
+            .min(u128::from(u64::MAX)) as u64;
+
+        let seq = self.counter.fetch_add(1, Ordering::Relaxed);
+
+        format!("{micros:016x}{seq:04x}")
```
shaneutt, line 115: With two instances you could _technically_ end up with non-unique IDs.

shaneutt (changes requested):
Thanks @leseb!

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: this is valid but i thought i had set that condition on the yaml (i have not 🤦🏻 ) - still the yaml config won't be sufficient if an operator forgets to add the branch condition. i'll do the code update AND the config update too - thanks!

leseb: ok just went with the code fix, the config is too heavy and the best guard is the code anyways

filter/src/builtins/http/ai/openai/responses/request_validate/validate.rs
leseb: that's right, we only validate things that the gw needs, the rest will fall to the inference engine to validate (the responses params etc) - updated

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: yes that's right, i removed that mid-pr since we don't validate the "metadata" request parameter anymore

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: [4c89b54](https://github.com/praxis-proxy/praxis/pull/410/commits/4c89b54e5f00b4130fecd55cc8594ad3dbdff848)

filter/src/builtins/http/ai/openai/responses/request_validate/validate.rs
leseb: 8bf0d8b

examples/configs/ai/openai/responses/request-validate.yaml
leseb: 07c6f87a561b411eecc7e9dff096fbf4fd06363d

filter/src/builtins/http/ai/openai/responses/request_validate/mod.rs
leseb: that's right, issue #460 has some similar vibes, we can enhance it with an IdGenerator in praxis-core that all filters could re-use when they need uniqueness.

leseb: added more details in #460 and a TODO in the code to update it once #460 lands

shaneutt (approved):

PR #411, opened 2026-05-28
chore: add pre-commit hook for fmt and lint

franciscojavierarceo:
## Summary

- Adds a tracked `.hooks/pre-commit` script that verifies commit signing is enabled, then runs `make fmt` and `make lint`
- Adds `make setup-hooks` Makefile target to install the hook via symlink
- Documents setup in `docs/development.md` and `.claude/CLAUDE.md`
- Adds `.hooks/` exception to `.gitignore`

## Test plan

- [x] Ran `make setup-hooks` — symlink created
- [x] Pre-commit hook: signing check, fmt, and lint all passed
- [x] All CI checks passing

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [ "$(git config --get commit.gpgsign 2>/dev/null)" != "true" ]; then
+    echo "pre-commit: commit signing is not enabled"
+    echo "pre-commit: run: git config --global commit.gpgsign true"
+    exit 1
+fi
+
+echo "pre-commit: formatting..."
+make fmt
+
+echo "pre-commit: linting..."
+make lint
+
+echo "pre-commit: all checks passed"
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -14,6 +14,7 @@ repository.
 ## Quick Reference
 
 ```console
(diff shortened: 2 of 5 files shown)


.hooks/pre-commit
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+MAX_ADDITIONS=500
```
leseb: Thought this 1000? Anyway i don't think we should have this in pre-commit, especially for those running pre-commit locally or at least we need a way to skip this locally (without using git --no-verify)

leseb (commented):
if we introduce pre-commit i think it should take over the entire ci flow too and run in ci too, also tools version must stay in sync, today there are not and i have https://github.com/praxis-proxy/praxis/pull/416/ to fix some of that

.hooks/pre-commit
franciscojavierarceo: I thought having in pre-commit is better than CI but I can take out

franciscojavierarceo:
> if we introduce pre-commit i think it should take over the entire ci flow too and run in ci too, also tools version must stay in sync, today there are not and i have https://github.com/praxis-proxy/praxis/pull/416/ to fix some of that 

Yeah definitely agree!

.hooks/pre-commit
franciscojavierarceo: removed

shaneutt (approved):

PR #416, opened 2026-05-29
fix(makefile): pin nightly toolchain version to match CI

leseb:
## Summary

- Introduces a reusable composite action (`.github/actions/install-nightly-rust/action.yml`) that centralizes the nightly Rust toolchain installation — single source of truth for the pinned version
- The dtolnay/rust-toolchain action registers the toolchain as generic `nightly`; a second step reads the date from the action's own comment and re-registers it under the date-pinned name (e.g. `nightly-2026-03-28`) so `cargo +nightly-YYYY-MM-DD` works
- All workflows (tests, nightly, fuzz) now reference the composite action instead of duplicating the pinned SHA
- The Makefile extracts `NIGHTLY_VERSION` from the composite action file and uses `cargo +$(NIGHTLY_VERSION)` for all nightly commands (fmt, lint, fuzz)
- When the required nightly is not installed locally, the prerequisite check fails with the specific version and the `rustup` command to install it
- Dependabot continues to bump the nightly SHA in the composite action; the Makefile and the registration step both read from the updated comment automatically

## Test plan

- [ ] CI lint job passes (`cargo +nightly-2026-03-28 fmt --all -- --check`)
- [ ] CI fuzz job passes (`cargo +nightly-2026-03-28 fuzz run ...`)
- [ ] Run `make fmt` and `make lint` locally with the pinned nightly installed
- [ ] Verify `make -p 2>/dev/null | grep NIGHTLY_VERSION` resolves correctly
- [ ] Uninstall the pinned nightly and confirm `make fmt` fails with a helpful message

### DIFF
--- a/.github/actions/install-nightly-rust/action.yml
+++ b/.github/actions/install-nightly-rust/action.yml
@@ -13,3 +13,13 @@ runs:
     - uses: dtolnay/rust-toolchain@5b842231ba77f5c045dba54ac5560fed2db780e2 # nightly-2026-03-28 (rustc 1.96.0)
       with:
         components: ${{ inputs.components }}
+
+    - name: Register toolchain under date-pinned name
+      run: |
+        NIGHTLY=$(grep -oE 'nightly-[0-9]{4}-[0-9]{2}-[0-9]{2}' "${{ github.action_path }}/action.yml" | head -1)
+        if [ -n "${{ inputs.components }}" ]; then
+          rustup toolchain install "$NIGHTLY" --profile minimal --component "${{ inputs.components }}"
+        else
+          rustup toolchain install "$NIGHTLY" --profile minimal
+        fi
+      shell: bash
--- a/.github/actions/install-nightly-rust/action.yml
+++ b/.github/actions/install-nightly-rust/action.yml
@@ -1,0 +1,15 @@
+name: Install nightly Rust
+description: Install the pinned nightly Rust toolchain
+
+inputs:
+  components:
+    description: "Comma-separated list of components to install (e.g. rustfmt)"
+    required: false
+    default: ""
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -5,6 +5,7 @@
 VERSION          ?= $(shell perl -ne 'print $$1 if /^version\s*=\s*"(.+)"/' Cargo.toml)
 IMAGE            ?= praxis
 CONTAINER_ENGINE ?= $(shell command -v podman 2>/dev/null || command -v docker 2>/dev/null)
+NIGHTLY_VERSION  := $(shell grep -m1 'rust-toolchain@' .github/actions/install-nightly-rust/action.yml | grep -oE 'nightly-[0-9]{4}-[0-9]{2}-[0-9]{2}')
```
shaneutt, line 8: It almost feels like this should be going in the other direction 🤔

leseb: i like this direction since we have dependabot running and updating the nightlies automatically. if we reverse it we lose that capability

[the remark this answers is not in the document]
leseb: done

.github/actions/install-nightly-rust/action.yml
```diff
@@ -0,0 +1,25 @@
...
+runs:
+  using: composite
+  steps:
+    - uses: dtolnay/rust-toolchain@5b842231ba77f5c045dba54ac5560fed2db780e2 # nightly-2026-03-28 (rustc 1.96.0)
+      with:
+        components: ${{ inputs.components }}
+
+    - name: Register toolchain under date-pinned name
+      run: |
+        NIGHTLY=$(grep -oE 'nightly-[0-9]{4}-[0-9]{2}-[0-9]{2}' "${{ github.action_path }}/action.yml" | head -1)
```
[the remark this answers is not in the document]
leseb: done

Makefile
shaneutt: Not sure we were talking about the same thing, but nothing that's needed to block :+1:

shaneutt (approved):

PR #420, opened 2026-05-29
proposal: Anthropic Messages API filters (What + Why)

franciscojavierarceo:
## Summary

- Proposes Anthropic Messages API (`/v1/messages`) support as composable `HttpFilter` implementations
- Three capabilities: classification/routing, bidirectional format transformation with OpenAI Chat Completions, and Anthropic-native feature preservation
- Documents the complete mapping rules between Anthropic Messages and OpenAI Chat Completions (request, response, and streaming), validated against OGX's production implementation
- Scoped to What? and Why? sections per proposal process — How? section will follow in a separate PR after direction is accepted

## Context

- Extends `AiRequestFormat` enum with `AnthropicMessages` variant alongside existing `Responses` and `ChatCompletions`
- Covers integration with vLLM, llm-d, KServe/MaaS, and Anthropic API backends
- OpenAI Responses API transformation is explicitly out of scope (covered by praxis-proxy/ai#93)

## Test plan

- [ ] Proposal review and direction approval from stakeholders
- How? section PR to follow with filter design, config schema, and implementation tiers


docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,299 @@
...
+Anthropic backends, these must be preserved
+end-to-end. When routing Anthropic requests to
+OpenAI-compatible backends, the filters must
+gracefully degrade: strip unsupported fields, map
+what can be mapped, and log what was dropped.
+
+### User Stories
+
+- As a platform engineer, I want to route
+  `/v1/messages` requests to vLLM backends that
```
leseb: `to vLLM backends that only support /v1/chat/completions` -> how is that determined?

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,299 @@
...
+- As a developer, I want to send Anthropic-format
+  requests with prompt caching to a Claude backend
+  through Praxis without losing the `cache_control`
+  blocks or `anthropic-version` header.
+- As an SRE, I want to use Praxis credential
+  injection to manage `x-api-key` headers for
+  Anthropic backends the same way I manage
+  `Authorization: Bearer` headers for OpenAI
+  backends.
+- As a platform engineer running llm-d with vLLM
```
leseb: isn't that a dup of the first one?

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,299 @@
...
+(`/v1/messages`) and OpenAI Chat Completions
+(`/v1/chat/completions`).
+
+The OpenAI Responses API (`/v1/responses`) is a
+fundamentally different protocol with stateful
+semantics and is out of scope for format
+transformation. Responses API support is covered
+separately.
+
+The scope covers three capabilities:
```
leseb: should we include a validation filter? similar to the responses one?

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,299 @@
...
+   top-level `system` parameter, required
+   `max_tokens`, `anthropic-version` header, and
+   typed content blocks. Note: on Opus 4.7+ models,
+   `"role": "system"` is now allowed inside the
+   messages array as a mid-conversation system
+   message, so role-based detection alone is not
+   sufficient. The `anthropic-version` header is the
+   strongest signal.
+
+2. **Format transformation**: bidirectional conversion
```
leseb: how do we determine that translation is needed? for instance, the inference backend could support /v1/messages natively.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,299 @@
...
+     (Opus 4.7+) → no direct OpenAI equivalent;
+     map to `finish_reason: "stop"` with refusal
+     metadata preserved in response headers or
+     filter metadata
+   - Usage: `prompt_tokens` → `input_tokens`,
+     `completion_tokens` → `output_tokens`,
+     `cached_tokens` → `cache_read_input_tokens`
+   - Response ID generated as `msg_{uuid}`
+
+   **Streaming (OpenAI chunks → Anthropic SSE):**
```
leseb: i think for streaming we need another filter, just like openai responses, since SSE are specific to anthropic

docs/proposals/00415_anthropic-messages-api-filters.md
franciscojavierarceo: The request format is detected dynamically by the `anthropic_messages_format` classifier filter, which promotes `x-praxis-ai-format: anthropic_messages` to headers and filter results. The backend's supported API formats are declared per cluster in config. A branch chain routes to the passthrough or transformation path based on the match between detected format and cluster capabilities. The operator doesn't hardcode which pipeline to use — the filters determine the right path at runtime. I'll update the implementation to wire the branch chain routing up.

docs/proposals/00415_anthropic-messages-api-filters.md
franciscojavierarceo: Good call — the llm-d story is covered by the first one since llm-d is the primary deployment topology. I'll remove it.

docs/proposals/00415_anthropic-messages-api-filters.md
franciscojavierarceo: Yes — I'll add an `anthropic_request_validate` filter following the same principle as #354's `request_validate`: validate proxy-needed fields (messages non-empty, first message is user role, max_tokens present and > 0, model non-empty, valid content block types), let the backend handle the rest. Unlike the Responses API, Anthropic Messages doesn't require persistence or stateful orchestration, so the validation filter is lighter — no shared state struct, no store initialization. It just validates, extracts routing facts, and passes through.

docs/proposals/00415_anthropic-messages-api-filters.md
franciscojavierarceo: Addressed in previous comment. The classifier detects whether or not translation is needed.

docs/proposals/00415_anthropic-messages-api-filters.md
franciscojavierarceo: Agreed — I'll split streaming SSE transformation into a separate `anthropic_stream_events` filter, matching the `stream_events` pattern in #354. This keeps request body transformation decoupled from response streaming, and lets operators use SSE event handling independently (e.g. for logging or guardrails on passthrough streams without transforming the request).

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
+---
+issue: # to be created
```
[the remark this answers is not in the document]
franciscojavierarceo: Renamed file to `xxxxx_anthropic-messages-api-filters.md` as a placeholder. Will rename to match the issue number once created.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+1. **Classification and routing**: detect Anthropic
+   Messages API requests by body structure and
+   promote routing facts to headers for downstream
+   filter chains and cluster selection. The
+   classifier must distinguish Anthropic Messages
+   from OpenAI Chat Completions even though both use
+   a `messages` field, using discriminating signals:
+   top-level `system` parameter, required
+   `max_tokens`, `anthropic-version` header, and
+   typed content blocks. Note: on Opus 4.7+ models,
```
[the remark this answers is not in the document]
franciscojavierarceo: Added link to Anthropic's [mid-conversation system messages docs](https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages). Changed wording from "Opus 4.7+" to "newer Anthropic models" to avoid tying to a specific version.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+   `"role": "system"` is now allowed inside the
+   messages array as a mid-conversation system
+   message, so role-based detection alone is not
+   sufficient. The `anthropic-version` header is the
+   strongest signal.
+
+2. **Request validation**: validate proxy-needed
+   fields before forwarding, following the same
+   principle as #354's `request_validate`. Checks
+   that `messages` is non-empty, first message has
```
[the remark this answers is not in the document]
franciscojavierarceo: Agreed. Removed the first-message role check from the proxy validation. The backend enforces this constraint; the proxy only validates what it needs for its own operation (messages non-empty, max_tokens > 0, model present).

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+   `role: user`, `max_tokens` is present and > 0,
+   `model` is non-empty, and content blocks have
+   valid `type` fields. Unlike the Responses API,
+   Anthropic Messages does not require persistence
+   or stateful orchestration, so the validation
+   filter is lighter — no shared state struct, no
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed. Replaced all em-dashes with colons or restructured sentences.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+   - `temperature`, `top_p`, `top_k` → not
+     supported on Opus 4.7+ models (returns 400);
+     transformation must strip these when targeting
+     newer Claude models
+   - `thinking` blocks → dropped (no OpenAI
+     equivalent)
+   - Image blocks: Anthropic uses `type: "image"`
+     with `source.type: "base64"|"url"|"file"`;
+     OpenAI uses `type: "image_url"` with
+     `image_url.url` — requires structural mapping
```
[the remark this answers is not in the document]
franciscojavierarceo: Called out explicitly: `source.type: "file"` requires resolution to a data URL before transformation. Deferred to the How section.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+   **Response (OpenAI → Anthropic):**
+   - `message.content` string → content block
+     with `type: "text"`
+   - `tool_calls` → content block per call with
+     `type: "tool_use"` and `input = JSON-parsed
+     arguments`
+   - Finish reason: `"stop"` → `"end_turn"`,
+     `"tool_calls"` → `"tool_use"`,
+     `"length"` → `"max_tokens"`,
+     `"content_filter"` → `"end_turn"`
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed. The original `finish_reason` is now preserved in filter metadata as `openai.finish_reason` so downstream filters can distinguish safety-filtered responses from natural completions. Documented as a known lossy mapping.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+     filter metadata
+   - Usage: `prompt_tokens` → `input_tokens`,
+     `completion_tokens` → `output_tokens`,
+     `cached_tokens` → `cache_read_input_tokens`
+   - Response ID generated as `msg_{uuid}`
+
+4. **Streaming SSE transformation**: a separate
+   filter (following the `stream_events` pattern in
+   #354) that transforms streaming responses between
+   Anthropic and `OpenAI` SSE formats. Decoupled
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed. Both `OpenAI` and `Anthropic` are now consistently backticked throughout.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Goals
+
+- Validate proxy-needed fields in Anthropic Messages
+  requests (`messages` non-empty, `max_tokens` > 0,
+  `model` present) and reject malformed requests
+  with consistent error responses before they reach
+  the backend.
+- Classify Anthropic Messages API requests and
+  promote `x-praxis-api-format: anthropic_messages`
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed. Changed to `x-praxis-ai-format` to match the existing `responses_format` convention. Also clarified that the classifier extends the shared `AiRequestFormat` enum at `ai/classifier/`.

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+misidentify these as `chat_completions`, leading to
+incorrect routing or transformation failures.
+
+The format transformation filters are needed because
+real deployments mix backends:
+
+- **vLLM and llm-d** expose OpenAI-compatible
+  endpoints (`/v1/chat/completions`) and also
+  support `/v1/messages` natively, but not all
+  deployments enable Anthropic compatibility. vLLM
```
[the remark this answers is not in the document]
franciscojavierarceo: Added detail: vLLM's `/v1/messages` supports core features (system, tools, thinking, streaming SSE) but does not support `cache_control` or `budget_tokens` (accepted but ignored).

docs/proposals/00415_anthropic-messages-api-filters.md
```diff
@@ -0,0 +1,322 @@
...
+- **KServe and MaaS** (Models as a Service) provide
+  model discovery and API key management. MaaS
+  returns model URLs that clients call directly;
+  the model endpoints may implement either format.
+  MaaS uses OpenAI-compatible API keys (`sk-oai-*`)
+  and `/v1/models` for discovery.
+- **Anthropic API** is the canonical backend for
+  Claude models and uses a distinct wire format
+  with features that have no OpenAI equivalent:
+  prompt caching with `cache_control` blocks (5m
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed. Clarified that 5-minute TTL is standard; 1-hour TTL requires the `extended-cache-ttl` beta header.

docs/proposals/00415_anthropic-messages-api-filters.md
leseb: no, if you set `x-praxis-ai-format: anthropic_messages` how do you know which backend to use whether translation happens or not? what's the detected format is we route to a native backend, and what's the detected format when we need translation? as far as i can tell the format is always `x-praxis-ai-format: anthropic_messages`, the classifier just determines whether the request is anthropic messages or something else, where does it say that translation might be needed or not? (depending on whether the backend supports it)

am i missing something here?

shaneutt (approved):

docs/proposals/00415_anthropic-messages-api-filters.md
franciscojavierarceo: my bad claude did my thinking for me here.

after some clanking and looking at how OGX does the routing/translation decision, i think we need support for cluster-level declarations so pipelines know whether to passthrough or translate after deciding to route.

OGX uses a static allowlist, here we need a way for filters to identify what a cluster supports after the router select it.

It's an open question where that metadata lives (in the cluster config, a top-level registry, or something else—i created https://github.com/praxis-proxy/praxis/issues/466 to discuss this). But the point is the classifier will detect the request's format and the translation decision is based on the cluster's declared capabilities.

I'll follow up to make this dynamic after we have an agreement on #466.

leseb: ok answering my own question. this is already addressed by the existing classifier + router pattern. the responses_format filter classifies the request body and promotes facts to internal headers (x-praxis-ai-format, x-praxis-ai-model, x-praxis-ai-stream). the router matches those headers for cluster selection, and any downstream filter can read them too.

for more details see this example: https://github.com/leseb/praxis/blob/8bf0d8b98511ec2bc1fdbd2179751301e7f0047b/examples/configs/ai/openai/responses/format-routing.yaml

PR #427, opened 2026-05-31
docs: add proposal for token count injection into FilterContext (#212)

abdallahsamabd:
## Summary

Adds proposal `00212_token-count-filter-context.md` with the What? and Why?
sections for exposing token usage fields on `HttpFilterContext`.

This is part of the Token Counting epic (#20), sub-task praxis-proxy/ai#221. The proposal
defines the motivation for a shared contract (typed `Option<u64>` fields)
that decouples token count producers (#210, praxis-proxy/ai#84, praxis-proxy/ai#87, #219) from consumers
(#214, rate limiting, cost tracking).

The How? section will be submitted in a follow-up PR after the direction
is accepted, per the proposal process.

## Related

- Epic: praxis-proxy/ai#71
- Task: praxis-proxy/ai#221

### DIFF
--- a/docs/proposals/00212_token-count-filter-context.md
+++ b/docs/proposals/00212_token-count-filter-context.md
@@ -1,0 +1,73 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/212
+epic: https://github.com/praxis-proxy/praxis/issues/20
+status: proposed
+authors:
+  - abdallahsamabd
+stakeholders:
+  - shaneutt
+  - szedan-rh
+---
+
+# Token Count Injection into FilterContext
+
+## What?
+
+Add typed token usage fields (`token_input`, `token_output`,
+`token_total`) to `HttpFilterContext` so that downstream filters
+can access token counts without coupling to provider-specific
+parsing logic. This is the shared contract that enables the
+entire Token Counting epic (#20).
+
+### Goals
+
+- Expose three `Option<u64>` fields on `HttpFilterContext` for
+  input, output, and total token counts.
+- Persist token values across Pingora lifecycle phases via
+  matching fields on `PingoraRequestCtx`.
+- Provide a convenience method `set_token_usage()` that sets
+  all three fields and mirrors values to `filter_metadata`.
+- Support both full updates (from provider response parsing)
+  and partial updates (from pre-request estimation).
+- Require zero new dependencies; pure struct/method additions.
+
+## Why?
+
+### Motivation
+
+The Token Counting epic (#20) adds token usage awareness to
+Praxis for AI inference workloads. Multiple filters need to
+produce and consume token counts:
+
+- **Producers:** response JSON parser (#210), SSE streaming
+  parser (#211), client-side estimator (#219), multi-provider
+  mapper (#216).
+- **Consumers:** response header injector (#214), future
+  token-based rate limiting, cost tracking, access logging.
+
+Today these filters have no shared, typed location to exchange
+token data. Without a well-defined contract, each filter would
+need to know about every other filter's implementation details,
+violating the pipeline's composability.
+
+This task defines the **interface contract**: typed fields on
+the per-request context that any filter can write to or read
+from. It decouples producers from consumers and enables
+independent, parallel development of all other epic sub-tasks.
+
+### User Stories
+
+- As a filter author implementing token counting (#210), I want
+  a well-defined place to store extracted token counts so that
+  downstream filters can access them without parsing response
+  bodies themselves.
+- As a filter author implementing token-based rate limiting, I
+  want typed `u64` fields on the context so that I can read
+  token counts without string parsing or key-name guessing.
+- As a filter author implementing client-side estimation (#219),
+  I want to set only `token_input` before the upstream responds
+  so that admission control can make pre-request decisions.
+- As a proxy operator, I want token counts available in
+  `filter_metadata` so that access log templates and branch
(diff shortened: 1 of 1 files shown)


docs/proposals/00212_token-count-filter-context.md
```diff
@@ -0,0 +1,73 @@
...
+Add typed token usage fields (`token_input`, `token_output`,
+`token_total`) to `HttpFilterContext` so that downstream filters
+can access token counts without coupling to provider-specific
+parsing logic. This is the shared contract that enables the
+entire Token Counting epic (#20).
+
+### Goals
+
+- Expose three `Option<u64>` fields on `HttpFilterContext` for
+  input, output, and total token counts.
```
alexsnaps, line 25: Based of #407, currently the input & output tokens would be made available atomically (i.e. you'd only get to both, or none at all). With #219 in mind, should there be a discriminator on the input tokens (i.e. something either beyond `Option<u64>`, or a dedicated field) for "estimated input tokens"?

abdallahsamabd: the atomicity from https://github.com/praxis-proxy/praxis/pull/407 gives us a natural discriminator: if token_output is Some, both values came from the provider (always returned together). If token_output is None but token_input is Some, it's an estimate.
In practice, set_token_usage() overwrites any estimate atomically once the provider responds — consumers in the response phase always see actuals.

Does that implicit check (token_output.is_some()) address your concern, or is there a scenario where an explicit discriminator is needed?

alexsnaps: I'll be honest, I don't know  about any scenario per se ¯\_(ツ)_/¯ 
But as you rightly said:

>the atomicity from https://github.com/praxis-proxy/praxis/pull/407 gives us a natural discriminator: if `token_output` is `Some`, both values came from the provider (always returned together). 

That's discriminating to  _YOU_ as the caller into "whatever" return the `None`. i.e. anything then querying these fields, users of the API would then need to query both fields and then draw a conclusion from the individual values. Raising questions around atomicity of even reading these 2 fields. Again, _this might not be an issue for the use cases we looking at addressing_. But I don't know. 

One I _could be making up_, is being able to know what the estimate was against the actual... as I guess on estimate, eventually the input field would be overwritten, yes?

alexsnaps: >consumers in the response phase always see actuals.

Is that true? consumers in the response phase "only enter that phase" when the response has been possibly entirely buffered?

abdallahsamabd: 1. Yes, the estimate gets overwritten by set_token_usage() when the actual arrives. If comparing estimate vs actual becomes a real need (e.g. estimator accuracy tracking), we can add a token_input_estimated field later without breaking the existing API. Keeping it minimal for now


2. Not exactly. There are two response phases:
a)  on_response — fires when response headers arrive (body not yet available)
b)  on_response_body — fires per body chunk, or after full buffering if StreamBuffer is active

Token fields are set during on_response_body (the token counting filter needs the full body to parse usage JSON). So a consumer reading token values must also operate in on_response_body, not on_response.

The buffering is guaranteed when token counting is in the pipeline because the token count filter declares response_body_mode() -> StreamBuffer, which forces full buffering before on_response_body(end_of_stream=true) fires. Response body filters execute in reverse pipeline order, so a consumer placed before the token counter in the YAML config sees the values (reverse = counter runs first in the body phase).

So: consumers see token values reliably, but they must be response-body-phase filters, not response-header-phase filters. This is a constraint worth documenting.

abdallahsamabd:
@alexsnaps 
Could you reopen this? The bot closed it due to some reason....

alexsnaps:
@shaneutt would the bot support it, if we'd reopen PRs? Or is the workflow meant to be for a new one to be opened?

docs/proposals/00212_token-count-filter-context.md
alexsnaps: >Token counting would require `StreamBuffer`

Well, I disagree with that statement, but I think it is a fair statement nonetheless given the existing. So let's say we keep it as a base assumption for now 👌 

But we could (in some distant & hypothetical  future) eventually support retrieving these bits of data from a partial JSON document and make them available sooner. As of now tho, I do seem to also see that's not where the spirit is with these filters tho. So... tl;dr 👍

abdallahsamabd:
not sure, but as you know if we open new PR then comments will be lost...

shaneutt:
Normally the idea is that it is re-opened, not a new one created. Sorry if this is a bit clunky, it's still being refined.

alexsnaps:
>Normally the idea is that it is re-opened, not a new one created. Sorry if this is a bit clunky, it's still being refined.

No worries, I did mention that's all still being tweaked. Now tho, no one seem to be able to re-open, or is there a `/reopen`-alike command enabling that?

shaneutt:
Everyone in @praxis-proxy/praxis-devs should be able to do it... others have done it before... :thinking: 

I'm looking into this, sorry about the problems. Ping me if things are getting stuck.

alexsnaps:
Wait, maybe my bad... now it's disabled because of the one PR per user rule 🤦 
I can indeed reopen [this one](https://github.com/praxis-proxy/praxis/pull/391), so either some other rule was the issue, or ... I somehow was blind to the button _not_ being disabled. Sorry @shaneutt

Issue #264, opened 2026-05-15
ext_proc HttpFilter wrapper


shaneutt:
Integrate ext_proc gRPC callout into the filter pipeline for request and response phases.

## Requirements

- Implement `HttpFilter` trait wrapping ext_proc gRPC client
- Invoke external processor at configured request/response phases
- Map filter pipeline phases to ext_proc processing phases
- Pass request/response data to external processor per processing mode
- Apply mutations returned by external processor

## Implementation Notes

- Filter under `filter/src/builtins/http/payload_processing/`
- Each filter phase maps to corresponding ext_proc phase

Parent epic: #17

Issue #263, opened 2026-05-15
ext_proc gRPC client


shaneutt:
gRPC client for connecting to external processing services, managing bidirectional stream lifecycle.

## Requirements

- Connect to external processing service via gRPC
- Manage bidirectional streaming RPC lifecycle
- Support Envoy ext_proc protocol messages
- TLS support for gRPC connection
- Configurable connection timeout and request timeout

## Implementation Notes

- Use `tonic` for gRPC client
- Leverage protobuf definitions from `praxis-proto` crate (ext-proc feature)
- Stream lifecycle: open on first request phase, close after response complete

Parent epic: #17

PR #428, opened 2026-05-31
feat(ext-proc): implement gRPC callout for request and response headers

usize:
## Summary

Implement the gRPC callout for the `ext_proc` filter, replacing the
skeleton no-op with a functional client that sends request and response
headers to an external processor and applies the returned mutations.

- Open a bidirectional `Process` stream per callout phase (request/response)
- Send `RequestHeaders` / `ResponseHeaders` with pseudo-headers matching Envoy convention
- Apply `HeaderMutation` to request (`extra_request_headers`) or response (`HeaderMap`)
- Handle `ImmediateResponse` as `FilterAction::Reject`
- Timeout and gRPC errors propagate as `FilterError` (pipeline `failure_mode` handles policy)
- Each callout opens its own stream; persistent full-duplex streaming is a follow-up

Addresses praxis-proxy/praxis#263 and praxis-proxy/praxis#264 from epic praxis-proxy/praxis#17.

## New files

- `callout.rs` — stream lifecycle, `CalloutError`, response dispatch
- `mutations.rs` — proto-to-Praxis conversions, header mutation application

## Test plan

- [x] 22 unit tests for mutation/conversion functions
- [x] 5 gRPC integration tests with mock `ExternalProcessor` server
  - request header mutation round-trip
  - response header mutation round-trip
  - `ImmediateResponse` rejection
  - no-op response
  - timeout behavior
- [x] Verify `make lint` passes
- [x] Verify container build succeeds

### DIFF
--- a/filter/ext-proc/src/callout.rs
+++ b/filter/ext-proc/src/callout.rs
@@ -1,0 +1,172 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! gRPC stream management for the `ext_proc` filter.
+//!
+//! Opens a bidirectional `Process` stream to the external processor,
+//! sends a single [`ProcessingRequest`], and receives a single
+//! [`ProcessingResponse`] within a configurable timeout.
+//!
+//! [`ProcessingRequest`]: praxis_proto::envoy::service::ext_proc::v3::ProcessingRequest
+//! [`ProcessingResponse`]: praxis_proto::envoy::service::ext_proc::v3::ProcessingResponse
+
+use std::time::Duration;
+
+use futures::stream;
+use praxis_filter::{FilterAction, FilterError, HttpFilterContext};
+use praxis_proto::envoy::service::ext_proc::v3::{
+    ProcessingRequest, ProcessingResponse, external_processor_client::ExternalProcessorClient, processing_request,
+    processing_response,
+};
+use tonic::transport::Channel;
+
+use crate::{
+    Phase,
+    mutations::{apply_headers_response, immediate_to_rejection, request_to_proto_headers, response_to_proto_headers},
+};
+
+// -----------------------------------------------------------------------------
+// CalloutError
+// -----------------------------------------------------------------------------
+
+/// Errors that can occur during a gRPC callout.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum CalloutError {
+    /// gRPC transport or protocol error.
+    #[error("ext_proc gRPC error: {0}")]
+    Grpc(#[from] tonic::Status),
+
+    /// The per-message timeout expired.
+    #[error("ext_proc message timeout")]
+    Timeout,
+
+    /// The server closed the stream without sending a response.
+    #[error("ext_proc server closed stream without response")]
+    EmptyStream,
+}
+
+// -----------------------------------------------------------------------------
+// Public callout functions
+// -----------------------------------------------------------------------------
+
+/// Send request headers to the external processor and apply mutations.
+///
+/// Opens a `Process` stream, sends a `RequestHeaders` message, and
+/// waits for one response within `timeout`. Returns [`FilterAction`]
+/// indicating whether the pipeline should continue or reject.
+pub(crate) async fn process_request_headers(
+    channel: Channel,
+    target: &str,
+    timeout: Duration,
+    ctx: &mut HttpFilterContext<'_>,
+) -> Result<FilterAction, FilterError> {
+    let headers = request_to_proto_headers(ctx);
+    let request = ProcessingRequest {
+        request: Some(processing_request::Request::RequestHeaders(headers)),
+        ..Default::default()
+    };
+
+    let response = send_and_receive(channel, request, timeout, target).await?;
+    dispatch_response(&response, ctx, Phase::Request)
+}
+
+/// Send response headers to the external processor and apply mutations.
+///
+/// Same pattern as [`process_request_headers`] but wraps
+/// `ResponseHeaders` and operates during the response phase.
+pub(crate) async fn process_response_headers(
+    channel: Channel,
+    target: &str,
+    timeout: Duration,
+    ctx: &mut HttpFilterContext<'_>,
+) -> Result<FilterAction, FilterError> {
+    let headers = response_to_proto_headers(ctx);
+    let request = ProcessingRequest {
+        request: Some(processing_request::Request::ResponseHeaders(headers)),
+        ..Default::default()
+    };
+
+    let response = send_and_receive(channel, request, timeout, target).await?;
+    dispatch_response(&response, ctx, Phase::Response)
+}
+
+// -----------------------------------------------------------------------------
+// Private helpers
+// -----------------------------------------------------------------------------
+
+/// Open a `Process` stream, send one request, and receive one response.
+///
+/// Each callout opens its own stream. The timeout covers the entire
+/// round-trip (stream open + send + receive).
+async fn send_and_receive(
+    channel: Channel,
+    request: ProcessingRequest,
+    timeout: Duration,
+    target: &str,
+) -> Result<ProcessingResponse, FilterError> {
+    let result = tokio::time::timeout(timeout, async {
+        let mut client = ExternalProcessorClient::new(channel);
+        let request_stream = stream::once(async { request });
+        let response = client.process(request_stream).await.map_err(CalloutError::Grpc)?;
+        let mut streaming = response.into_inner();
+
+        streaming
+            .message()
+            .await
+            .map_err(CalloutError::Grpc)?
+            .ok_or(CalloutError::EmptyStream)
+    })
+    .await;
+
+    match result {
+        Ok(Ok(response)) => Ok(response),
+        Ok(Err(e)) => {
+            tracing::warn!(target = %target, error = %e, "ext_proc callout failed");
+            Err(e.into())
+        },
+        Err(_elapsed) => {
+            tracing::warn!(target = %target, "ext_proc callout timed out");
+            Err(CalloutError::Timeout.into())
+        },
+    }
+}
+
+/// Route a [`ProcessingResponse`] variant to the correct mutation handler.
+///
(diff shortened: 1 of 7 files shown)


usize:
For llm-d-router compatibility ( see path B in praxis-proxy/ai#169 ) we need full duplex mode, which this is not. However, it builds us a client and the necessary scaffolding we need to build on. Setting up a stack of PRs to follow up, to implement that mode.

filter/ext-proc/src/callout.rs
```diff
@@ -0,0 +1,172 @@
...
+///
+/// Returns [`FilterAction::Continue`] for header mutations or
+/// [`FilterAction::Reject`] for immediate responses. Unexpected
+/// response types produce a [`FilterError`].
+fn dispatch_response(
+    response: &ProcessingResponse,
+    ctx: &mut HttpFilterContext<'_>,
+    phase: Phase,
+) -> Result<FilterAction, FilterError> {
+    let Some(resp) = &response.response else {
```
nerdalert, line 233: `max_message_timeout_ms` is accepted in config, but `override_message_timeout` is not handled here. A processor can legally respond with only `override_message_timeout`; Envoy would extend the timer and keep waiting, but this code treats the missing response oneof as `Continue` and lets the request proceed. Couple of options are either reject non-default `max_message_timeout_ms` until this is implemented, or handle the override by restarting the timeout and reading the next response? Ty!

filter/ext-proc/src/mutations.rs
```diff
@@ -0,0 +1,264 @@
...
+    let mut headers = Vec::new();
+
+    headers.push(HeaderValue {
+        key: ":method".to_owned(),
+        value: ctx.request.method.as_str().to_owned(),
+        raw_value: Vec::new(),
+    });
+    headers.push(HeaderValue {
+        key: ":path".to_owned(),
+        value: ctx.request.uri.path().to_owned(),
```
nerdalert: `:path` drops the query string by using `uri.path()`. HTTP/Envoy-style `:path` should include both path and query, so `/search?q=secret` is sent to the processor as only `/search`. That’s a security issue because the external processor may make allow/deny decisions without seeing important query parameters.Ty!

```suggestion
    value: ctx.request
        .uri
        .path_and_query()
        .map_or(ctx.request.uri.path(), http::uri::PathAndQuery::as_str)
        .to_owned(),
```

filter/ext-proc/src/mutations.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/// Apply header mutations to the upstream request.
+///
+/// Adds headers via [`HttpFilterContext::extra_request_headers`].
+/// Pseudo-headers (`:` prefix) are skipped because Praxis sets
+/// method and path through dedicated context fields. Removals
+/// are logged and skipped: request headers are immutable by
+/// filter time in the Praxis pipeline.
+pub(crate) fn apply_request_header_mutation(mutation: &HeaderMutation, ctx: &mut HttpFilterContext<'_>) {
```
nerdalert, line 124: This makes request-side ext_proc mutations effectively append-only, even though the filter exposes Envoy-style remove and overwrite behavior.

Request header mutations are not being applied with the available request mutation queues. `remove_headers` is skipped entirely, and all `set_headers` are pushed through  `extra_request_headers`, so Envoy append/overwrite/remove semantics are not honored. Since `HttpFilterContext` already has `request_headers_to_set` and `request_headers_to_remove`, can we map the ext_proc mutation actions into those queues and add tests for remove, overwrite-if-exists, add-if-absent, and append/default behavior?

Without this, processors can appear to request header removal or overwrite, but Praxis will still forward the original request headers.

I'm sure you have dug into this more than I but here is a suggested codex snippet if it helps, ty!


```rust
  pub(crate) fn apply_request_header_mutation(mutation: &HeaderMutation, ctx: &mut HttpFilterContext<'_>) {
      use praxis_proto::envoy::service::common::v3::header_value_option::HeaderAppendAction;

      for name in &mutation.remove_headers {
          if is_pseudo_header(name) {
              continue;
          }
          if let Ok(header_name) = http::HeaderName::try_from(name.as_str()) {
              ctx.request_headers_to_remove.push(header_name);
          }
      }

      for hvo in &mutation.set_headers {
          let Some(hv) = &hvo.header else {
              continue;
          };
          if is_pseudo_header(&hv.key) {
              continue;
          }

          let Ok(name) = http::HeaderName::try_from(hv.key.as_str()) else {
              continue;
          };
          let value = header_value_string(hv);
          let Ok(header_value) = http::HeaderValue::from_str(&value) else {
              continue;
          };

          let will_be_removed = ctx.request_headers_to_remove.iter().any(|n| n == &name);
          let exists_after_removal = !will_be_removed && ctx.request.headers.contains_key(name.as_str());

          match hvo.append_action {
              x if x == HeaderAppendAction::AddIfAbsent as i32 => {
                  if !exists_after_removal {
                      ctx.extra_request_headers.push((Cow::Owned(hv.key.clone()), value));
                  }
              },
              x if x == HeaderAppendAction::OverwriteIfExistsOrAdd as i32 => {
                  ctx.request_headers_to_set.push((name, header_value));
              },
              x if x == HeaderAppendAction::OverwriteIfExists as i32 => {
                  if exists_after_removal {
                      ctx.request_headers_to_set.push((name, header_value));
                  }
              },
              _ => {
                  if let Some(existing) = ctx.request.headers.get(name.as_str())
                      && let Ok(existing_str) = existing.to_str()
                      && !will_be_removed
                  {
                      let combined = format!("{existing_str},{value}");
                      if let Ok(combined_value) = http::HeaderValue::from_str(&combined) {
                          ctx.request_headers_to_set.push((name, combined_value));
                          continue;
                      }
                  }

                  ctx.extra_request_headers.push((Cow::Owned(hv.key.clone()), value));
              },
          }
      }
  }
```

filter/ext-proc/src/mutations.rs
```diff
@@ -0,0 +1,349 @@
...
+        if let Some(hv) = &hvo.header {
+            if is_pseudo_header(&hv.key) {
+                continue;
+            }
+            let value = header_value_string(hv);
+            if let (Ok(name), Ok(val)) = (http::HeaderName::try_from(&hv.key), http::HeaderValue::try_from(&value)) {
+                if should_append(hvo) {
+                    resp.headers.append(name, val);
+                } else {
+                    resp.headers.insert(name, val);
```
elevran: The `else` branch here collapses `OverwriteIfExistsOrAdd`, `OverwriteIfExists`, and `AddIfAbsent` into an unconditional `headers.insert`. Compare with `dispatch_request_header` on the request side, which correctly handles all four variants (e.g,,`OverwriteIfExists` should skip the insert when the header is absent).

Consider switching `set_response_headers` to use `resolve_append_action` with a four-way match mirroring `dispatch_request_header`.

filter/ext-proc/src/lib.rs
```diff
@@ -512,11 +533,35 @@ impl ExtProcFilter {
...
+
+    /// Convert a callout error into a rejection with [`status_on_error`].
+    ///
+    /// On success the action passes through unchanged. On error the
+    /// processor failure is logged and a [`FilterAction::Reject`] is
+    /// returned with the configured status code, matching Envoy's
+    /// error-handling behaviour.
+    ///
+    /// [`status_on_error`]: ExtProcConfig::status_on_error
+    fn call_or_reject(&self, result: Result<FilterAction, FilterError>) -> FilterAction {
```
elevran, line 559: All callout errors are converted to `FilterAction::Reject(status_on_error)` here, so the filter always returns `Ok(Reject(...))` on processor failure. Because `failure_mode` is pipeline-level and only fires when a filter returns `Err(FilterError)`, a pipeline configured with `failure_mode: open` can never produce fail-open behavior for ext_proc callout errors: the pipeline sees a clean `Reject`, not an error.

If the intent is for `status_on_error` to take precedence (matching Envoy's `status_on_error` semantics), a note in the ext_proc config doc comment would prevent user confusion: that `status_on_error` always applies to processor errors regardless of the pipeline-level `failure_mode`.

filter/ext-proc/src/callout.rs
```diff
@@ -0,0 +1,239 @@
...
+/// read. Without a configured `max_timeout`, override requests are
+/// ignored and the response is returned as-is.
+async fn send_and_receive(
+    channel: Channel,
+    request: ProcessingRequest,
+    timeout: Duration,
+    max_timeout: Option<Duration>,
+    target: &str,
+) -> Result<ProcessingResponse, FilterError> {
+    let deadline = tokio::time::Instant::now() + timeout;
```
elevran, line 113: The outer `timeout_at(deadline, ...)` uses a fixed absolute deadline set here. `receive_with_override` runs as a future inside this outer deadline and adds a relative `timeout(override_dur, ...)` for the second message. Both timers race, so whichever fires first wins.

For `override_message_timeout` to be useful it needs to replace the deadline rather than race against it. One approach: compute a new absolute deadline when the override is received (`Instant::now() + override_dur.min(max)`) and use a second `timeout_at` for the subsequent `next_message` call, removing the outer wrapper once the override fires.

filter/ext-proc/src/tests.rs
```diff
@@ -715,12 +726,1412 @@ max_message_timeout_ms: -1
...
+        other => panic!("expected Reject, got {other:?}"),
+    };
+    assert_eq!(
+        rejection.status, 503,
+        "rejection status should match configured status_on_error"
+    );
+}
+
+#[tokio::test]
+async fn grpc_override_timeout_extends_deadline() {
```
elevran, line 1931: nit: this test uses `timeout=5s` and `override_ms=500ms` with the mock sending both messages back-to-back (no delay). The test validates message plumbing but not timing: it would pass even if the override mechanism were completely broken. The timing issue described in `send_and_receive` is not caught here.

A more meaningful case: `timeout=100ms`, server delays 200 ms before the real response, `override_dur=2s`, `max_timeout=5s`. Without the override the callout times out; with it the response arrives and the header is present.

shaneutt (approved):

PR #454, opened 2026-06-01
WIP: docs(proposals): add proposal for storage layer

rikatz:
**Storage Layer Proposal:**

* Introduces two pluggable storage backend traits for Praxis: a key-value trait for small, low-latency lookups, and an object store trait for large, higher-latency blob storage, each with reference implementations for in-memory and distributed backends.
* Specifies backend requirements including multi-tenancy, per-entry TTLs, encryption at rest, size limits, and configurable failure semantics to ensure robust, secure, and isolated storage.
* Provides a reference schema for conversation and response persistence, supporting use cases like conversation rehydration, caching, and multi-step workflows, while allowing other consumers to define their own schemas.
* Clarifies that the proposal does not replace the existing in-memory `KvBackend` cache, does not define domain-specific APIs, and does not make Praxis a general-purpose database or storage proxy.
* Includes motivation,

### DIFF
--- a/docs/proposals/00412_storage_layer.md
+++ b/docs/proposals/00412_storage_layer.md
@@ -1,0 +1,220 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/412
+status: proposed
+authors:
+  - rikatz
+graduation_criteria:
+  - How? section with requirements and design
+  - Storage trait API reviewed by stakeholders
+  - Reference schema for conversation/response storage
+stakeholders:
+  - shaneutt
+  - tbd
+---
+
+# Storage Layer
+
+## What?
+
+Pluggable storage backend traits for Praxis, enabling
+filters to persist and retrieve state across requests
+and across proxy instances. This proposal defines two
+storage abstractions — one for key-value lookups and
+one for object/blob storage — along with the contract
+each backend must satisfy (multi-tenancy, TTL,
+encryption, size limits) and reference implementations
+for common backends.
+
+This proposal is about internal storage for proxy
+operation, not exposing storage backends to clients.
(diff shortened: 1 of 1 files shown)


shaneutt (commented):
<deleted>

docs/proposals/00412_storage_layer.md
```diff
@@ -0,0 +1,220 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/412
+status: proposed
```
shaneutt:
```suggestion
status: holding # WIP but no priority yet, we expect to come back to this soon
```

docs/proposals/00412_storage_layer.md
```diff
@@ -0,0 +1,220 @@
...
+authors:
+  - rikatz
+graduation_criteria:
+  - How? section with requirements and design
+  - Storage trait API reviewed by stakeholders
+  - Reference schema for conversation/response storage
+stakeholders:
+  - shaneutt
+  - tbd
```
shaneutt:
```suggestion
  - twghu
```

docs/proposals/00412_storage_layer.md
```diff
@@ -0,0 +1,220 @@
...
+---
+
```
shaneutt, line 14:
```suggestion

> **Important**: This proposal is currently WIP and on hold, we'll try and get back to this at a later time and move it forward.

```

shaneutt (dismissed):
I know this is draft/wip, but I'm inclined to take it as is for now and just mark it with caveats so we can store it until we're ready to revisit, as I expect we wont be revisiting it for some time but I want to save the content.

PR #462, opened 2026-06-02
Proposal/token count filter context

abdallahsamabd:
## Summary

Adds proposal `00212_token-count-filter-context.md` with the What? and Why?
sections for exposing token usage fields on `HttpFilterContext`.

This is part of the Token Counting epic (#20), sub-task praxis-proxy/ai#221. The proposal
defines the motivation for a shared contract (typed `Option<u64>` fields)
that decouples token count producers (#210, praxis-proxy/ai#84, praxis-proxy/ai#87, #219) from consumers
(#214, rate limiting, cost tracking).

The How? section will be submitted in a follow-up PR after the direction
is accepted, per the proposal process.

## Related

- Epic: praxis-proxy/ai#71
- Task: praxis-proxy/ai#221


abdallahsamabd:
discussion were done in this PR 
https://github.com/praxis-proxy/praxis/pull/427, it was closed automatically...
as a result of that I opened a new proposal PR (current one), since we didn't have the option to reopen the original PR

docs/proposals/00212_token-count-filter-context.md
```diff
@@ -0,0 +1,73 @@
...
+  - shaneutt
+  - szedan-rh
+---
+
+# Token Count Injection into FilterContext
+
+## What?
+
+Add typed token usage fields (`token_input`, `token_output`,
+`token_total`) to `HttpFilterContext` so that downstream filters
```
shaneutt, line 17: Unfortunately adding these to `HttpFilterContext` isn't quite appropriate as AI is an optional feature flag in Praxis, and the core proxy wouldn't need this. We'll need to think on some other way we can provide this information 🤔

shaneutt (changes requested):

docs/proposals/00212_token-count-filter-context.md
abdallahsamabd: @shaneutt  
Alternative approach: Use the existing filter_metadata: HashMap<String, String> as the transport. A token counting filter (compiled only under the ai feature flag) would write well-known keys like token.input, token.output, token.total into metadata. Consumers (rate limiters, access log templates, cost trackers) read from the same map using get_metadata().
This keeps the core struct untouched — no new fields, no feature-gate complexity. The contract becomes a set of documented key conventions rather than struct fields. We could define those key names as constants in the AI module.

Tradeoffs:
Values are strings rather than typed u64, so consumers need a .parse() call — but I think this is negligible....

Does that direction work, or do you have a different mechanism in mind?

shaneutt: Using the metadata is probably the best approach currently: mcp, a2a and responses already set a precedent for that. :+1:

docs/proposals/00212_token-count-filter-context.md
```diff
@@ -0,0 +1,73 @@
...
+issue: https://github.com/praxis-proxy/praxis/issues/212
+discussion: https://github.com/praxis-proxy/praxis/issues/20
+status: proposed
+authors:
+  - abdallahsamabd
+stakeholders:
+  - shaneutt
+  - szedan-rh
+---
+
```
shaneutt:
```suggestion
> **Important**: This proposal is currently WIP and on hold, we'll try and get back to this at a later time and move it forward.
```

shaneutt (dismissed):
Let's get this in as a WIP so we can return back to it later

PR #463, opened 2026-06-02
docs(filter): add build-time filter registry proposal

leseb:
## Summary

Proposal for a build-time auto-registration mechanism that lets external filter crates self-register into the `FilterRegistry`. Operators add a `Cargo.toml` dependency and write YAML config with zero Rust code changes to Praxis. Uses `build.rs` + `cargo metadata` discovery with a `[package.metadata.praxis-filters]` convention and an `export_filters!` macro.

## Test plan

- [ ] Review proposal structure and content
- [ ] Validate design against existing `register_filters!` and `ext-proc` patterns
- [ ] Confirm `cargo metadata` approach handles workspace members and external deps


shaneutt (approved):
This seems reasonable to park and keep considering as a future need, thank you @leseb

PR #480, opened 2026-06-04
feat: add local A2A task routing

nerdalert:
Adds local A2A task-ownership routing to the a2a filter. When task routing is enabled, Praxis records which backend cluster created a task from SendMessage JSON responses and routes follow-up task operations back to that cluster through a reserved internal header and the existing router.

  When an A2A `SendMessage` response creates a task, the `a2a` filter records which backend cluster produced that task. Follow-up task operations such as `GetTask`, `CancelTask`, `SubscribeToTask`, and push-notification config methods can then be routed back to the owning backend by injecting a reserved internal route header before the `router` filter runs.

  This PR adds:

  - `task_routing` config for the existing `a2a` filter
  - In-process `LocalTaskRouteStore` with TTL and terminal-task TTL support
  - Response-body task route capture for successful non-SSE `SendMessage` responses
  - Follow-up task route lookup by task ID
  - Reserved internal route header injection on lookup hit
  - Route-miss behavior that continues to normal router fallback
  - Spoofing protection by requiring route headers to use the reserved `x-praxis-a2a-*` namespace
  - Bounded byte-safe response accumulation using hex scratch storage in `filter_metadata`
  - Oversized response handling that skips capture and passes the response through unchanged
  - Lazy expiry cleanup for stale task routes
  - Example config for A2A task routing

  ## Behavior

  With task routing enabled:

  1. A client sends `SendMessage`.
  2. Praxis routes the initial request normally, typically by method or path.
  3. The selected backend returns a JSON-RPC response containing a task ID.
  4. The `a2a` filter captures `task_id -> cluster` locally.
  5. A later `GetTask` or `CancelTask` request for that task injects `x-praxis-a2a-route-cluster`.
  6. The `router` can match that internal header and send the request back to the task-owning backend.

  ### Streaming Scope

  SSE/streaming responses pass through unchanged. That behavior is already supported by the existing A2A classifier path and is not changed by this PR.

  This PR intentionally captures task IDs only from non-streaming `SendMessage` JSON responses. It does not capture task ownership from `SendStreamingMessage` yet, because that requires incrementally parsing SSE frames across chunked response bodies while preserving the response bytes exactly. Streaming task-route capture is planned as a follow-up PR stack mapped out in the [agentic stack protocol epic](https://github.com/praxis-proxy/ai/issues/148#issuecomment-4368239729).

Until that follow-up lands, a streaming `SendStreamingMessage` can create a task on the backend, but Praxis will not learn task ownership from the SSE stream. A later `GetTask` for that task will therefore follow the router’s static fallback route unless some other route rule matches it.

  ## Manual Validation

  A manual curl-based demo for reviewers  is available [here](https://github.com/nerdalert/praxis-research-spikes/blob/main/demo/a2a-task-routing/demo-output.md)

  The demo runs Praxis in front of two A2A demo agents and verifies:

  - `SendMessage` routes to `agent-a`
  - Praxis captures the returned task ID and owning cluster
  - follow-up `GetTask` for that task routes back to `agent-a`
  - unknown `GetTask` falls through to `agent-b`
  - client spoofing of `x-praxis-a2a-route-cluster` is rejected
  - streaming/SSE pass-through remains unchanged, with streaming task capture deferred to the follow-up PR

  ## Safety

  - Client-supplied reserved internal headers are rejected before routing.
  - `task_routing.route_cluster_header` must use the `x-praxis-a2a-` prefix.
  - Response capture is bounded by `max_response_body_bytes`.
  - Oversized responses continue through the proxy unchanged and do not create task routes.
  - Response body accumulation is byte-safe across arbitrary HTTP chunk boundaries.
  - Invalid JSON or message-only responses do not create task routes.

  ## Tests

  Added unit coverage for:

  - Task routing config defaults and validation
  - Reserved route header validation
  - Route lookup hit/miss behavior
  - Route metadata on hit
  - Task route store TTL, lazy expiry, terminal removal, replacement, and invalid IDs
  - Task route extraction from A2A task response shapes
  - Routable vs non-routable A2A methods
  - Mixed-case SSE content type handling
  - Oversized response pass-through
  - Multibyte UTF-8 split across chunks

  Added integration coverage for:

  - Capture from `SendMessage`, then route `GetTask` back to the creating backend
  - Message-only responses not creating task routes
  - Lookup miss falling through to router fallback
  - Client spoofing of the internal route header being rejected
  - Chunked JSON response capture
  - SSE pass-through with task routing enabled
  - Example config validation and routing behavior

  ## References

  Refs praxis-proxy/ai#148
  Refs praxis-proxy/ai#213

### DIFF
--- a/examples/configs/ai/a2a-task-routing.yaml
+++ b/examples/configs/ai/a2a-task-routing.yaml
@@ -1,0 +1,79 @@
+# A2A Task Routing
+#
+# Captures task ownership from non-streaming JSON SendMessage responses
+# and routes follow-up task operations back to the backend cluster that
+# created the task.
+#
+# Local task routing is useful for development, single-process deployments,
+# and proving the routing path. Multi-replica task routing requires the
+# follow-up Redis/ValKey state store.
+#
+# Flow:
+# 1. SendMessage is routed to agent-a by static router rules.
+# 2. The a2a filter captures task IDs from the JSON response body
+#    and records task_id -> cluster_name in a local in-process store.
+# 3. Follow-up GetTask, CancelTask, SubscribeToTask, or push-notification
+#    config calls with a known task ID inject x-praxis-a2a-route-cluster,
+#    which the router matches to select the owning cluster.
+# 4. Unknown task IDs follow the static fallback route.
+
+listeners:
+  - name: a2a
+    address: "127.0.0.1:8080"
+    filter_chains: [a2a-task-routing]
+
+filter_chains:
+  - name: a2a-task-routing
+    filters:
+      - filter: a2a
+        max_body_bytes: 65536
+        on_invalid: continue
+        headers:
+          method: x-praxis-a2a-method
+          family: x-praxis-a2a-family
+          task_id: x-praxis-a2a-task-id
+          streaming: x-praxis-a2a-streaming
+        task_routing:
+          enabled: true
+          store: local
+          route_cluster_header: x-praxis-a2a-route-cluster
+          ttl_seconds: 3600
+          terminal_ttl_seconds: 300
+          max_response_body_bytes: 65536
+
+      - filter: router
+        routes:
+          # Task route hit: route to the cluster that owns the task
+          - path_prefix: "/"
+            headers:
+              x-praxis-a2a-route-cluster: "agent-a"
+            cluster: "agent-a"
+          - path_prefix: "/"
+            headers:
+              x-praxis-a2a-route-cluster: "agent-b"
+            cluster: "agent-b"
+
+          # Initial SendMessage goes to the primary agent
+          - path_prefix: "/"
+            headers:
+              x-praxis-a2a-method: "SendMessage"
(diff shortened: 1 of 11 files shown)


examples/configs/ai/a2a-task-routing.yaml
```diff
@@ -0,0 +1,79 @@
+# A2A Task Routing
+#
+# Captures task ownership from non-streaming JSON SendMessage responses
+# and routes follow-up task operations back to the backend cluster that
+# created the task.
+#
+# Local task routing is useful for development, single-process deployments,
+# and proving the routing path. Multi-replica task routing requires the
+# follow-up Redis/ValKey state store.
```
shaneutt:
```suggestion
# This example uses `local` storage mode. This is useful for development and
# testing or single-instance deployments of Praxis, but multi-replica task routing
# requires the an external state store (e.g. Valkey).
```

shaneutt (approved):

PR #486, opened 2026-06-05
docs: disambiguate routing vs pipelining terminology

franciscojavierarceo:
## Summary

- Adds a "Terminology: Routing vs Pipelining" section to CLAUDE.md distinguishing runtime cluster selection (routing) from config-time chain composition (pipelining)
- Clarifies the router → load_balancer handoff in the filter reference (router sets `ctx.cluster`, load balancer reads it and sets `ctx.upstream`)
- Adds pipelining/routing distinction callouts in the architecture overview and filter system docs

## Context

"Routing" was overloaded to mean two architecturally different things, which caused confusion during Anthropic Messages API work. The original author agreed "pipelining" is a better term for the config-time concept, since the code already uses `FilterPipeline` and `resolve_pipelines`.

## Test plan

- [ ] Docs-only change, no code affected
- [ ] Verify terminology is consistent across the four changed files


.claude/CLAUDE.md
```diff
@@ -223,12 +223,28 @@ Branches rejoin at configurable points (next,
 terminal, named filter, re-entrance with iteration
 limits).
 
+## Terminology: Routing vs Pipelining
+
+These two concepts are distinct. Do not conflate them.
```
shaneutt:
```suggestion
These two concepts are distinct, take care to not conflate them.
```

shaneutt (dismissed):

PR #483, opened 2026-06-04
feat(filter): classify CRUD requests by method + path pattern

leseb:
## Summary

Extends `openai_responses_format` to classify Responses API requests by method + path pattern in `on_request_body`, not just body inspection. GET, DELETE, and POST sub-resource paths (`/v1/responses/{id}`, `/v1/responses/{id}/input_items`, `/v1/responses/{id}/cancel`, `/v1/responses/input_tokens`, `/v1/responses/compact`) are classified as `format=openai_responses` without parsing the body. `POST /v1/responses` (create) continues to use body classification. Classification happens in `on_request_body` so promoted headers flow through the StreamBuffer pre-read pipeline and are visible to downstream filters like the router.

Closes praxis-proxy/ai#11

## Test plan

- [ ] 79 unit tests pass (15 classifier path tests + 12 filter-level path tests + existing body tests)
- [ ] 25 integration tests pass (5 new GET/DELETE/branch routing tests + existing)
- [ ] `make lint` passes
- [ ] Verify GET/DELETE requests route correctly via header matching
- [ ] Verify POST /v1/responses still classifies via body inspection

PR #505, opened 2026-06-08
docs: expand combinator style guidance in CLAUDE.md

leseb:
## Summary

- Expands the combinator style guideline in CLAUDE.md to explicitly list `filter()` and `map()` and state the general principle: prefer `Option`/`Result` combinator chains over `if/else` blocks when the logic is a linear transform

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -100,7 +100,10 @@ coding style guide. Key points:
 - Prefer `to_owned()` over `to_string()` for
   `&str` to `String`
 - Use inline format args: `format!("{var}")`
-- Use let-chains, `is_some_and()`, `strip_prefix()`
+- Use let-chains, `is_some_and()`, `strip_prefix()`,
+  `filter()`, `map()`; prefer `Option`/`Result`
+  combinator chains over `if/else` blocks when the
+  logic is a linear transform
 - Reference-style rustdoc links, not inline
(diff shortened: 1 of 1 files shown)


leseb:
Based on this comment https://github.com/praxis-proxy/praxis/pull/483#discussion_r3357827402

leseb:
> ## PR Review
> **Summary**: Expands the combinator style bullet in `.claude/CLAUDE.md` to explicitly list `filter()` and `map()` and state the general principle of preferring `Option`/`Result` combinator chains over `if/else` for linear transforms.
> 
> **Overall**: Clean, minimal documentation change. The guidance is sound and consistent with the project's existing idiomatic-Rust preferences. One consistency concern noted below.
> 
> Severity	Count
> Medium	1
> ### Findings without inline placement
> * **[Medium]** The monorepo-level `CLAUDE.md` (at the repo root) contains the same bullet point (`Use let-chains, `is_some_and()`, `strip_prefix()`) but is not updated by this PR. This creates a divergence between the two files -- the project-level `.claude/CLAUDE.md`will list`filter()`, `map()`, and the combinator-chain preference, while the monorepo `CLAUDE.md`will not. Consider updating the root`CLAUDE.md` in the same PR to keep them in sync, or note that the project-level file is authoritative and the root is intentionally a subset.

slop - there is no top-level CLAUDE.md.

shaneutt (approved):

PR #587, opened 2026-06-11
feat(filter): add shared AI classifier and anthropic_messages_format filter

franciscojavierarceo:
## Summary

- Extracts AI request format classification into a shared `classifier` module at `filter/src/builtins/http/ai/classifier/mod.rs`, adding the `AnthropicMessages` variant to `AiRequestFormat`
- Adds the `anthropic_messages_format` filter that classifies Anthropic Messages API requests and promotes format facts to internal headers for downstream routing
- Updates `openai/responses` to import from the shared classifier instead of its local copy

**Behavioral changes from extraction:** The shared classifier adds `prompt` object detection (requests with a `prompt` object but no `input` now classify as `Responses` instead of `UnknownJson`). New fields (`has_tools`, `has_prompt_id`, `max_tokens`) are extracted into `ClassifiedRequest` for Anthropic consumption; existing `openai_responses_format` users are unaffected.

Example configs and integration tests are in praxis-proxy/praxis#592 (PR 6 of this stack).

Part 1 of the Anthropic Messages API filter stack (epic praxis-proxy/ai#103).

## Test plan

- [x] `cargo test -p praxis-proxy-filter` -- all unit tests pass including classifier and messages_format tests
- [x] `make lint` -- clippy and fmt clean
- [x] `make test-unit` -- no regressions across workspace


franciscojavierarceo:
looks like the Coding Conventions are failing due to the `PRAXIS_BOT` token being expired.

>Run labels=$(gh pr view "$PR_NUMBER" \
>HTTP 401: Bad credentials (https://api.github.com/graphql)
>Try authenticating with:  gh auth login -h github.com

filter/src/builtins/http/ai/anthropic/messages_format/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+        };
+
+        let mut classified = classify_request_body(bytes);
+
+        let has_anthropic_header = ctx.request.headers.get(ANTHROPIC_VERSION_HEADER).is_some();
+        let is_messages_path = is_anthropic_messages_path(ctx.request.uri.path());
+
+        if classified.format == AiRequestFormat::ChatCompletions && (has_anthropic_header || is_messages_path) {
+            classified.format = AiRequestFormat::AnthropicMessages;
+        }
```
leseb, line 141: why do we have to call the other classifier? why not rely on our own classifier for messages?

filter/src/builtins/http/ai/classifier/mod.rs
```diff
@@ -0,0 +1,813 @@
+// SPDX-License-Identifier: MIT
```
leseb, line 1: i like the idea of a generic filter that can handle all known formats!

filter/src/builtins/http/ai/classifier/mod.rs
```diff
@@ -0,0 +1,813 @@
...
+}
+
+/// Check for Anthropic-specific structural signals beyond `max_tokens`.
+///
+/// Returns true if any of:
+/// - Top-level `system` field is present as a string or array (Anthropic separates system from messages; `OpenAI` puts
+///   it in the messages array)
+/// - Any message in `messages` has typed content blocks (array of objects with a `type` key, e.g. `[{"type": "text",
+///   ...}]`)
+fn has_anthropic_signals(obj: &serde_json::Map<String, serde_json::Value>) -> bool {
```
leseb, line 185: do we also need to handle CRUD GET / DELETE?

leseb (changes requested):
Thanks! A few observations:

* Splitting into two PRs would have been easier (one for the reworked classifier and one for the message format
* we are missing integration tests for the new filters
* we are missing YAML examples for the new filters 


EDIT: i just noticed https://github.com/praxis-proxy/praxis/pull/592 but i think it's better to have CI validating new config and tests in the same PR :)

filter/src/builtins/http/ai/anthropic/messages_format/mod.rs
franciscojavierarceo: The shared classifier handles all the body parsing and error cases (`NonJson`, `InvalidJson`, `UnknownJson`, etc.), if we wrote our own we'd duplicate all that. The override covers the case where the body looks like `ChatCompletions` but the request context (anthropic-version header or /v1/messages path) tells us it's actually Anthropic.

filter/src/builtins/http/ai/classifier/mod.rs
franciscojavierarceo: No, `/v1/messages` doesn't have CRUD operations, it's stateless. LMK if you meant something else though.

franciscojavierarceo:
> Thanks! A few observations:
> 
> * Splitting into two PRs would have been easier (one for the reworked classifier and one for the message format
> * we are missing integration tests for the new filters
> * we are missing YAML examples for the new filters
> 
> EDIT: i just noticed praxis-proxy/praxis#592 but i think it's better to have CI validating new config and tests in the same PR :)

Yeah, that makes sense that splitting would have been cleaner.

As you saw the integration tests and examples are in subsequent PRs. I wanted to try to keep the changes setup in moderate scope to not overload the amount of stuff I was going to ask you to review, but LMK if you want me to include them here.

filter/src/builtins/http/ai/classifier/mod.rs
leseb: oh so this is just POST?

franciscojavierarceo: yeah that's all Anthropic supports rn https://platform.claude.com/docs/en/api/messages

leseb (approved):

PR #588, opened 2026-06-11
feat(filter): add anthropic_validate filter

franciscojavierarceo:
## Summary

- Adds the `anthropic_validate` filter for Anthropic Messages API requests.
- Validates proxy-owned request envelope requirements only: body present, valid JSON, and top-level JSON object.
- Leaves backend-owned Anthropic API semantics such as required fields, parameter types/ranges, role ordering, and content block validation to the inference backend.
- Adds `examples/configs/ai/anthropic/request-validate.yaml` plus functional example tests proving valid requests are forwarded, malformed JSON is rejected, and backend-owned semantics are forwarded.
- Documents the classifier wrapper usage in the example and the `max_body_bytes` default rationale.

Part 2 of the Anthropic Messages API filter stack for epic praxis-proxy/ai#103. praxis-proxy/praxis#587 has merged, so this PR now stacks directly on `main`. The broader AI inference validation principle is being captured separately in praxis-proxy/praxis#624.

## Review notes

- Scope is intentionally small and self-contained: envelope validation plus the matching validate example only.
- No passthrough headers, path rewriting, transformation, streaming, or docs live here.
- The downstream branches have been rebased onto this PR after the praxis-proxy/praxis#587 merge.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo +nightly fmt --check --all`
- [x] `cargo test -p praxis-proxy-filter --lib -- anthropic::validate`
- [x] `cargo test -p praxis-tests-integration --features ai-inference --test suite -- anthropic_validate`
- [x] `cargo test -p praxis-tests-integration --features ai-inference --test suite -- anthropic`
- [x] `cargo xtask sync-example-readme`
- [x] `cargo xtask lint-example-tests`
- [x] `cargo test -p praxis-tests-schema`

PR #589, opened 2026-06-11
feat(filter): add anthropic_messages_protocol filter

franciscojavierarceo:
## Summary

- Adds the `anthropic_messages_protocol` filter for native Anthropic Messages API backends.
- Supplies a gateway-managed default `anthropic-version` header when internal or non-SDK callers omit it, while preserving caller-provided versions.
- Validates the configured default as a legal HTTP header value.
- Coordinates with the existing `credential_injection` filter for proxy-managed `x-api-key` injection; this filter does not manage credentials itself.
- Adds `examples/configs/ai/anthropic/messages-protocol.yaml` plus a functional example test proving the default version header reaches the upstream backend.

Stacks on praxis-proxy/praxis#588. Part 3 of the Anthropic Messages API filter stack for epic praxis-proxy/ai#103.

## Review notes

- Scope is intentionally limited to Anthropic Messages protocol header normalization for native `/v1/messages` backends.
- This is separate from classifier promotion: `anthropic_messages_format` detects request format and promotes facts; `anthropic_messages_protocol` mutates upstream request headers when protocol defaults are centrally managed by the gateway.
- No path rewriting or body transformation lives here; those remain separate PRs.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo +nightly fmt --check --all`
- [x] `cargo test -p praxis-proxy-filter --lib -- anthropic::protocol anthropic_messages_protocol`
- [x] `cargo test -p praxis-tests-integration --features ai-inference --test suite -- anthropic --nocapture`
- [x] `cargo xtask sync-example-readme`
- [x] `cargo xtask lint-example-tests`
- [x] `cargo test -p praxis-tests-schema`


filter/src/builtins/http/ai/anthropic/protocol/config.rs
```diff
@@ -0,0 +1,47 @@
...
+    DEFAULT_VERSION.to_owned()
+}
+
+// -----------------------------------------------------------------------------
+// Config Validation
+// -----------------------------------------------------------------------------
+
+/// Validate the parsed configuration.
+pub(crate) fn build_config(cfg: AnthropicPassthroughConfig) -> Result<AnthropicPassthroughConfig, FilterError> {
+    if cfg.default_version.is_empty() {
```
leseb, line 45: how can this be empty since we return DEFAULT_VERSION in default_version()?

franciscojavierarceo: bad config

franciscojavierarceo: see the `invalid_header_value_rejected` test

leseb (changes requested):
my first reaction is why do you need a new filter for this when you can promote things from the classifier? but later i see you're injecting the protocol and thus mutating the content. i'm not sure i understand the scenario, the yaml example calls the chain `anthropic-native` and then use `anthropic_messages_format` which led me to think an anthropic client sent the request. if so, wouldn't the version be part of the header already?

i'm not opposed to the new filter, i just need more clarification.

my last point is that the filter name is misleading, i'd rename it for something like `anthropic_message_protocol`

thanks!

filter/src/builtins/http/ai/anthropic/protocol/config.rs
leseb: ah right, empty string or something passes by the user

franciscojavierarceo:
> my first reaction is why do you need a new filter for this when you can promote things from the classifier? but later i see you're injecting the protocol and thus mutating the content. i'm not sure i understand the scenario, the yaml example calls the chain `anthropic-native` and then use `anthropic_messages_format` which led me to think an anthropic client sent the request. if so, wouldn't the version be part of the header already?
> 
> i'm not opposed to the new filter, i just need more clarification.
> 
> my last point is that the filter name is misleading, i'd rename it for something like `anthropic_message_protocol`
> 
> thanks!

this was actually in the ADR https://github.com/praxis-proxy/praxis/blob/main/docs/proposals/00484_anthropic-messages-api-filters.md?plain=1#L420 as `anthropic_passthrough` but i updated the the proposal since the name was misleading.

i renamed this to `anthropic_messages_protocol` and tightened the docs/example around the intended use case: gateway-managed protocol header defaults for internal or non-SDK callers hitting native `/v1/messages` backends.

the existing `anthropic-version` headers are preserved and the filter only supplies the configured default when missing. the classifier only detects/promotes format facts, this filter is the request-header mutation step.

leseb (approved):
all clear!

PR #610, opened 2026-06-16
docs: add request-scoped filter extensions proposal

leseb:
## Summary

Adds proposal 00432 for a generic `RequestExtensions` TypeMap container on `HttpFilterContext`. This provides type-safe, request-scoped state that persists across all Pingora lifecycle phases, replacing the `Arc<Mutex<ResponsesState>>` approach from praxis-proxy/ai#93.

The What? and Why? sections only, per the proposal lifecycle. The How? section follows in a separate PR after direction is accepted.

First consumer: Responses API filters (#354). Future consumers: MCP session state, protocol negotiation, caching metadata.

## Test plan

- [ ] Stakeholder review of proposal direction
- [ ] Agreement on goals and non-goals

### DIFF
--- a/docs/proposals/00432_responses-state.md
+++ b/docs/proposals/00432_responses-state.md
@@ -1,0 +1,118 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/354
+discussion: # XXX
+status: proposed
+authors:
+  - leseb
+graduation_criteria:
+  - RequestExtensions API reviewed by stakeholders
+  - Cross-phase survival tests pass for all writeback sites
+  - First consumer integrated (ResponsesState)
+stakeholders:
+  - shaneutt
+  - nerdalert
+  - twghu
+  - jojosnegros
+---
+
+# Request-Scoped Filter Extensions
+
+## What?
+
+Add a generic, type-safe, request-scoped extension
+container (`RequestExtensions`) to `HttpFilterContext`.
+Filters store and retrieve arbitrary typed values that
+persist across all Pingora lifecycle phases (request,
+request body, response, response body, logging).
+
+The container follows the TypeMap pattern used by tower
+and axum for passing arbitrary state through request
+contexts. The framework has no knowledge of what
+filters store in it. The cost when unused is an empty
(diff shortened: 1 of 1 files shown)


shaneutt (changes requested):
I understand the motivation, and I generally agree with the need.

What's problematic here is that we need to have a more holistic solution for this problem domain in time (see #412).

This seems to be proposing a solution for #412 in part or in whole, but I want us to build something that is (for now) more bespoke and specific to responses rather than generic so that it can soak and over time influence the holistic solution.

Let me know if that makes sense, and what your thoughts are? 🤔

leseb:
> I understand the motivation, and I generally agree with the need.
> 
> What's problematic here is that we need to have a more holistic solution for this problem domain in time (see praxis-proxy/ai#147).

i think this generic solution is already holistic (although i can't see the future). i understand that the concern seems to be about timing, not direction and committing to an API change shouldn't be taken lightly. but i feel like it's acceptable since we are not v1.

> 
> This seems to be proposing a solution for praxis-proxy/ai#147 in part or in whole, but I want us to build something that is (for now) more bespoke and specific to responses rather than generic so that it can soak and over time influence the holistic solution.

choosing a contained approach in the "responses store" filter is what delays the holistic solution.
the question is whether the soak period creates more tech debt than it avoids. when i look at what a sidecar approach would cost (a lot of guardrails) i think it could be detrimental, a lot of things need to be taken care of that we get for "free" with the current proposal.

> 
> Let me know if that makes sense, and what your thoughts are? 🤔

i still think that the generic solution is worth trying, i feel like this is what the holistic solution will end up being anyways - at least for the state part. what's more generic than TypeMap on a filter context? conceptually i think this plays well for other filters. like A2A already hex-encodes sse state into metadata strings as a workaround. other filters can benefit from this mechanism.

shaneutt (approved):
I'm OK with moving forward, I just want to see our touch on core be as light as is feasible, since I am working on a more general approach to storage and state in #412 which will inevitably result in some refactors throughout.

PR #613, opened 2026-06-16
fix(test-utils): verify database queryable before declaring postgres ready

### LINKED ISSUE #612: fix(test-utils): PostgreSQL integration test flaky due to container readiness race
## Summary

The `response_store_persists_response_to_postgres` integration test fails intermittently in CI with:

```
relation "openai_responses_0_threadid_2_" does not exist
```

The test was introduced in #605 and has been failing consistently on CI runners while passing locally.

## Root Cause

`wait_for_postgres` uses `pg_isready` to check container readiness. `pg_isready` returns success when the PostgreSQL server accepts connections, but this can occur before the Docker entrypoint script finishes creating the `POSTGRES_DB` database. On CI runners under load, this race window is large enough for the proxy's lazy store initialization to connect before the database exists, causing the store init to fail silently (logged as a warning, request still proxied to backend with 200). The subsequent direct SQL query then finds no table.

## Fix

Add a second readiness phase that verifies the target database is actually queryable via `psql -d praxis -c 'SELECT 1'` after `pg_isready` succeeds. This ensures the Docker entrypoint has completed database creation before tests proceed.

franciscojavierarceo:
## Summary

- Fix flaky `response_store_persists_response_to_postgres` test (#612)
- `pg_isready` can return success before the Docker entrypoint finishes creating `POSTGRES_DB`; add a second readiness phase that runs `psql SELECT 1` against the target database
- Extract polling loop into `poll_until` helper to stay within clippy line limit

## Test plan

- [x] Both `openai_response_store_postgres` tests pass locally with fix
- [x] `cargo clippy -p praxis-test-utils -- -D warnings` clean
- [x] `cargo +nightly fmt -- --check` clean
- [x] CI integration tests pass (this PR specifically targets the PostgreSQL test step)

Fixes #612

### DIFF
--- a/tests/utils/src/net/postgres.rs
+++ b/tests/utils/src/net/postgres.rs
@@ -207,11 +207,34 @@ fn wait_for_postgres(engine: &str, container_id: &str) {
 
 /// Repeatedly call `check` until it returns `true` or the deadline passes.
 fn poll_until(deadline: Instant, check: impl Fn() -> bool) {
-    while Instant::now() < deadline {
+    loop {
         if check() {
             return;
+        }
+        if Instant::now() >= deadline {
+            break;
         }
         thread::sleep(READY_POLL_INTERVAL);
     }
     panic!("`PostgreSQL` container did not become ready within {READY_TIMEOUT:?}");
+}
+
+#[cfg(test)]
+mod tests {
+    use std::cell::Cell;
+
+    use super::*;
+
+    #[test]
+    fn poll_until_checks_readiness_after_sleep_crosses_deadline() {
+        let attempts = Cell::new(0);
+
+        poll_until(Instant::now() + Duration::from_millis(1), || {
+            let next_attempt = attempts.get() + 1;
+            attempts.set(next_attempt);
+            next_attempt == 2
+        });
+
+        assert_eq!(attempts.get(), 2);
+    }
 }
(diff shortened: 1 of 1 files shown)


tests/utils/src/net/postgres.rs
```diff
@@ -172,21 +172,46 @@ fn detect_container_engine() -> String {
...
+            ])
+            .output()
+            .is_ok_and(|o| o.status.success())
+    });
+}
+
+/// Repeatedly call `check` until it returns `true` or the deadline passes.
+fn poll_until(deadline: Instant, check: impl Fn() -> bool) {
+    while Instant::now() < deadline {
+        if check() {
```
shaneutt, line 211: This _technically_ introduces a small race condition in that after the `thread::sleep()` the `<` operator could return `false`, even though the sleep time was enough that the `check()` would have returned `true`.

Given what it's used for and where it's at it's certainly a small thing. However you could switch `while {}` to a `loop {}` and do the `check()` followed by deadline, followed by sleep to make it more logically consistent.

shaneutt (dismissed):

tests/utils/src/net/postgres.rs
franciscojavierarceo: done!

shaneutt (approved):

Issue #238, opened 2026-05-15
CEL policy engine integration


shaneutt:
## Summary

Embed a CEL (Common Expression Language) evaluator for complex filter conditions, if the CEL approach is chosen in the spike evaluation.

## Scope

- Integrate a Rust CEL evaluator (e.g., `cel-interpreter` crate)
- Define the CEL environment: available variables (request headers, path, method, status, body fields)
- Compile CEL expressions at config load time; cache compiled programs
- Sandbox CEL evaluation: no side effects, bounded execution time
- Bridge CEL evaluation results to the filter condition system
- Unit tests and doctests for CEL condition evaluation
- Example configs demonstrating CEL expressions

## Acceptance Criteria

- CEL expressions can be used in filter conditions
- CEL programs are compiled and cached at config load time
- Invalid CEL expressions produce clear config validation errors
- CEL evaluation is sandboxed with bounded execution time
- Integration with the existing condition system is seamless

## Context

Depends on spike evaluation outcome. Related: #158. Parent epic: #189.

shaneutt:
Not prioritizing a CEL policy engine at this time. This depends on the CEL vs predicate DSL spike (#232) completing first. We'll revisit in a future iteration after that decision is made.

PR #615, opened 2026-06-17
feat: add CPEX policy filter

### LINKED ISSUE #614: Spike: CPEX filter
## Summary

Add a `cpex` filter that embeds the CPEX policy runtime in-process in Praxis. CPEX is the core framework for policy enforcement and extensibility in ContextForge, so a native CPEX filter brings Praxis to parity with ContextForge on the policy and enforcement layer, rather than only proxying to it.

CPEX lets operators define authorization flows declaratively and turn decisions into ordered effects: redact a field, delegate a token, taint a session, call a PDP or rules engine in process, write an audit record.

The motivation are authorization patterns that are difficult to express with stateless decision engines:

- **Cross-call session reasoning** — an agent reads compensation data then sends an email; each call is fine alone, but a session label carried across calls blocks the exfiltration.
- **Post-result enforcement** — evaluate the actual tool response, not just the pre-invoke request.
- **Ordered effects** — `on_allow` / `on_deny` drive concrete actions in a defined order, not just a boolean.

## Requirements

- Support ContextForge's plugin runtime and configuration natively, bridged by the CPEX filter
- Support CPEX Authorization Policy Logic (APL) with:    
    - In-process Cedar and CEL PDPs for authorization logic
    - Effects: deny, redact, taint, delegate, run
- Translate Praxis filter context into CPEX Common Message Format + Typed Extensions for policy evaluation (provides common vocabulary/input schema for writing policies)
- Unit tests plus an end-to-end demo in the praxis-demos repo.

## Relation to the epic

This delivers ContextForge's enforcement layer directly in Praxis. It may subsume or reframe parts of the routing and compatibility subtasks (#349 to praxis-proxy/ai#163), since parity on policy enforcement no longer depends on proxying to a ContextForge backend.

araujof:
## Summary

Adds a `cpex` HTTP filter that embeds the CPEX policy runtime in-process. CPEX is ContextForge's core framework for policy enforcement and extensibility, so a native filter brings Praxis to parity on the enforcement layer rather than only
proxying to it. Policies are declarative per-route chains that turn decisions into ordered effects: deny, redact, delegate, taint, or run a plugin, with an in-process PDP as one step in the chain.

The filter is feature-gated (`--features cpex-policy-engine`) and runs the CPEX / APL (Authorization Policy Logic) runtime as linked Rust crates. No sidecar, no FFI.

### Why use this filter?

The filter lets operators define authorization flows declaratively and turn decisions into ordered effects: redact a field, delegate a token, taint a session, call a PDP or rules engine in process, write an audit record.

The motivation are authorization patterns that are difficult to express with stateless decision engines:

- **Cross-call session reasoning** — an agent reads compensation data then sends an email; each call is fine alone, but a session label carried across calls blocks the exfiltration.
- **Post-result enforcement** — evaluate the actual tool response, not just the pre-invoke request.
- **Ordered effects** — `on_allow` / `on_deny` drive concrete actions in a defined order.

Closes praxis-proxy/praxis#614 Part of epic praxis-proxy/ai#180

## Changes

- **Filter** (`filter/src/builtins/http/security/cpex/`): identity resolution, APL route evaluation, request and response body rewriting, plugin dispatch, and session taint, wired into both the request and response phases.
- **PDPs**: in-process Cedar (`apl-pdp-cedar-direct`) and CEL (`apl-pdp-cel`), both registered. A route's `cedar:` or `cel:` step selects one.
- **Effects**: `deny('reason', 'code')`, `redact` / `mask` on `args.*` and `result.*`, `delegate(...)` (RFC 8693 token exchange), `taint(label, session)`, and `run(name)` (alias for `plugin`).
- **CMF bridge**: translates Praxis filter context (JSON-RPC tool / prompt / resource, identity, headers) into CPEX Common Message Format plus typed extensions, the input schema policies evaluate against.
- **Session taint**: `X-Session-Id` maps to `agent.session_id`, subject-scoped as `H(subject : session_id)`, so taint persists across requests and never crosses principals.
- **Config**: `CpexFilterConfig` (`config_path`, `body_access`, `require_mcp_metadata`, `init_timeout_secs`); simplified APL with an optional `apl:` wrapper.
- **Deps**: pin the `contextforge-org/cpex` crates.
- **Docs**: filter README and an example config under `examples/configs/security/`.

## Testing

- Unit tests for identity, CEL route allow and deny, session taint (within-session, per-session, cross-principal), and response-phase framing and redaction.
- `cargo test --features cpex`, `make lint`, and `clippy --features cpex -D warnings` are green.
- [End-to-end demo](https://github.com/praxis-proxy/demos/pull/1) in the praxis-demos repo: nine scenarios under both Cedar and CEL. All pass.

## Notes

Related: praxis-proxy/praxis#238 praxis-proxy/ai#144 (demo shows an end-to-end authorization flow using CEL as PDP/rules engine for authz decisions)


araujof:
Addressed praxis-bot reviews. 

- **`error.rs` fail-open (Critical):** not actually a fail-open. The body is an already-built `json!` value (serialization cannot fail in practice) and every caller is a deny path with status already committed. Hardened anyway to a `const` valid deny envelope, so it stays total and never panics.
- **Response-phase JWT re-validation:** does not apply. The filter parses no JWTs. It re-resolves identity in-process over the frozen request headers and already fails closed. No expiry step, no false deny. Unchanged.
- **Header snapshots:** `build_cmf_extensions` now snapshots once and reuses it.
- **`cargo deny`:** added the cpex git source to `allow-git`.
- **Integration test:** added a happy-path test that mints a valid HS256 JWT and asserts end-to-end pass-through (200).
- **Atomic ordering:** Acquire on load, Release on stores.
- **Unbounded buffer:** new `max_buffer_bytes` config (default 10 MiB), bounded the `read_write` StreamBuffer, with tests and docs.
- **File-level lint suppression:** replaced the `#![allow]` in `json_rpc.rs` with function-level `#[expect]` (only `too_many_lines` fires).
- **Registry test:** added the `cfg(cpex)` assertion.

shaneutt:
Some asynchronous conversations going on about this as well. Will try to provide a summary and updates on those talks here as we progress.

Cargo.toml
```diff
@@ -39,9 +39,10 @@ async-trait = "0.1.89"
 benchmarks = { path = "benchmarks" }
 bytes = "1.12.0"
 chrono = { version = "0.4.45", default-features = false, features = ["clock"] }
-dashmap = "6.2.1"
+cpex = { version = "0.2.0", git = "https://github.com/contextforge-org/cpex.git", rev = "702ab163c64fe762318cb2bc5fb9b088869dd1af", default-features = false, features = ["jwt", "oauth", "pii", "audit", "cedar", "cel", "valkey"] }
```
usize: This git-rev pin is the core of the coupling concern. 

As an external crate using `export_filters!`, this dependency would live in the CPEX filter crate's own `Cargo.toml`, not here.

deny.toml
```diff
@@ -35,4 +35,6 @@ highlight = "all"
 unknown-registry = "deny"
 unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
-allow-git = []
+# CPEX policy-runtime crates (feature-gated behind `--features cpex`).
+# Pinned by exact rev in Cargo.toml until the crates are published.
+allow-git = ["https://github.com/contextforge-org/cpex"]
```
usize: This exception exists solely because of the git-rev pin above. An external filter crate wouldn't require any changes to Praxis's `deny.toml`.

filter/src/builtins/http/security/cpex/filter.rs
```diff
@@ -0,0 +1,881 @@
...
+        // The init future is wrapped in `tokio::time::timeout` so a
+        // misbehaving plugin's `initialize()` future can't hang startup
+        // / hot-reload indefinitely. The bundled identity-jwt plugin
+        // already has its own JWKS connect/request timeouts plus
+        // soft-fail-at-boot, so this is defense-in-depth for other
+        // init paths (custom plugins, future hooks) where a future
+        // could legitimately stall.
+        let mgr_for_init = Arc::clone(&mgr);
+        let init_timeout = std::time::Duration::from_secs(cfg.init_timeout_secs);
+        let init: Result<(), String> = std::thread::spawn(move || -> Result<(), String> {
```
usize, line 156: **Impedance mismatch: sync filter factory vs async plugin init.**

CPEX's `PluginManager::initialize` is async, but Praxis's filter factory signature is sync. The workaround is spawning a dedicated OS thread to build a throwaway single-threaded tokio runtime and `block_on` the init future. The comment explains this is needed because `block_on` would panic if a runtime is already attached (which is the case under `#[tokio::test]`).

This is correct but it's a signal: when integrating a library requires spawning an OS thread to build a temporary runtime to drive a future to completion because the host's lifecycle doesn't match, the library may be better off in a crate that owns its own lifecycle.

filter/src/builtins/http/security/cpex/filter.rs
```diff
@@ -0,0 +1,881 @@
...
+
+// State of the one-shot tokio runtime-flavor check performed on the
+// first request. See `CpexFilter::on_request` for the rationale.
+
+/// Initial state — no request has been served yet.
+const RUNTIME_UNCHECKED: u8 = 0;
+/// First request saw a multi-thread runtime; subsequent requests skip the check.
+const RUNTIME_OK: u8 = 1;
+/// First request saw a current-thread runtime; all requests reject.
+const RUNTIME_REJECTED: u8 = 2;
```
usize, line 52: **Impedance mismatch: sync response body vs async hooks.**

CPEX hooks are async but `on_response_body` is sync (Pingora constraint). The filter works around this with `block_in_place` + `Handle::current().block_on()`, which requires a multi-threaded tokio runtime. This constant plus the check at lines 356–367 permanently reject `work_stealing: false` deployments.

This is a hard constraint that narrows Praxis's deployment flexibility for all users who enable the feature, driven by an architectural mismatch between CPEX and Pingora. In an external crate, this is a documented requirement of that crate rather than a constraint on Praxis's runtime model.

filter/src/builtins/http/security/cpex/filter.rs
```diff
@@ -0,0 +1,881 @@
...
+        };
+
+        let body_bytes = body.as_ref().cloned().unwrap_or_else(Bytes::new);
+        let id_str = json_rpc_id(&body_bytes);
+
+        // praxis's `on_response_body` is sync (the Pingora response_body
+        // callback can't be awaited). We're on a tokio worker so
+        // `block_in_place` lets us drive the async CMF dispatch without
+        // stalling other tasks.
+        let extensions = match tokio::task::block_in_place(|| {
```
usize: **Impedance mismatch: identity re-resolution in response phase.**

The JWT is re-validated here because CPEX's hook model doesn't carry request-phase identity state forward. If the token expires between request and response phases (slow upstream), this produces a false deny on a request that was already authorized and already processed by the upstream. The prior review flagged this, the filter treats it as "fail closed" and replaces the body with a deny envelope, but the upstream has already acted on the request.

An external crate could evolve its own solution (e.g., caching the resolved identity in filter state) without requiring changes to Praxis's filter context API.

filter/src/registry.rs
```diff
@@ -120,6 +120,8 @@ fn register_http_builtins(factories: &mut HashMap<String, FilterFactory>) {
     register_http(factories, "circuit_breaker", CircuitBreakerFilter::from_config);
     register_http(factories, "compression", CompressionFilter::from_config);
     register_http(factories, "cors", CorsFilter::from_config);
+    #[cfg(feature = "cpex")]
+    register_http(factories, "cpex", crate::CpexFilter::from_config);
```
usize: This `#[cfg(feature = "cpex")]` registration inside `register_http_builtins` is what the `export_filters!` macro replaces. An external crate using `export_filters! { http "cpex" => CpexFilter::from_config }` plus `[package.metadata.praxis-filters]` in its `Cargo.toml` would auto-register via the `build.rs` discovery system.

docs/filters/http/security/policy.md
```diff
@@ -0,0 +1,35 @@
+<!-- Generated by: cargo xtask generate-filter-docs -->
+<!-- Do not edit manually -->
+
+# `cpex`
+
```
shaneutt, line 5:
```suggestion

> **Warning**: Experimental feature, requires `cpex-policy-engine` crate feature.

```

docs/filters/http/security/policy.md
```diff
@@ -0,0 +1,35 @@
...
+## Example
+
+```yaml
+filter: cpex
+config_path: /etc/praxis/cpex.yaml
+body_access: read_write       # optional; default read_only
+require_mcp_metadata: true    # optional; default true
+init_timeout_secs: 30         # optional; default 30
+max_buffer_bytes: 10485760    # optional; default 10 MiB (read_write only)
+```
```
shaneutt, line 35: Rather than calling this `cpex.md`, I recommend renaming it to `policy.md` to generalize the scope upwards slightly.

docs/filters/reference.md
```diff
@@ -46,6 +44,7 @@ Built-in filters organized by protocol and category.
 | Filter | Feature | Description |
 |--------|---------|-------------|
 | [`cors`](http/security/cors.md) | - | Spec-compliant CORS filter implementing origin validation, preflight handling, and response header injection. |
+| [`cpex`](http/security/cpex.md) | `cpex` | Configuration block for a `cpex` filter slot in a Praxis filter chain. |
```
shaneutt:
```suggestion
| [`policy`](http/security/policy.md) | `policy` | Configuration for policies in Praxis using `cpex` policy engine |
```

examples/configs/security/cpex.yaml
```diff
@@ -0,0 +1,78 @@
+# CPEX Security Filter
```
shaneutt: Similar to above: let's rename the high level thing to policy, e.g. this file as `policy.yaml`, and `Policy Security Feature`, etc.

Keep using `cpex` when we're referring to the engine for our policies, and places like that where the engine is a focus.

examples/configs/security/cpex.yaml
```diff
@@ -0,0 +1,78 @@
...
+# `tests/integration/fixtures/cpex-policy.yaml` for the minimal
+# single-HS256-identity-plugin policy the integration test loads.
+#
+# `require_mcp_metadata: true` (the default) fail-closes when the
+# `mcp` filter is missing or ordered after `cpex` — this guards
+# against a misconfigured chain silently bypassing policy.
+#
+# This example builds only when the `cpex` cargo feature is enabled:
+#
+#   cargo run --features cpex -p praxis -- \
```
shaneutt: Let's make the feature a bit more verbose, so someone can have a good guess at what it is just by a glance at the name.

```suggestion
#   cargo run --features cpex-policy-engine -p praxis -- \
```

examples/configs/security/cpex.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - main
+
+filter_chains:
+  - name: main
+    filters:
+      - filter: mcp
+        # Default mode parses MCP JSON-RPC bodies and stashes
+        # mcp.method / mcp.name in filter_metadata. Must precede cpex.
+
+      - filter: cpex
```
shaneutt:
```suggestion
      - filter: policy
```

examples/README.md
```diff
@@ -131,6 +129,7 @@ page.
 | File | Description |
 | ------ | ------------- |
 | [cors.yaml](configs/security/cors.yaml) | Spec-compliant CORS filter with preflight handling, origin validation, and credential support |
+| [cpex.yaml](configs/security/cpex.yaml) | Embeds the CPEX policy runtime in-process to enforce multi-source JWT identity, APL route policy, RFC 8693 OAuth 2.0 token exchange, PII scanning, audit emission, and (under `body_access: read_write`) request / response body rewriting |
```
shaneutt:
```suggestion
| [policy.yaml](configs/security/policy.yaml) | Embeds the CPEX policy engine in-process to enforce multi-source JWT identity, APL route policy, RFC 8693 OAuth 2.0 token exchange, PII scanning, audit emission, and (under `body_access: read_write`) request / response body rewriting |
```

(use `runtime` if you prefer, `engine` is just how Praxis had been referring to the concept thus far)

filter/src/builtins/http/security/policy/common_message_format.rs
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
```
shaneutt, line 1: Related to above: filter name leveled up to the overall category/theme, and then the individual file names spelled out in verbosity for clarity-at-a-glance.

`filter/src/builtins/http/security/cpex/cmf.rs` => `filter/src/builtins/http/security/policy/common_message_format.rs`

filter/src/builtins/http/security/policy/config.rs
```diff
@@ -0,0 +1,128 @@
+// SPDX-License-Identifier: MIT
```
shaneutt, line 1: Since this config is specific to the cpex policy engine, we should spell that out in the name.

`filter/src/builtins/http/security/cpex/config.rs` => `filter/src/builtins/http/security/policy/cpex_config.rs`

filter/src/builtins/http/security/cpex/config.rs
```diff
@@ -0,0 +1,128 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Deserialized YAML configuration for the CPEX security filter.
```
shaneutt:
```suggestion
//! Deserialized YAML configuration for the policy security filter.
```

filter/src/builtins/http/security/cpex/README.md
```diff
@@ -0,0 +1,225 @@
+# cpex filter
```
shaneutt: I appreciate the effort to create a README for this. I'm not necessarily opposed to keeping it, however, the pattern we've been mostly following so far for this sort of thing is rustdoc and our examples. Can we express the supportive information provided here in the rustdoc and examples instead? These will get picked up by the [website](https://praxis.fast).

filter/src/builtins/http/security/mod.rs
```diff
@@ -2,9 +2,12 @@
 // Copyright (c) 2024 Praxis Contributors
 
 //! HTTP security filters: CORS, CSRF, IP access control, credential injection,
-//! forwarded-header injection, and guardrails.
+//! forwarded-header injection, guardrails, and the (feature-gated) CPEX policy
+//! filter.
 
 mod cors;
+#[cfg(feature = "cpex")]
```
shaneutt:
```suggestion
#[cfg(feature = "cpex-policy-engine")]
```

deny.toml
shaneutt: Agreed. I talked with @araujof today and it sounds like he's moving to publish the crates soon so we should be able to switch over.

filter/src/builtins/http/security/cpex/tests.rs
```diff
@@ -0,0 +1,1460 @@
...
+use super::{config::CpexFilterConfig, filter::CpexFilter};
+use crate::{
+    FilterAction,
+    filter::HttpFilter as _,
+    test_utils::{make_filter_context, make_request},
+};
+
+// =====================================================================
+// Fixtures
+// =====================================================================
```
shaneutt: Note: everywhere else uses `// ----`

shaneutt (approved):
I agree with @usize's comments here, but after having a 1:1 conversation with @araujof I think some of those concerns can be mitigated: Context Forge is exploring adopting Praxis to build their proxy server. Fred has suggested that this means merging CPEX into Praxis to become its native policy engine (which then Context Forge still consumes through Praxis) is a possibility which we're going to explore. And ultimately, we need a [Policy Engine](#678) and this seems to align well.

I have several comments, most of which are on the smaller side.

Once the following are complete, I'm down to merge:

- Rename "CPEX" in file names, docs, and anywhere where what we're really referring to is the policy engine for Praxis. Continue to use CPEX nomenclature when the framework itself is the most relevant focus
- Keep all of this behind a `cpex-policy-engine` feature which starts _off_ by default
- Resolve all open comments from team members as well as @praxis-bot
- (Optional): Migrate the `README.md` into rustdocs and examples instead of a nested README.md in the filters

Once those are done, I say we merge this flagged as experimental and continue the experiment, in particular doing spikes to see how the CPEX engine can support [Policy Engine](#678) requirements.

Thank you @araujof! 🖖

filter/src/builtins/http/security/policy/config.rs
araujof: Renamed docs and object names to generalize as a policy engine, so leaving the file name agnostic for now.

Cargo.toml
araujof: Resolved. Now depends on the published crate from crates.io; the git-rev pin is gone.

deny.toml
araujof: Removed; no longer needed now that the crates are published in crates.io.

filter/src/builtins/http/security/cpex/filter.rs
araujof: Acknowledged. CPEX 0.2.0 still offers no sync init path, so the OS-thread + block_on stays. Since CPEX is staying in-tree as Praxis's native policy engine (experimental, off by default), I opened an issue to track the addition of a sync initialization interface to CPEX: https://github.com/contextforge-org/cpex/issues/96

filter/src/builtins/http/security/cpex/filter.rs
araujof: Fixed. Identity resolved in the request phase is stashed in `ctx.extensions` and reused in the response phase via `apply_to_extensions`, resulting in no re-validation, so a token expiring between phases can no longer cause a false deny.

filter/src/registry.rs
araujof: Staying in-tree as an experimental filter as per discussion with @shaneutt; noted the `export_filters!` path if that decision changes.

araujof:
> ## PR Review (third pass)
> **Summary**: Third-pass review of the CPEX policy filter after two prior praxis-bot reviews (9 findings in pass 1, all addressed; 5 findings in pass 2 still open) plus maintainer architectural feedback. This pass found 2 new medium findings not covered by prior reviews.
> 
> Severity	Count
> Medium	2
> All other potential findings overlap with the 5 open items from the second-pass review or the maintainer's architectural comments. Those are not repeated here.

All addressed. Reuse the already-cloned `body_bytes` on the mutate/deny paths, `json_rpc_id` now delegates to `json_rpc_id_value` (single parse), the module-level `#![allow]` is replaced with per-function `#[expect]`, the Bearer token is built in a Zeroizing buffer, and `PolicyFilterConfig` fields are alphabetized.

araujof:
> I agree with @usize's comments here, but after having a 1:1 conversation with @araujof I think some of those concerns can be mitigated: Context Forge is exploring adopting Praxis to build their proxy server. Fred has suggested that this means merging CPEX into Praxis to become its native policy engine (which then Context Forge still consumes through Praxis) is a possibility which we're going to explore. And ultimately, we need a [Policy Engine](#678) and this seems to align well.
> 
> I have several comments, most of which are on the smaller side.
> 
> Once the following are complete, I'm down to merge:
> 
> * Rename "CPEX" in file names, docs, and anywhere where what we're really referring to is the policy engine for Praxis. Continue to use CPEX nomenclature when the framework itself is the most relevant focus
> * Keep all of this behind a `cpex-policy-engine` feature which starts _off_ by default
> * Resolve all open comments from team members as well as @praxis-bot
> * (Optional): Migrate the `README.md` into rustdocs and examples instead of a nested README.md in the filters
> 
> Once those are done, I say we merge this flagged as experimental and continue the experiment, in particular doing spikes to see how the CPEX engine can support [Policy Engine](#678) requirements.
> 
> Thank you @araujof! 🖖

Thanks @shaneutt! Done: renamed CPEX→policy across the filter name, module/files (cmf.rs→common_message_format.rs, etc.), docs, and examples; gated behind a new `cpex-policy-engine` feature that's off by default; added the experimental warning; migrated the nested README into module rustdoc + the example; and resolved the open team/bot comments. CPEX nomenclature is kept where the engine itself is the focus. Also moved to the published cpex 0.2.0 and renamed the `X-Cpex-Violation` header to `X-Policy-Violation`. Tested the end-to-end demo in praxis-demos. Ready for the experimental merge.

shaneutt (approved):

filter/src/builtins/http/security/cpex/filter.rs
araujof: Acknowledged and documented in the module rustdoc. CPEX 0.2.0 still doesn't offer a sync hook-dispatch API. Tracking this in the same initialize issue https://github.com/contextforge-org/cpex/issues/96

PR #643, opened 2026-06-21
docs(proposals): add How? section to #210 response-based token counting

mkoushni:
# docs(proposals): add How? section to praxis-proxy/ai#83 response-based token counting

## Summary

Completes the graduation criteria for proposal [#210](https://github.com/praxis-proxy/ai/issues/83) by adding the **How?** section — the design and implementation plan for the `token_count` filter that extracts token usage from AI provider response bodies and headers.

---

## What changed

`docs/proposals/00210_response-based-token-counting.md` — 193 lines added, status updated from `proposed` → `accepted`.

### Open questions resolved

All four open questions from the What?/Why? section are answered in the new How? section:

| Question | Decision |
|---|---|
| Provider identification | Explicit `provider:` YAML key — no auto-detection. Azure and OpenAI share the same JSON schema so auto-detection would be ambiguous. |
| Streaming completion signal | `BodyMode::StreamBuffer` — proxy buffers all response body bytes and delivers them once with `end_of_stream: true`. Stream close is the authoritative trigger, covering providers that omit `[DONE]` (Google Gemini). |
| Streaming accumulation per provider | Per-provider strategy: single terminal-chunk scan for OpenAI/Azure/Google/Bedrock Converse; two-event scan (`message_start` + `message_delta`) for Anthropic; header-only for Bedrock InvokeModel. |
| Partial usage data | Only the final assembled payload is parsed — no summing of intermediate chunks — to avoid double-counting. |

### Design content added

- **Requirements** — 7 concrete implementation requirements.
- **Filter struct and config** — `TokenCountConfig`, `ProviderKind` enum, `TokenCountFilter` struct with YAML snippet.
- **`HttpFilter` hook table** — behaviour of each hook (`on_request`, `on_response`, `response_body_access`, `response_body_mode`, `on_response_body`).
- **SSE extraction detail** — dispatch tree: Anthropic two-event scan vs. last-valid-chunk scan for all other providers.
- **Bedrock InvokeModel path** — header-only extraction in `on_response`; `BodyAccess::None` prevents unnecessary buffering.
- **`FilterContext` metadata keys** — `token.input`, `token.output`, `token.total` written via `ctx.set_token_usage`.
- **Module registration** — step-by-step wiring for `ai/mod.rs`, `http/mod.rs`, `builtins/mod.rs`, and `registry.rs`.
- **YAML configuration example** — minimal working filter chain showing `token_count` with `provider: openai`.

---

## Related

- Issue: [#210 Response-based token counting from provider JSON](https://github.com/praxis-proxy/ai/issues/83)
- Token counting epic: [#20](https://github.com/praxis-proxy/ai/issues/71)
- Provider mapping library: [#216](https://github.com/praxis-proxy/ai/issues/87)
- FilterContext metadata: [#212](https://github.com/praxis-proxy/ai/issues/221)
- Token header injection filter: [#214](https://github.com/praxis-proxy/ai/issues/86)
- Integration tests proposal: [#220](https://github.com/praxis-proxy/ai/issues/88)


docs/proposals/00210_response-based-token-counting.md
```diff
@@ -152,3 +152,205 @@ chunks as well as the final chunk. Should the filter accumulate and
...
+
+## How?
+
+### Requirements
+
+1. A new `token_count` filter in `filter/src/builtins/http/ai/token_count.rs`.
+2. The filter accepts a single required YAML key `provider` that selects
+   the extraction strategy.
+3. For non-streaming responses the full body is buffered and parsed once
+   at end-of-stream.
```
shaneutt, line 164: This is pretty random, but I'm curious: has anyone considered an optional alternative to full token counting where you estimate tokens based on the average tokens for any given byte total, and then apply that average to the payload size for an "estimate"?

I'm curious if some users would opt for an estimate to avoid the additional buffering and payload processing here (estimate instead of count, and then stream through without any memory buffering).

🤔

docs/proposals/00210_response-based-token-counting.md
```diff
@@ -152,3 +152,205 @@ chunks as well as the final chunk. Should the filter accumulate and
...
+  - name: gateway
+    address: "127.0.0.1:8080"
+    filter_chains:
+      - token-counting
+
+filter_chains:
+  - name: token-counting
+    filters:
+      - filter: token_count
+        provider: openai   # openai | anthropic | google | bedrock | bedrock_invoke_model | azure
```
shaneutt, line 339: We would link in count for llm-d, or raw vllm here?

docs/proposals/00210_response-based-token-counting.md
mkoushni: First, we need to create at least a database and know how many tokens we can expect. We can simply run it this way, collect statistics, and then analyze them.
Optionally, for some models, we can set an average number without wasting time calculating each time.

docs/proposals/00210_response-based-token-counting.md
mkoushni: The difference is operational, not in the filter config:

Raw vLLM — point the ai-provider cluster endpoint directly at the vLLM server (e.g. 127.0.0.1:8000). The token_count filter reads the usage fields from the JSON response exactly as shown.

llm-d — same thing, just point the endpoint at the llm-d gateway instead. llm-d wraps vLLM and preserves the OpenAI response schema, so provider: openai still applies.

The only case where you'd need a different provider value is if you're calling Bedrock or Anthropic directly rather than through an OpenAI-compatible shim. For both vLLM and llm-d the example YAML is correct as-is.

But for the implementation part, I will do it according to the llm-d.

shaneutt (approved):

PR #644, opened 2026-06-22
Proposal/http callout filter

usize:
Adds a how section to the http call out filter proposal. 

This proposal splits the work into two deliverables:

  1. A new crate supporting http callouts from any filter.
  2. A general purpose http call out filter, doubling as reference implementation. 
  
  praxis-proxy/ai#76 is called out as a target consumer of the new http call out library.


docs/proposals/00358_http-callout-filter.md
```diff
@@ -197,9 +196,350 @@ filter_chains:
...
+### Requirements
+
+1. **HTTP client pool** — per-instance, connection-pooled
+   HTTP client independent of Pingora's upstream pool.
+   Never shared across filter instances to prevent
+   credential and connection leakage in multi-tenant
+   deployments.
+2. **SSRF prevention** — callout targets fully declared
+   in config. The filter never constructs a URL from
+   request data.
```
shaneutt, line 223: 👍

docs/proposals/00358_http-callout-filter.md
```diff
@@ -197,9 +196,350 @@ filter_chains:
...
+### Design
+
+#### Crate layout
+
+```text
+server
+  ├── praxis-http-callout (filter, opt-in feature flag)
+  │     ├── praxis-callout (shared client)
+  │     ├── praxis-filter (HttpFilter, FilterResultSet)
+  │     └── serde_json_path
```
shaneutt, line 265: For other builds of Praxis using the framework, I expect we'll want to provide this facility. So we should move this into `core/` to make sure it's available for custom proxies?

docs/proposals/00358_http-callout-filter.md
```diff
@@ -197,9 +196,350 @@ filter_chains:
...
+  failure_threshold: 5
+  recovery_timeout: 30s
+```
+
+### Implementation
+
+Proposed PR sequence:
+
+- **PR 1** — `praxis-callout` crate (`callout/`) with
+  the shared HTTP callout client: `reqwest` client pool,
```
shaneutt: In this doc we should write out an explicit justification for our use of `reqwest` for this purpose, specifically. Doesn't have to be long, but just something that makes it clear why this is the best choice since it's setting a foundation and we may carry this dependency a long time.

docs/proposals/00358_http-callout-filter.md
```diff
@@ -197,9 +196,350 @@ filter_chains:
...
+  (`filter/http-callout/`) with `HttpCalloutFilter`
+  struct, config parsing, SSRF validation, JSONPath
+  extraction, body forwarding, and unit tests.
+  Registered in the server binary behind a feature flag.
+
+- **PR 3** — integration test and example config:
+  Lakera Guard example in `examples/configs/ai/`,
+  functional integration test in
+  `tests/integration/tests/suite/examples/` using a
+  mock HTTP server that returns Lakera-shaped responses.
```
shaneutt: Optional: seems like you could just merge 2 and 3?

shaneutt (changes requested):

docs/proposals/00358_http-callout-filter.md
usize: Covered at line 298 now. tl;dr it's the most popular HTTP client in the rust ecosystem.

docs/proposals/00358_http-callout-filter.md
usize: Done. I'm going to change the proposal to suggest a feature flag since reqwest is heavy.

shaneutt (approved):

PR #649, opened 2026-06-22
refactor(core): Removed `Arc` from sigs where not required

alexsnaps:
Moved the `Arc::clone` on the call site, as it isn't "optional", `spawn_config_watcher` needs to hold on to the `Arc`, from the `spawn_watcher` call.
otoh, `register_protocols` doesn't need an `Arc` at all, "only" requires to borrow the ref to `ListenerPipelines`.

Makes the code easier to read, as to what the contract on each of those calls are (without needing to dig, potentially deep, into the call tree) wrt ownership & lifetimes. I think that's generally desirable. 

There are a few other places where this could be aligned, if this makes sense to all. wdyt @shaneutt @twghu

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -102,7 +102,8 @@ pub fn run_server_with_registry(config: Config, registry: FilterRegistry, config
 
     let health_shutdown = Arc::new(Mutex::new(CancellationToken::new()));
     spawn_health_check_tasks(&config, &health_registry, &health_shutdown);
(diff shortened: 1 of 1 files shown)


alexsnaps:
Darn... Sorry, looks like I have a stale `main` 🤦 
Question remains relevant, but not this PR, let me close and see if it even matters still, looks like all is in `ServerState` now. Sorry about this again...

PR #651, opened 2026-06-22
refactor(store): No need to clone the `Arc`, deref store right away

alexsnaps:
Real minor change, but there is no need to `Arc::clone` the `ResponseStore`, `deref`ing and using the reference right away is sufficient as we do all this while `ResponseStoreFilter` being deref'ed itself, or am I missing some (possibly future?) intent here?

/cc @leseb


leseb (approved):
good catch, this wasn't intentional - thanks!

PR #677, opened 2026-06-22
Add Responses SSE framing coverage

franciscojavierarceo:
Part of praxis-proxy/ai#85.

## Summary
- add data-only SSE frame serialization for translated Responses stream events
- cover event payload round-tripping and response.completed termination without a [DONE] sentinel

## Stack
- stacked on praxis-proxy/praxis#696

## Validation
- cargo test -p praxis-proxy-filter --lib --features ai-inference openai::translation
- make lint


shaneutt:
Is there a place where we're currently describing in some detail the the relationship between OGX and Praxis today?

franciscojavierarceo:
> Is there a place where we're currently describing in some detail the the relationship between OGX and Praxis today?

Codex got ahead of itself here. The only description here is the reuse of the test coverage from OGX, which is a good thing and was explicitly requested.

PR #700, opened 2026-06-23
feat(ai_guardrails):add NeMo provider for request-side evaluation

christinaexyou:
Implements the NeMo provider for the ai_guardrails filter, completing request-side evaluation against NeMo Guardrails. Address praxis-proxy/ai#48.

The following changes were made:
`providers/nemo.rs`
* Implements NemoProvider which sends a POST to the configured /v1/guardrail/checks endpoint and maps the NeMo response to a GuardResult

`filter.rs`
* on_request_body which checks phase.request (default true), waits for end_of_stream, parses the body, calls extract_messages, invokes the provider, and acts on the verdict:
    > Pass → Continue
    > Block → Reject 403 with the blocked rail name(s) in the body
    > Redact → Continue (body replacement placeholder for praxis-proxy/ai#49)
* extract_messages: parses OpenAI Chat Completions API formats, returns FilterError for unsupported formats

## Testing
* Unit tests (`filter.rs`, `tests.rs`, `providers/nemo.rs`)
* Integration tests (`tests/integration/tests/suite/ai_guardrails.rs`) using a mock NeMo Guardrails server

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1174,8 +1174,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff2abc00be7fca6ebc474524697ae276ad847ad0a6b3faa4bcb027e9a4614ad0"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -1185,9 +1187,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "r-efi 5.3.0",
  "wasip2",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


liavweiss:
Hey @christinaexyou, thank you for the work here, much appreciated!
I focused my review on the logic and correctness rather than code style, and I have a few things I need clarification on:

1. **`model` field in NeMo request** - It looks like the `NemoRequest` only sends `messages` to NeMo, but NeMo expects a `model` field as well (it can also be `model: ""`). Am I wrong?

2. **MCP format support** - The ticket mentions extracting messages from both OpenAI and MCP body formats, but `extract_messages()` only handles OpenAI Chat. We don't need to support this here, right?

3. **Strict error on unrecognized body format** - Currently if a request body doesn't have a `"messages"` field, `extract_messages()` returns an error. Should we allow those requests to pass through instead (in case non-chat requests like model listing hit this filter)?

4. **NeMo response size limit** - Minor, but should we cap the response body we read from NeMo (e.g. 1 MiB) to avoid unbounded memory usage if the provider returns something unexpected?

christinaexyou:
@liavweiss thanks for the review ! i updated my PR to address your comments and fix the code style issues. i made the following changes:

* Added the `model` field to the outgoing NeMo request. It's optional string and set to "" if empty
* I deleted the stray comment regarding MCP format support 
* I prefer having a strict error in if the request body is missing the messages field - if there's a use case for model listing that needs to be filtered through NeMoProvider, I can add a schema + function to send request to `v1/models` endpoint
* I added input validation to ensure the response body is under the size limit

liavweiss:
Hey @christinaexyou, tnx for the changes, it looks good!

One last note on the `model` field - I know NeMo doesn't actually use it for guardrail evaluation (the guard config is defined in NeMo's `config.yml`), but it's required by NeMo's OpenAI-compatible API schema. Right now, it's always empty and skipped via `skip_serializing_if`, so it's never sent. It would be better to extract it from the request body and forward it to keep consistency with NeMo's expected payload format.

You can use as a reference the way we do it in the IPP request plugin: https://github.com/opendatahub-io/ai-gateway-payload-processing/blob/f0ddb042126ed581be692edaa86af4e334968a96/pkg/plugins/nemo/request_guard.go#L101

What do you think?

shaneutt:
Please see https://github.com/orgs/praxis-proxy/discussions/752 and note that we're migrating the AI features from this repository to https://github.com/praxis-proxy/ai. You'll need to re-target this PR on that repository since it is the new home of the relevant capabilities. Please let us know if you need any assistance! :vulcan_salute:

PR #702, opened 2026-06-23
Add Responses request translation core

franciscojavierarceo:
Part of praxis-proxy/ai#85.

## Summary
- add the internal Responses-to-Chat-Completions request translator
- cover instructions, message input, text/image content parts, function tools, tool choice, and max_output_tokens mapping
- keep the helper internal for later HTTP filter wiring

## Stack
- depends on praxis-proxy/praxis#701
- base of praxis-proxy/praxis#703
- followed by praxis-proxy/praxis#696 and praxis-proxy/praxis#677

## Validation
- cargo test -p praxis-proxy-filter --lib --features ai-inference translation
- cargo test -p praxis-proxy-filter --lib --features ai-inference openai::responses
- make lint

PR #701, opened 2026-06-23
Share Responses body size limits

franciscojavierarceo:
Part of praxis-proxy/ai#85.

## Summary
- add a shared Responses body-size alias for filters that inspect or transform Responses request bodies
- switch Responses validation to the shared 64 MiB Responses limit
- keep the format classifier on the generic JSON inspection default

## Stack
- base of praxis-proxy/praxis#702
- followed by praxis-proxy/praxis#703, praxis-proxy/praxis#696, and praxis-proxy/praxis#677

## Validation
- cargo test -p praxis-proxy-filter --lib --features ai-inference openai::responses
- cargo test -p praxis-tests-integration openai_responses_to_chat_completions
- make lint

PR #696, opened 2026-06-23
Add Responses to Chat Completions filter

franciscojavierarceo:
Part of praxis-proxy/ai#85.

## Summary
- add the openai_responses_to_chat_completions HTTP filter and config
- gate transformation to POST /v1/responses while leaving other Responses endpoints alone
- add docs, example config, registry wiring, and integration coverage

## Stack
- stacked on praxis-proxy/praxis#703
- base for praxis-proxy/praxis#677

## Validation
- cargo test -p praxis-proxy-filter --lib --features ai-inference openai::translation
- cargo test -p praxis-proxy-filter --lib --features ai-inference openai::responses::to_chat_completions
- cargo test -p praxis-tests-integration openai_responses_to_chat_completions
- make lint

PR #703, opened 2026-06-23
Add Responses translation semantics

franciscojavierarceo:
Part of praxis-proxy/ai#85.

## Summary
- extend the translator with schema-complete Responses resource and stream-event conversion
- preserve function-call history, structured output formats, allowed tool choices, strict function schemas, logprobs settings, and content_filter incomplete status
- add curated Chat Completions recording fixtures; the fixture provenance comes from OGX compatibility testing

## Stack
- built on current main after praxis-proxy/praxis#701 and praxis-proxy/praxis#702
- base for praxis-proxy/praxis#696
- praxis-proxy/praxis#677 is stacked on praxis-proxy/praxis#696

## Validation
- cargo test -p praxis-proxy-filter --lib --features ai-inference openai::translation
- make lint


filter/src/builtins/http/ai/openai/translation/chat_completions.rs
```diff
@@ -520,3 +718,1212 @@ fn json_type_name(value: &Value) -> &'static str {
...
+        "store": context.store,
+        "temperature": number_or_default(context.temperature.as_ref(), 1.0),
+        "text": context.text,
+        "tool_choice": tool_choice_value(context),
+        "tools": context.tools,
+        "top_p": number_or_default(context.top_p.as_ref(), 1.0),
+        "truncation": DEFAULT_TRUNCATION,
+        "usage": parts.usage,
+        "metadata": metadata_value(context),
+        "background": false,
```
leseb, line 779: i know we don't support background yet, so is it intentional to always mark it false? if so, we should have a TODO to change this, or you can handle this now in ResponseContext - up to you. feels like a small change here

filter/src/builtins/http/ai/openai/translation/chat_completions.rs
```diff
@@ -520,3 +718,1212 @@ fn json_type_name(value: &Value) -> &'static str {
...
+        "parallel_tool_calls": context.parallel_tool_calls,
+        "previous_response_id": previous_response_id_value(context),
+        "reasoning": Value::Null,
+        "store": context.store,
+        "temperature": number_or_default(context.temperature.as_ref(), 1.0),
+        "text": context.text,
+        "tool_choice": tool_choice_value(context),
+        "tools": context.tools,
+        "top_p": number_or_default(context.top_p.as_ref(), 1.0),
+        "truncation": DEFAULT_TRUNCATION,
```
leseb, line 774: ditto - i know this one is complex but can we leave a comment in the code with some reasoning?

filter/src/builtins/http/ai/openai/translation/chat_completions.rs
```diff
@@ -352,70 +539,81 @@ impl Default for ConvertedContentParts {
...
+    Ok(json!({
         "type": "image_url",
         "image_url": Value::Object(image_url)
     }))
 }
 
 /// Convert a `Responses` file content part into Chat Completions shape.
-fn convert_input_file_part(part: &Value) -> Option<Value> {
-    let obj = part.as_object()?;
+fn convert_input_file_part(part: &Value) -> Result<Value, TranslationError> {
```
leseb, line 544: this function ignores file_url - is it on purpose because we need to fetch it etc? we can do it in a followup

leseb (changes requested):
the approach to the streaming implementation is incorrect, the entire response is buffered then SSEs are emitted, it's more a batch than a streaming at this point. can you remove the streaming bit from this PR? let's redesign it as a followup. we need to parse each SSE frame, translate it to Responses event, etc etc. or perhaps this was intentional? if so why?

thanks!

filter/src/builtins/http/ai/openai/translation/chat_completions.rs
franciscojavierarceo: fixed

filter/src/builtins/http/ai/openai/translation/chat_completions.rs
franciscojavierarceo: yeah

filter/src/builtins/http/ai/openai/translation/chat_completions.rs
franciscojavierarceo: left TODO

franciscojavierarceo:
> the approach to the streaming implementation is incorrect, the entire response is buffered then SSEs are emitted, it's more a batch than a streaming at this point. can you remove the streaming bit from this PR? let's redesign it as a followup. we need to parse each SSE frame, translate it to Responses event, etc etc. or perhaps this was intentional? if so why?
> 
> thanks!

stripped out

leseb (approved):

PR #705, opened 2026-06-24
feat(filter): support wildcard Responses model aliases

nerdalert:
## Summary

Follow-up to the merged `/v1/responses` passthrough work. This lets openai_responses_model_rewrite alias sources use a single wildcard, so operators can map a model family like `codex-*` to a backend deployment without updating config every time the client-facing model name changes.

The filter now:

- Keeps exact model aliases working as before.
- Allows single-wildcard alias keys such as `codex-*` or `gpt-4.1-*`
- Resolves exact aliases before wildcard aliases.
- Resolves wildcard aliases by literal specificity, with deterministic lexical tie-breaking.
- Rejects alias source patterns with more than one wildcard.

## Why

The original filter used exact HashMap keys only. That works for stable names, but it is fragile for client-controlled or provider-controlled model names. For example, if Codex sends `codex-mini-2026-06-24` instead of `codex-mini-latest`, an exact alias would miss and the request would reach the backend with an unknown model name.

With this change, operators can configure:

```yaml
model_aliases:
  "codex-*": "llama-3.3-70b"
```

That keeps the rewrite policy stable across client model-name churn while preserving exact aliases for cases that need a narrower override.

## Validation

- cargo test -p praxis-proxy-filter --features ai-inference model_rewrite::tests
- cargo test -p praxis-tests-integration openai_responses_model_rewrite
- make lint

### DIFF
--- a/docs/filters/http/ai/openai_responses_model_rewrite.md
+++ b/docs/filters/http/ai/openai_responses_model_rewrite.md
@@ -7,6 +7,10 @@ Rewrites the `model` field in Responses API request bodies.
 
 Requires Cargo feature: `ai-inference`.
 
+## Configuration Notes
+
+Quote wildcard alias keys in YAML, such as `"gpt-4.1-*"`, so `*` is parsed as a literal character rather than YAML alias syntax. The examples quote all alias keys for consistency.
+
 ## Configuration
 
 | Field | Type | Required | Description |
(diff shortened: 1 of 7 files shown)


docs/filters/http/ai/openai_responses_model_rewrite.md
```diff
@@ -16,7 +16,7 @@ Requires Cargo feature: `ai-inference`.
 | `headers.effective_model` | string | no | Header name for the effective (post-rewrite) model value. |
 | `headers.original_model` | string | no | Header name for the original (pre-rewrite) model value. |
 | `max_body_bytes` | usize | no | Maximum request body size to buffer before parsing. |
-| `model_aliases` | object<string, string> | no | Map from client-facing model names to backend model names. |
+| `model_aliases` | object<string, string> | no | Map from client-facing model names or single-wildcard patterns to backend model names. Exact aliases win before wildcard aliases; wildcard aliases are matched by literal specificity. |
```
leseb: can add a callout in the docs explaining that wildcard keys need quotes because of yaml syntax? or/and quote all keys always?

filter/src/builtins/http/ai/openai/responses/model_rewrite/tests.rs
```diff
@@ -115,6 +132,22 @@ model_aliases:
...
+        r#"
+model_aliases:
+  "gpt-*-mini-*": "qwen-2.5-72b"
+"#,
+    )
+    .unwrap();
+    let result = ModelRewriteFilter::from_config(&yaml);
+    assert!(
+        result.is_err(),
+        "alias source with more than one wildcard should be rejected"
```
leseb, line 147: 👍🏻

leseb (commented):
nit

docs/filters/http/ai/openai_responses_model_rewrite.md
nerdalert: Done ty!

leseb (approved):
Thanks for the followup!

PR #709, opened 2026-06-24
fix(filter): use retryable Postgres init on GET/DELETE store paths

leseb:
## Summary

- The response store's GET and DELETE handlers used `get_or_init` with `init_permanent_store`, which writes `None` into the `OnceCell` on any build failure — permanently poisoning the store for POST persistence too
- Switched `handle_delete` and `ensure_store` (GET handlers) to delegate through `get_or_init_store()`, which uses `get_or_try_init` for Postgres so transient failures are retried on subsequent requests
- Unified all read/write/delete paths to go through the same retry-aware initialization

## Test plan

- [x] `cargo test -p praxis-proxy-filter -- response_store` — all 3 tests pass
- [x] `make lint` — clean
- [x] `make build` — clean


leseb:
> **Purpose**: Fix Postgres store init poisoning on GET/DELETE paths by routing all code paths through the retry-aware `get_or_init_store()` method.
> 
> **Assessment**: Clean, minimal, correct bugfix. All three hunks are straightforward delegations that eliminate duplicated `get_or_init` + `init_permanent_store` calls that permanently cached `None` for Postgres failures. The fix unifies GET, DELETE, and POST paths through the single retry-aware entry point.
> 
> No inline findings -- the code changes are correct as written.
> 
> Severity	Count
> Medium	1
> **Non-inline findings:**
> 
> **[Medium]** The existing test `postgres_store_init_failure_is_not_cached` only exercises the `on_request` (POST) path -- it does not cover the two paths this PR actually fixes. Add tests that call `handle_delete` and `ensure_store` (via `handle_get_response`) against a Postgres-configured filter with an unreachable host, and assert that `filter.store.get().is_none()` after the failed init, matching the pattern of the existing test. Without these, the exact regression that motivated this PR remains untested.

https://github.com/praxis-proxy/praxis/pull/709/commits/4442f3dd6f76ad842d100c628d4dd5444a2718d9

franciscojavierarceo (approved):

Issue #110, opened 2026-05-01
Slow Start for Upstream Endpoints


shaneutt:
Gradually ramp traffic to newly healthy or newly added upstream endpoints instead of immediately sending full load.

Scope:
- Configurable slow start window (duration over which weight ramps from 0 to configured weight)
- Linear or aggression-factor ramp curve
- Triggered on: endpoint recovery from health check failure, newly added endpoint
- Per-cluster configuration
- Integration with weighted load balancing (effective weight increases over the window)

PR #746, opened 2026-07-01
feat(policy): AuthPolicy support (transpiler + generic-HTTP authorization)

### LINKED ISSUE #716: Spike: Supporting AuthPolicy
Kuadrant provides an `AuthPolicy` API which is highly customizable and is what higher-level policies build upon. We need to look into supporting that here. For an initial POC, I recommend making a simple CLI that ingests `AuthPolicy` resources and converts them to Praxis configuration to demo.

araujof:
Supports Kuadrant `AuthPolicy` in Praxis using the CPEX policy engine as the PDP (CEL). 

Closes #716 

Depends on the CPEX engine additions in **contextforge-org/cpex#110**.

> Draft. Consumes CPEX via a **temporary path dep** (`Cargo.toml`, marked TEMPORARY); swap back to a published `cpex` version once #110 releases before merge.

## Offline transpiler (`cargo xtask transpile-authpolicy`)

Translates a Kuadrant `AuthPolicy` (`kuadrant.io/v1` / Authorino `v1beta3`) into a Praxis `policy`-filter block + a CPEX policy document, best-effort, with a coverage report. Fragment-only; no cluster/Gateway-API access.

- **Parser** (`xtask/src/authpolicy/model.rs`) — supported subset, tolerant of unknown constructs.
- **CEL namespace remap** (`cel.rs`) — `auth.identity.*`→`claim.*`, `request.headers[x]`→`http.request_headers.x`, `request.method/path/host`→`http.*`. Un-remappable refs become reported gaps, never emitted (wrong-namespace CEL would fail closed at runtime).
- **Fail-closed** — an `AuthPolicy` that declares authz but loses every rule to gaps emits `require(false)` + a FATAL report + non-zero exit (never fail-open).
- **Security posture** — explicit JWT algorithm pinning, `denyWith` CR/LF rejection, header/nested-claim warnings.
- **Golden tests** over a 3-example corpus, incl. a round-trip through the real cpex-core parser.

## Runtime enforcement + live demo

Experimental `enforcement: http` mode on the `policy` filter (behind the `experimental-http-authz` feature, off by default; MCP path unchanged):

- Populates the HTTP request line + headers into the CPEX bag (`host` from the validated authority, not a raw `Host`), evaluates the CPEX `global` policy via `cmf.http_request`, and maps denials to plain HTTP responses carrying the custom `denyWith` (status/body/headers).
- **End-to-end demo** (`examples/configs/security/authpolicy-http.yaml` + integration test): GET+JWT → 200, POST+JWT → **403 with custom denyWith body/header**, no-JWT → 401.

## Notes / pre-merge

- **Transpiler emission refinement (follow-up):** APL `require(...)` is truthiness-only; comparisons need the predicate:action form and CEL functions need a `cel:` PDP step. The current `require(<cel>)` emission must be refined before transpiled policies compile as APL end-to-end.
- Keycloak nested-claim mapping (CPEX `identity/jwt`) deferred; the demo uses method-based authz.

Run the demo:
```
cargo test -p praxis-tests-integration --features experimental-http-authz --test suite authpolicy_http
```


shaneutt (changes requested):
Good work @araujof, I appreciate the effort!

Right now though, we're not actually looking to adopt `AuthPolicy` just yet, the spike was to ascertain the gaps (if any). How do you feel about converting this into a demo in the demos repository, and then sharing it in Slack so we can see if we can get some interest and engagement from Kuadrant folks? Then create issues for any gaps that you found so we can start tracking those as sub-tasks of the policy engine epic?

shaneutt:
Checking in, need any assistance?

shaneutt:
Gonna close for now Fred and I'll catch up with you about getting this into our demos :vulcan_salute:

araujof:
> Good work @araujof, I appreciate the effort!
> 
> Right now though, we're not actually looking to adopt `AuthPolicy` just yet, the spike was to ascertain the gaps (if any). How do you feel about converting this into a demo in the demos repository, and then sharing it in Slack so we can see if we can get some interest and engagement from Kuadrant folks? Then create issues for any gaps that you found so we can start tracking those as sub-tasks of the policy engine epic?

Hi @shaneutt, apologies for the delay in responding to this. I was traveling and working part time in the past 10 days. 

Indeed, this spike was useful to bridge a feel gaps in the policy engine and better align it to support policies like the ones defined in Kuadrant AuthPolicy. I will convert this work into a standalone CLI for a demo in the Praxis demos as suggested. Thanks!

araujof:
> Checking in, need any assistance?

I missed these messages! In any case, fully agree with moving this as a demo for now.

PR #757, opened 2026-07-02
ci: add unicode safety check for PR context poisoning

aslakknutsen:
## Summary
- Adds a standalone `Unicode Safety Check` workflow that scans PR diffs for hidden Unicode characters used in ASCII smuggling / AI context poisoning attacks
- Uses `dcondrey/unicode-safety-check@v3.0.0` (SHA-pinned), matching [odh-model-controller#823](https://github.com/opendatahub-io/odh-model-controller/pull/823)
- Checkout and runner pins match existing praxis CI (`actions/checkout@v7.0.0`, `ubuntu-24.04`)

## Test plan
- [x] Local scan of this PR diff: 1 file scanned, 0 findings
- [ ] GitHub Actions `Unicode Safety Check` job passes on this PR
- [ ] After merge: optionally enable as required status check in branch protection

### DIFF
--- a/.github/workflows/unicode-safety.yaml
+++ b/.github/workflows/unicode-safety.yaml
@@ -1,0 +1,77 @@
+name: Unicode Safety Check
+
+# ------------------------------------------------------------------------------
+# Workflow Settings
+# ------------------------------------------------------------------------------
+#
+# Scans PR diffs for hidden Unicode (tag chars, zero-width joiners, bidi
+# overrides, PUA) used in ASCII smuggling / AI context poisoning attacks.
+#
+# Uses unicode-safety-check v3.0.0 via its release binary rather than the
+# composite action: praxis-proxy org policy only allows GitHub-verified
+# Marketplace actions (see odh-model-controller#823 for the upstream pattern).
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+permissions: {}
+
+jobs:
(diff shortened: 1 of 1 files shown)


aslakknutsen:
**Unicode Safety Check** CI is green on this PR (run after switching to release-binary invocation due to org Marketplace action policy).

After merge, a repo admin can optionally add \`Unicode Safety Check\` / \`Detect hidden unicode characters\` as a required status check under **Settings → Branches → Branch protection rules** for \`main\`.

Alternative: allowlist \`dcondrey/unicode-safety-check@*\` in **Settings → Actions → General → Allow specified actions** to use the composite action directly (as in opendatahub-io/odh-model-controller#823).

.github/workflows/unicode-safety.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          else
+            git fetch origin main
+            BASE_SHA="$(git merge-base origin/main HEAD)"
+          fi
+
+          TARBALL="unicode-safety-check-x86_64-unknown-linux-gnu.tar.gz"
+          BASE_URL="https://github.com/dcondrey/unicode-safety-check/releases/download/v${SCANNER_VERSION}"
```
shaneutt, line 56: Let's avoid this kind of dependency, as it creates a visibility problem in our supply chain.

shaneutt (changes requested):
I like the concept, but can we implement this without adding a new dependency?

.github/workflows/unicode-safety.yaml
aslakknutsen: The visibility problem is an inline URL fetch? Would a GitHub Action dependency be more visible?

PR #759, opened 2026-07-03
fix(config): enforce loopback-only admin endpoint binding

aslakknutsen:
## Description

Admin endpoint validation only rejected `0.0.0.0` / `[::]` bindings. LAN addresses such as `192.168.1.50:9901` or `10.0.0.5:9901` passed validation while docs and `AdminConfig` claimed loopback-only binding. The unauthenticated admin/KV API (`/api/kv/*`, `/metrics`, `/ready`) could therefore be exposed on routable interfaces without setting `insecure_options.allow_public_admin`.

This PR rejects any non-loopback `admin.address` unless `allow_public_admin` is set, normalizes IPv4-mapped loopback (`[::ffff:127.0.0.1]`) before the check, updates docs, and adds regression tests.

## Steps to Reproduce

1. Configure Praxis with:
   ```yaml
   admin:
     address: "192.168.1.50:9901"
   ```
2. Start the proxy (config validates successfully on v0.4.0).
3. From another host on the same network segment, call `GET http://192.168.1.50:9901/api/kv/{store}` or `PUT` to modify KV data.

## Expected Behavior

Non-loopback admin binds should fail config validation unless `insecure_options.allow_public_admin: true` is explicitly set. Loopback forms (`127.0.0.1`, `[::1]`, IPv4-mapped loopback) should be accepted by default.

## Praxis Version

v0.4.0 (also affects main at `8cd1059`)

## Area

Core / Config

## Summary

- Gate `admin.address` on `normalize_mapped_ipv4(ip).is_loopback()` instead of only blocking unspecified addresses
- Document `allow_public_admin` override on `AdminConfig` and in operating docs
- Add tests for LAN rejection, IPv4-mapped loopback acceptance, and LAN override with `allow_public_admin`

## Test plan

- [x] `cargo test -p praxis-proxy-core admin`
- [x] `cargo test -p praxis-proxy-core allow_public_admin`
- [x] `cargo test -p praxis-tests-schema admin`


shaneutt (approved):
Cool 👍

PR #760, opened 2026-07-04
fix(pipeline): preserve StreamBuffer header mutations for broker routing

nerdalert:
## Summary

  Fixes core pipeline plumbing needed by body-inspecting broker filters.

  This PR has two related fixes:

 1. Pipeline validation now understands that a broker-configured `mcp` filter can select an upstream cluster, similar to `router`. In this context, “broker-configured” means the `mcp` filter has a `servers:` catalog that lets it answer MCP discovery requests and route MCP tool calls to configured backend clusters.
  2. `StreamBuffer` request-body pre-read now preserves all request header mutations made during body-phase filtering.

  This does not add new MCP behavior by itself. It fixes the core pipeline path so follow-up broker filters can safely inspect a request body, choose a backend, mutate forwarding headers, and then continue through the normal upstream request flow.

  ## What Was Broken

  Before this PR, body-phase filters could set request header mutations during `StreamBuffer` pre-read, but not all mutation types survived into upstream forwarding.

  Specifically:

  - `extra_request_headers` survived
  - `request_headers_to_remove` did not
  - `request_headers_to_set` did not

  That meant a body-inspecting filter could correctly decide what needed to change, but the actual upstream request could still miss the header set/remove changes.

  Pipeline validation also only treated `router` as a cluster selector. That meant a broker filter configured with explicit backend servers could select a cluster at runtime, but validation still viewed it like a non-routing filter.

  ## What This Fixes

  This PR updates `StreamBuffer` pre-read to carry all request header mutation types forward:

  - promoted extra headers
  - headers to remove
  - headers to set or overwrite

  Those mutations are applied to both:

  - the Pingora session request headers
  - the local request snapshot used by the remaining pipeline

  The mutations are applied before header-phase processing continues, so later filters and upstream forwarding see the same request state.

  This PR also updates pipeline validation so:

  - `mcp` with `servers:` counts as a cluster-selecting broker filter
  - classifier-only `mcp` does not count as a cluster selector
  - malformed `servers` does not count as broker mode
  - `servers[].cluster` participates in cluster cross-reference validation
  - configs with both `router` and broker-configured `mcp` before the same `load_balancer` are rejected instead of silently allowing one cluster selection to overwrite the other

  ## Why This Matters

  This is a prerequisite fix for the AI MCP stateless `tools/call` routing work.

  That follow-up PR needs the MCP broker to:

  - read the JSON-RPC body
  - identify the requested tool
  - select the configured backend cluster
  - rewrite the upstream tool name/header
  - forward the request through the normal load-balancer path

  Without this core fix, the broker can make the right decision during body processing but lose header mutations before the backend request is sent.

## Design Context

  MCP static broker mode has its own protocol-level catalog. The configured `servers:` list defines the backend MCP servers the broker exposes, the Praxis cluster each server uses, and the tools that appear in `tools/list`.

  For example:

  ```yaml
  - filter: mcp
    protocol_profile: stateless
    servers:
      - name: weather
        cluster: weather-mcp
        path: /mcp
        tools:
          - name: get_weather
            description: Get current weather by city.
          - name: get_forecast
            description: Get a multi-day forecast by city.
  ```

  In this mode, servers: is not just descriptive config. It is the broker's source of truth for both discovery and routing.

  When a client sends tools/list, the broker builds the response from the configured catalog. In the example above, the client sees both `get_weather` and `get_forecast`.

  When a client later sends tools/call for `get_weather`, the broker can look up that tool in the same catalog and determine that the call belongs to the `weather-mcp cluster`. A call for get_forecast resolves the same way because it is also owned by the configured weather server.

  That means the catalog answers two questions:

  - Which tools should this broker advertise to MCP clients?
  - Which backend cluster should receive a call for each advertised tool?

  The forwarded request may also need to rewrite headers or body fields before continuing through the normal upstream path. For example,  the public tool name advertised to clients may differ from the backend tool name, so the broker needs the body/header mutation path to survive after request-body inspection.

  Because of that, a broker-configured mcp filter with servers: is a cluster-selecting filter, while classifier-only mcp is not. This PR teaches core pipeline validation that distinction and fixes the StreamBuffer header mutation path needed for the broker to forward calls correctly.

  The MCP broker itself lives in the AI filter crate, not in Praxis core. Praxis core does not implement MCP discovery or `tools/call` behavior.

  This PR only updates the generic core pipeline machinery that the AI MCP broker depends on:

  - recognizing that broker-configured `mcp` filters can select clusters
  - validating those selected clusters against the load balancer
  - preserving request header mutations made after request-body inspection

  That keeps MCP protocol behavior in the AI repo while allowing the core pipeline to correctly validate and forward requests for filters  that make routing decisions after reading the body.

  ## Scope

  This is core pipeline plumbing only.

  It does not add:

  - new MCP protocol behavior
  - new AI filter behavior
  - backend MCP `tools/call` routing by itself
  - new config examples
  - new public filter configuration

  The behavior change is limited to making existing body-phase request header mutations survive correctly and improving validation for broker-style cluster selection.

  ## Validation

  - `cargo test -p praxis-proxy-filter pipeline`
  - `cargo test -p praxis-tests-integration --test suite stream_buffer_adapter`
  - `cargo clippy --workspace --all-targets -- -D warnings`
  - `cargo +nightly fmt --all --check`
  - `git diff --check`

  ## References

  This PR is a core prerequisite for the MCP broker/gateway work tracked in the agentic epic.

  It specifically supports the following epic items:

  - [praxis-proxy/praxis#196](https://github.com/praxis-proxy/praxis/issues/196) — StreamBuffer/body mutation adapter work needed for body-inspecting filters that rewrite forwarded requests.
  - [praxis-proxy/praxis#197](https://github.com/praxis-proxy/praxis/issues/197) — Pipeline validation for MCP broker/gateway-style cluster selection.
  - [praxis-proxy/praxis#344](https://github.com/praxis-proxy/praxis/issues/344) — MCP gateway/broker static catalog and stateless `tools/call` routing.

  The MCP protocol behavior remains in the AI repo. This PR only fixes the core pipeline plumbing needed for that follow-up work.

  Refs praxis-proxy/ai#148


filter/src/pipeline/checks.rs
```diff
@@ -137,6 +137,29 @@ pub(super) fn check_duplicate_load_balancers(names: &[&str], errors: &mut Vec<St
...
+/// later one silently overwrites the earlier selection.
+#[expect(clippy::indexing_slicing, reason = "enumeration bounds")]
+pub(super) fn check_conflicting_cluster_selectors(entries: &[FilterEntry], errors: &mut Vec<String>) {
+    for (i, entry) in entries.iter().enumerate() {
+        if entry.filter_type != "load_balancer" {
+            continue;
+        }
+        let preceding = &entries[..i];
+        let has_router = preceding.iter().any(|e| e.filter_type == "router");
+        let has_mcp_broker = preceding.iter().any(is_mcp_broker_entry);
```
usize: local LLM review:

`is_mcp_broker_entry` checks for a `servers` key in the config YAML but it doesn't verify `entry.filter_type == "mcp"`. The `entry_selects_cluster` caller is safe because it gates on the `"mcp"` match arm first, but here the predicate is applied to all preceding entries regardless of filter type. Any non-MCP filter whose config happens to contain a `servers:` sequence value would be falsely identified as a broker, producing a spurious conflict error.

Consider adding the type guard to `is_mcp_broker_entry` itself so both call sites are safe:

```rust
fn is_mcp_broker_entry(entry: &FilterEntry) -> bool {
    entry.filter_type == "mcp"
        && entry.config.get("servers")
            .and_then(|servers| servers.as_sequence())
            .is_some()
}
```

This is backward-compatible with `entry_selects_cluster` (the redundant `"mcp"` match arm just becomes a double-check).

protocol/src/http/pingora/handler/request_filter/mod.rs
```diff
@@ -100,14 +100,25 @@ pub(in crate::http) async fn execute(
     if matches!(caps.request_body_mode, BodyMode::StreamBuffer { .. }) {
         tracing::debug!("pre-reading request body for StreamBuffer inspection");
         match stream_buffer::pre_read_body(pipeline, session, ctx, &request).await {
-            Ok(extra_headers) => {
-                tracing::debug!(count = extra_headers.len(), "injecting promoted headers");
-                for (name, value) in extra_headers {
+            Ok(pre_read_headers) => {
+                let req_headers = session.req_header_mut();
```
usize: nit: remove->set->extra behavior could be non-obvious. A brief comment noting the intended ordering would help future readers understand it's deliberate rather than incidental.

usize:
Core of this looks good to me. Two small things from my local review. Otherwise 🚢

usize (approved):

filter/src/pipeline/checks.rs
nerdalert: Added the `entry.filter_type == "mcp"` guard inside `is_mcp_broker_entry` so the predicate is safe at every call site, not just behind `entry_selects_cluster`. Also added a regresion test with a non-MCP filter that happens to have a servers, sequence to prove it does not conflict with router. Thanks!

protocol/src/http/pingora/handler/request_filter/mod.rs
nerdalert: Added a smol comment before the body-phase header mutation loops explaining the deliberate ordering: removals first, then explicit replacements, then promoted metadata headers. Ty!

shaneutt (changes requested):
Needs some conflict resolution!

nerdalert:
@shaneutt resolved!

filter/src/pipeline/checks.rs
```diff
@@ -167,6 +190,29 @@ pub(super) fn check_misaligned_clusters(entries: &[FilterEntry], errors: &mut Ve
...
+        "mcp" => is_mcp_broker_entry(entry),
+        _ => false,
+    }
+}
+
+/// Broker mode is identified by the `servers` list because classifier-only
+/// MCP filters do not select clusters for load balancing.
+fn is_mcp_broker_entry(entry: &FilterEntry) -> bool {
+    entry.filter_type == "mcp"
+        && entry
```
shaneutt: This means the core pipeline validation is learning MCP config semantics. Note that `router` has some precedents around this we could potentially follow: perhaps a declarative approach (e.g. `selects_cluster` on `FilterEntry` or a trait method) so future cluster-selecting filters don't require core changes. Otherwise we might have to add special handling for all these AI protocols, and we need to try and keep that a separate concern where feasible.

shaneutt (changes requested):

filter/src/pipeline/checks.rs
nerdalert: Excellent. I spent some time evaluating some options with your feedback. Core now asks each filter whether it routes to a cluster instead of hardcoding MCP specific config.

Here is a summary of the changes and reasoning. Ty!

---

• ## Design Decision

  This PR originally handled the MCP broker case by teaching core pipeline validation that an `mcp` filter with `servers:` can select an upstream cluster. That fixed the immediate validation problem, but it also made core validation aware of MCP-specific configuration structure.

  The revised approach keeps the validation rule in core, but moves protocol/filter-specific knowledge back to the filter implementations.
  
Core now validates a generic pipeline contract:

  - some filters can select a cluster
  - some filters can consume/load-balance clusters
  - every selected cluster must exist in the downstream load balancer
  - multiple cluster-selecting filters before the same load balancer are ambiguous and should be rejected

  To support that, HTTP filters can now declare pipeline capabilities:

  - `selects_cluster()`
  - `selected_clusters()`
  - `load_balancer_clusters()`

  Core pipeline validation asks instantiated filters for those capabilities instead of parsing filter-specific YAML. This keeps core focused on routing semantics, not protocol configuration schemas.

  Options considered:

  - Keep the MCP-specific special case in core. This was the smallest change, but it would make core validation understand AI protocol details such as MCP `servers:`. Future cluster-selecting filters would likely need more special cases.
  - Add a declarative cluster-selection field to `FilterEntry`. This would avoid hard-coded MCP parsing, but it would put behavioral metadata in config rather than in the filter implementation that actually owns the behavior. That creates drift risk if config declarations and runtime behavior diverge.
  - Add capability methods on filters. This is the chosen approach. The filter implementation declares what it does, and validation consumes those declarations generically.

  With this change, the built-in `router` declares itself as a cluster selector and reports its route target clusters. The built-in `load_balancer` reports the cluster names it can consume. A follow-up AI MCP broker can implement the same capability hooks when it routes `tools/call` to backend MCP servers, without adding MCP-specific logic to core validation.

  This keeps the extension point reusable for other protocol-aware filters as well: core does not need to know whether a cluster selection came from router config, MCP broker catalog config, or a future filter. It only needs to know that the filter may set `ctx.cluster` and which cluster names are valid for cross-reference checks.

tests/integration/tests/suite/stream_buffer_adapter.rs
```diff
@@ -114,6 +208,23 @@ fn registry_with_mutator() -> praxis_filter::FilterRegistry {
...
+                Ok(Box::new(praxis_test_utils::filters::HeaderMutatingStreamBufferFilter))
+            })),
+        )
+        .expect("duplicate filter name");
+    registry
+}
+
+// -----------------------------------------------------------------------------
+// YAML Helpers
+// -----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
nerdalert: Fixed. Renamed the separator to YAML Utilities to match repo convention

filter/src/pipeline/clusters.rs
```diff
@@ -168,16 +131,72 @@ mod tests {
...
-            filter_type: filter_type.to_owned(),
-            config: serde_yaml::from_str(yaml).expect("valid test YAML"),
+            filter: AnyFilter::Http(Box::new(filter)),
+            filter_id: 0,
             name: None,
             response_conditions: vec![],
         }
     }
+
+    struct CapabilityFilter {
```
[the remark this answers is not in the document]
nerdalert: Fixed. Extracted the duplicated test-only capability filter into filter/src/pipeline/test_filters.rs and updated both checks.rs and clusters.rs tests to use the shared utility.

filter/src/pipeline/checks.rs
```diff
@@ -137,11 +137,39 @@ pub(super) fn check_duplicate_load_balancers(names: &[&str], errors: &mut Vec<St
     }
 }
 
+/// Multiple cluster-selecting filters before the same load balancer
+/// compete for `ctx.cluster`; the later one silently overwrites the
+/// earlier selection.
+#[expect(clippy::indexing_slicing, reason = "enumeration bounds")]
+pub(super) fn check_conflicting_cluster_selectors(filters: &[PipelineFilter], errors: &mut Vec<String>) {
+    for (i, filter) in filters.iter().enumerate() {
```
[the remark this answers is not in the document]
nerdalert: Fixed. check_conflicting_cluster_selectors now collapses duplicate router entries so duplicate-router validation owns `[router, router, load_balancer]`, while `[router, router, custom_selector, load_balancer]` still reports the real selector conflict. Added regression tests for both cases.

filter/src/filter.rs
```diff
@@ -67,6 +67,41 @@ pub trait HttpFilter: Send + Sync {
...
+    ///
+    /// Pipeline validation uses this to detect `load_balancer`
+    /// configurations that cannot receive a cluster selection and to
+    /// reject multiple competing cluster selectors before the same
+    /// load balancer. Filters that route to named clusters should
+    /// override this.
+    ///
+    /// [`HttpFilterContext::cluster`]: crate::HttpFilterContext::cluster
+    fn selects_cluster(&self) -> bool {
+        false
```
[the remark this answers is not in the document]
nerdalert: Fixed. Added test coverage for the default selects_cluster, selected_clusters, and load_balancer_clusters trait methods.

tests/integration/tests/suite/stream_buffer_adapter.rs
```diff
@@ -101,6 +101,100 @@ fn stream_buffer_readwrite_content_length_repaired() {
...
 }
 
+// -----------------------------------------------------------------------------
+// Body-phase Header Writeback Tests
+// -----------------------------------------------------------------------------
+
+#[test]
+fn stream_buffer_body_phase_headers_to_set_reach_backend() {
+    let backend_guard = start_header_echo_backend();
+    let proxy_port = free_port();
```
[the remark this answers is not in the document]
nerdalert: Checked this against the current branch. start_header_echo_backend() returns BackendGuard, so .port() is valid here. 

The focused `stream_buffer_adapter` integration test also passed, confirms  the `.port()` call is valid on this branch.

shaneutt (approved):

PR #764, opened 2026-07-07
feat(filter): per-filter duration Prometheus metrics

### LINKED ISSUE #9: Per-filter metrics
Track execution timing per filter per phase, enabling operators to identify slow filters and optimize pipelines.

Parent: #8

- [ ] `praxis_filter_duration_seconds` histogram (labels: filter, phase = request | response)
- [ ] Instrument `FilterPipeline` execution around each `on_request`, `on_response`, and body hook
- [ ] Skip recording for condition-gated filters that didn't execute
- [ ] Integration test verifying metrics are emitted

aslakknutsen:
Closes #9

## Summary

- Emit `praxis_filter_duration_seconds` from HTTP filter hooks (headers and body, request and response) with `filter`, `phase`, and `stream` labels
- Opt in via `metrics.filter_duration` (default off); requires admin endpoint to scrape
- Refactor: consolidate hook runners in `http_utils.rs` with unified `HeaderFilterOutcome`; move TCP/condition gating into `execute_http_request`

## Test plan

- [x] Unit tests for all four hooks, condition-gated skip, and disabled config
- [x] Integration test via `admin-interface.yaml` example
- [x] Schema test for `metrics.filter_duration` parsing

## Reviewer note

Probably easier to review the two commits separate. The first does a refactor so the second looks cleaner. :)


core/src/config/validate/rules.rs
```diff
@@ -49,6 +49,7 @@ impl Config {
         validate_filter_chains(&self.filter_chains, &self.listeners)?;
         validate_branch_chains(&self.filter_chains)?;
         validate_admin_address(self.admin.address.as_deref(), self.insecure_options.allow_public_admin)?;
+        warn_filter_duration_without_admin(self.metrics.filter_duration, self.admin.address.is_some());
```
aslakknutsen, line 53: Should this be a warning, or should "admin" be a master switch. If admin is off, metrics is off. (maybe admin is enabled later via some remote protocol? _shrug_)

core/src/config/validate/rules.rs
```diff
@@ -184,6 +185,16 @@ fn validate_admin_address(addr: Option<&str>, allow_public: bool) -> Result<(),
     Ok(())
 }
 
+/// Warn when filter duration metrics are enabled but the admin endpoint is disabled.
+pub(super) fn warn_filter_duration_without_admin(filter_duration: bool, admin_enabled: bool) {
+    if filter_duration && !admin_enabled {
+        warn!(
```
didierofrivia, line 191: Q: Is there any value on recording the metrics without exposing them?

core/src/config/metrics.rs
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! Metrics and observability configuration.
+
+use serde::Deserialize;
```
didierofrivia: Minor/Nit: `use serde::{Deserialize, Serialize};` So in line 25: `#[derive(Debug, Clone, Default, Deserialize,Serialize)]`

core/src/config/validate/rules.rs
aslakknutsen: That's the big question. :) 

I would say generally no, beyond perhaps some vague idea of performance testing without exposing the admin part, but that's so far fetched I'm not sure why I'm even bringing it up. 

The only valid case I have in mind is if the admin listener could be dynamically started/stopped via some remote api endpoint.. then you could be able to get the metrics since the server started, not since the admin listener was started. but yeah..

core/src/config/validate/rules.rs
didierofrivia: didn't see this comment, hence [my question here](https://github.com/praxis-proxy/praxis/pull/764#discussion_r3536812758) I was wondering the value on recording without exposing, maybe... is there a way to invalidate the `metrics.filter_duration` if admin is off?

core/src/config/metrics.rs
aslakknutsen: fixed https://github.com/praxis-proxy/praxis/pull/764/commits/80e07da5961beff4bf3ad1862ecda5d6035a897d

core/src/config/validate/rules.rs
aslakknutsen: I'm sure we can set it to false if admin is off and warn about not collecting instead. Depends how smart we want to be.

filter/src/pipeline/mod.rs
```diff
@@ -180,6 +183,16 @@ impl FilterPipeline {
         self.health_registry = Some(registry);
     }
 
+    /// Enable or disable recording of per-filter duration metrics.
+    pub fn record_filter_duration_metrics(&mut self, enabled: bool) {
```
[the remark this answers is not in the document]
aslakknutsen: fixed

filter/src/pipeline/tests.rs
```diff
@@ -3351,3 +3365,247 @@ fn multiple_pipeline_extensions_are_all_injected() {
...
+mod filter_duration_metrics_tests {
+    use std::sync::{Mutex, OnceLock};
+
+    use bytes::Bytes;
+    use metrics_exporter_prometheus::PrometheusBuilder;
+
+    use super::*;
+
+    static RECORDER: OnceLock<metrics_exporter_prometheus::PrometheusHandle> = OnceLock::new();
+    static RECORDER_LOCK: Mutex<()> = Mutex::new(());
```
alexsnaps: Confused, that's this guarding?

aslakknutsen: Oh, right.. Canada. I suppose you don't double lock your doors. 🔒

aslakknutsen: fixed https://github.com/praxis-proxy/praxis/pull/764/commits/ee5bf26e0ff69275945153766c8978fea854bbc9

shaneutt (approved):

filter/src/pipeline/tests.rs
alexsnaps: Oh duh. Hadn't seen that pattern in ages. Now I get you. But no, none of that "here" ;)

PR #767, opened 2026-07-07
fix(observability): report correct response_body_bytes in access_log

### LINKED ISSUE #383: access_log: response_body_bytes always 0 when no response-body filter is configured
## Summary

access_log filter reports `response_body_bytes=0` for any response when no filter in the chain declares response-body access. Request-side counter works correctly because the experimental config typically has a filter that declares request-body access.

## Reproducer

```yaml
listeners:
  - name: gw
    address: "0.0.0.0:8080"
    filter_chains: [main]
filter_chains:
  - name: main
    filters:
      - filter: model_to_header
        header: X-Model
      - filter: router
        routes: [{ path_prefix: "/", cluster: backend }]
      - filter: load_balancer
        clusters:
          - name: backend
            load_balancer_strategy: { consistent_hash: { header: X-Model } }
            endpoints: ["upstream:8000"]
      - filter: access_log
```

## Observed

```
access method=POST path=/v1/completions client_ip=10.128.x.x status=200
       duration_ms=286 cluster="vllm" upstream="-" request_id="e08febd9-..."
       request_body_bytes=143 response_body_bytes=0
```

`response_body_bytes=0` despite the upstream returning a non-empty response. `request_body_bytes=143` is correct.

## Expected

`response_body_bytes` reflects the bytes written downstream, independent of whether any filter declared response-body access. 

overall if this is not expected behavior I can resolve it.

crstrn13:
## Summary

- Fixes `access_log` reporting `response_body_bytes=0` when no filter in the chain declares response-body access
- Moves logging from `on_response` to `on_response_body` (at end-of-stream) so the accumulated byte count is available
- Declares `response_body_access: ReadOnly` so the pipeline's byte-counting path runs
- Saves response status in `filter_state` during `on_response` since `response_header` is unavailable during the body phase

Closes #383

## Test plan

- [x] Existing `access_log` unit tests pass (28 tests)
- [x] New unit tests for `on_response` status storage, `on_response_body` behavior, and `response_body_access` declaration
- [x] Integration test `access_logging` passes
- [x] Manual verification: start proxy with `access_log` filter, curl, confirm `response_body_bytes` is non-zero


filter/src/builtins/http/observability/access_log.rs
```diff
@@ -120,16 +121,31 @@ impl HttpFilter for AccessLogFilter {
...
+        BodyAccess::ReadOnly
+    }
+
+    fn on_response_body(
+        &self,
+        ctx: &mut HttpFilterContext<'_>,
+        _body: &mut Option<Bytes>,
+        end_of_stream: bool,
+    ) -> Result<FilterAction, FilterError> {
+        if end_of_stream && self.should_log() {
```
shaneutt, line 175: `on_response_body` isn't invoked for bodyless responses (HTTP 204, 304, HEAD requests). Pingora's `read_response_task()` returns `HttpTask::Header(header, end_of_body=true)` for these and never produces `HttpTask::Body`, so `response_body_filter` is never called. This silently drops the access log for all bodyless requests (a regression).

Affected request types:
- **204 No Content** — common for REST DELETE/PUT/PATCH
- **304 Not Modified** — cache revalidation (If-None-Match, If-Modified-Since)
- **HEAD** requests — health checks, pre-flight probes

Suggested Fix: detect in `on_response` when the response will have no body (by checking status code 204/304 or noting the method) and emit the log directly there with `response_body_bytes = 0`, falling through to `on_response_body` only for responses that will have a body.

filter/src/builtins/http/observability/access_log.rs
```diff
@@ -431,6 +447,95 @@ mod tests {
...
+        let mut body = Some(Bytes::from_static(b"partial"));
+        let action = filter.on_response_body(&mut ctx, &mut body, false).unwrap();
+        assert!(
+            matches!(action, FilterAction::Continue),
+            "on_response_body should continue before end_of_stream"
+        );
+    }
+
+    #[test]
+    fn on_response_body_continues_at_end_of_stream() {
```
shaneutt: Need tests to cover all the bodyless response paths

shaneutt (changes requested):

shaneutt (approved):

PR #805, opened 2026-07-14
docs: add iterative-request-router proposal

### LINKED ISSUE #786: Spike: Agentic Loop Support
> **Note**: Discussed in https://github.com/orgs/praxis-proxy/discussions/777
> **Important**: https://github.com/orgs/praxis-proxy/discussions/87 relates

## Problem

Agentic AI loops need a decision after an upstream response, followed by another upstream exchange:

```
call model -> read tool calls -> run tools -> call model again -> repeat
```

Each iteration must target a **configured upstream cluster** (with routing, load balancing, endpoint selection, TLS, retries, and health checks) and pass through the **full filter pipeline**: guardrails, auth, token accounting, logging, model rewrite.

This cannot be solved with existing primitives:

- **Branch chains** evaluate during the request phase, before the upstream response exists. Response-derived decisions (like tool calls from a model) cannot trigger branch re-entry.
- **Sub-requests from a filter** (see [#87](https://github.com/orgs/praxis-proxy/discussions/87)) go to arbitrary endpoints and bypass the filter pipeline. Intermediate model calls need the same guardrails and observability as the initial call. A sub-request API also pushes proxy-level responsibility (cluster routing, TLS, retries, response replacement) into filter logic.

## Assessing New Primitives

What is missing is an **exchange-level loop**: run a complete HTTP exchange (request filters, upstream, response filters), inspect the result, and optionally run another complete exchange before returning a final response to the client.

This is different from branch chains (request-phase composition within a single exchange) and different from sub-requests (filter-owned callouts to external services). It loops over the proxy's own machinery; each iteration is a full proxy exchange through the configured pipeline targeting the configured cluster.

shaneutt:
This proposal describes the motivations and use cases regarding sub-request processing which supports multiple AI, and non-AI specific features including agentic loop orchestration, and inference provider failover.

### DIFF
--- a/docs/proposals/00786_iterative-request-router.md
+++ b/docs/proposals/00786_iterative-request-router.md
@@ -1,0 +1,151 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/786
+discussion:
+  - https://github.com/praxis-proxy/praxis/discussions/777
+  - https://github.com/orgs/praxis-proxy/discussions/87
+  - https://github.com/praxis-proxy/ai/discussions/287
+status: proposed
+authors:
+  - shaneutt
+graduation_criteria:
+  - How? section with requirements and design
+  - Sub-request executor design validated
+  - Security model validated (SSRF, credential isolation)
+stakeholders:
+  - alexsnaps
+  - leseb
+  - usize
+  - franciscojavierarceo
+  - shaneutt
+---
+
+# Iterative Request Router
+
+## What?
+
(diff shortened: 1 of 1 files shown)


alexsnaps (approved):
Not that I "disapprove" of the non-goals. But I can see how doing things a certain way _could_ either enable some of those (e.g. parallel requests), or make them harder to incorporate in the future. 

So, while I agree these shouldn't be included in this proposal, possibly accounting for them in some of our decision making might be desirable.

PR #806, opened 2026-07-14
feat(policy): generic-HTTP authorization derived from the policy

### LINKED ISSUE #716: Spike: Supporting AuthPolicy

araujof:
Adds generic-HTTP (L7) authorization to the `policy` filter and removes the operator-facing enforcement mode: the filter derives how it evaluates from the loaded policy.

Closes: #716 

This PR helps Praxis better support AuthPolicy L7 HTTP policies. [Demo](https://github.com/praxis-proxy/demos/tree/main/demos/authpolicy-transpiler): https://github.com/praxis-proxy/demos/pull/6

## Behavior
- A policy with only a `global` HTTP policy (no entity routes) is authorized at the HTTP layer in `on_request` over `http.*` + identity. No classifier, no body buffering. This is the L7 HTTP-proxy path.
- A policy with entity routes (tool/prompt/resource) is authorized per-entity at the body phase after classification, enriched with `http.*` so one rule can combine HTTP and entity/`args.*` attributes.
- An entity-route policy that reaches the body phase without `protocol.method` fails closed with a diagnostic warning (classifier missing or ordered after `policy`). A policy that declares both a global HTTP policy and entity routes with `require_protocol_metadata: false` warns at startup, since non-classified requests would skip the global HTTP policy.

## Changes
- `policy` filter: derive `http_global` / `entity_routes` from the loaded policy via `has_hooks_for`; dispatch by shape; enrich entity evaluation with `http.*`.
- Remove the `enforcement` config field (behavior is derived).
- Example `examples/configs/security/policy-http.yaml`, filter doc, README, and integration test `policy_http`.
- Requires cpex 0.2.2, which grants `read_headers` to entity routes so `http.*` is available at the entity phase.

Tests: filter lib and `policy_http` e2e pass under `cpex-policy-engine`; schema/example validation passes; the default build (feature off) is unaffected; clippy and fmt clean.

## Note: hot-reload test flakiness (unrelated)
The coverage gate intermittently failed on `hot_reload::reload_removes_filter`, unrelated to this change: the coverage build runs with default features, so the `policy` module (all changes here) is not compiled. The reload tests asserted post-reload behavior after a fixed settle window, which races the async watcher apply under `cargo llvm-cov` instrumentation. This PR also makes those tests poll for the new behavior (bounded 15s timeout) instead of relying on a fixed sleep, so the gate is deterministic.


tests/integration/tests/suite/hot_reload.rs
```diff
@@ -14,6 +14,27 @@ use praxis_test_utils::{
...
+/// caller's assertion produces a clear diff).
+///
+/// Hot reload is applied asynchronously by the file watcher (500ms debounce
+/// then pipeline rebuild). Under coverage instrumentation the apply can exceed
+/// any fixed settle window, so a post-reload assertion must poll for the new
+/// behavior rather than race a single request. This does not mask failures:
+/// if the reload never takes effect the poll times out and the last (wrong)
+/// response is returned for the assertion to reject.
+fn get_eventually(addr: &str, path: &str, pred: impl Fn(u16, &str) -> bool) -> (u16, String) {
+    let deadline = Instant::now() + Duration::from_secs(15);
```
alexsnaps, line 28: This is quite some long time... until it then now fails, but... yeah... timing  ¯\_(ツ)_/¯

alexsnaps (approved):
I really want to unblock this. Sorry for having taken that long.

tests/integration/tests/suite/hot_reload.rs
araujof: Agreed. I think we can reduce it to 5 secs.

araujof: I was looking at this again, and think that maybe we should leave the deadline as is. The 15s here is a poll deadline, not a sleep; `get_eventually` polls every 50ms and returns the instant the reload takes effect, so passing runs finish in milliseconds. The 15s only elapses when a reload genuinely never applies (then we return the stale response so the assertion fails with a clear diff). The generous ceiling is there to absorb the 500ms watcher debounce + pipeline rebuild under coverage instrumentation, where a tighter window flakes. So it's a worst-case bound rather than a per-run cost, and I'd rather keep it to avoid coverage-run flakiness. Thanks for the review!

PR #815, opened 2026-07-16
fix: protect inline credential config values

franciscojavierarceo:
## Summary

Inline `credential_injection` values were deserialized as plain `String`s. Store inline values as `SecretString` instead, exposing them only when constructing the outbound header.

- deserialize `clusters[].value` as `SecretString`
- borrow the secret at the injection boundary rather than cloning it
- preserve existing environment-variable resolution and header-injection behavior
- update the generated filter reference to mark inline values as secret

## Compatibility

The YAML schema and runtime behavior are unchanged. Existing inline `value` and `env_var` configurations continue to work; only the in-memory representation of inline values changes.

## Validation

- `cargo test -p praxis-proxy-filter -q`
  - 1,566 unit tests passed; 58 doctests passed
- `make lint`

## Review

- Rust-guidelines review: no findings
- pre-landing review: no findings


leseb (approved):

alexsnaps (approved):

shaneutt:
The `secrecy` dependency we've added looks stale: the last release was in 2024.

Now, having no release for a long time isn't bad within itself. The software may just be "stable" (imagine that in this age). However I'm noticing a lot of its own dependencies are out-of-date and may have some issues. I agree with the desired effect, but we may want to give a second look at the dependency.

I recommend putting an issue in for v0.5.0 to do a dependency review, which would cover this dependency as well as others as a follow-up.

PR #818, opened 2026-07-17
Remove AI authorship enforcement workflow

franciscojavierarceo:
## Summary

- Remove the `ai-authorship-check` job that rejects commits based on AI/tool authorship metadata.
- Document that human contributors remain responsible for authorship, review, and sign-off in the shared agent guidance and PR template.

## Why

Claude Code is becoming a standard development tool for RH engineers working on Praxis. The workflow adds friction to normal development without replacing human accountability. The responsibility remains explicit in contributor guidance, while enforcement is handled by the human submitting and reviewing the pull request.

## Validation

- `git diff --check`
- Ruby YAML parse of `.github/workflows/conventions.yaml`


shaneutt (changes requested):
This authorship check is part of our developer conventions: it is meant to be a gate that the human author needs to recognize. Acknowledging responsibility for code people submit is part of our social contract.

Contributors are meant to take responsibility for the code they submit, and this check is their first signal of that social contract for new contributors (if they did not bother to read the development docs, especially). If the desire to remove this is in some part due to not wanting to have to stop and review, modify, and sign-off on commits that contain generated code then this check is working as intended.

On a personal note: Call me an iconoclast, but I find it _extremely silly_ to see `Co-authored by $tool` on commits all over Github. As if you would every say `Co-authored by rustfmt`, or `Co-authored by my copy-paste-buffer`. As if an LLM is some entity that understands the meaning of authorship, rather than being a auto-correct on steroids which was trained on massive amounts of code and dumps code out from that corpus as contextually relevant anagrams.

I'm not open to removing this outright given the established philosophies of the project. If we can very clearly articulate the issues we're having that need fixing, I am open to suggestions for alternatives / other approaches we might take.

Issue #796, opened 2026-07-10
Pipeline info admin endpoint


shaneutt:
## Summary

Admin endpoint that returns the resolved filter pipelines for each listener. Enables operators to inspect what filters are active, in what order, with what conditions and branch points, without reading the config file.

## Design

New admin endpoint: `GET /api/pipelines`

Response structure (JSON):
- Per listener: address, protocol, TLS status
- Resolved filter chain: filter name, type, phase hooks (on_request, on_response, etc.), conditions, body mode
- Branch points: condition, target chain, rejoin point
- Chain metadata: total filter count, pipeline name

Implementation reads from the live `Arc<ArcSwap<FilterPipeline>>`, the same source the handler uses at runtime. Returns the resolved view, not the raw config.

## Acceptance criteria

- [ ] Endpoint returns resolved pipelines per listener
- [ ] Includes filter order, types, conditions, branch points
- [ ] Reflects current state after dynamic config reload
- [ ] Integration test: load config, query endpoint, verify structure matches expected pipeline
- [ ] JSON output parseable by `praxisctl pipelines` (#793)

## Related issues

- #793 praxisctl CLI diagnostic tool
- #785 Spike: gRPC/HTTP control plane API for runtime config push

PR #823, opened 2026-07-20
docs(proposals): add pipeline info admin endpoint proposal

henschwartz:
## Summary

- Adds the What?/Why? proposal for [#796](https://github.com/praxis-proxy/praxis/issues/796) to expose a read-only admin API for live resolved filter pipelines per listener.
- Covers listener transport facts, ordered filters, conditions, branch points, and reload-accurate runtime state.
- Leaves How? out of this PR per proposal process; open questions cover field set, admin authz, empty/partial states, and aggregate vs per-listener paths.

## Test plan

- [ ] Confirm goals and acceptance criteria match #796
- [ ] Confirm commit is signed and includes `Signed-off-by`
- [ ] Wait for proposal-check / signed-commits-check / signoff-check to pass


docs/proposals/00796_pipeline-info-admin-endpoint.md
twghu, line 1: Open Questions:

1. It would be reasonable to use serde_json, beyond that recommendation we can then filter out what we might not  want to emit as a follow up task.
2. Follow existing policy for admin surfaces.  This is essentially appears to be a query only option with operators etc controlling updates.   Bind to loopback by default. 
3. Reload failures are atomic (nothing to represent), topology-pending listeners are simply absent from the live-sourced response, empty chains are shown as empty (not omitted), and pipeline_name should be replaced with a chain_names array to handle zero/one/many named chains uniformly.  Represent empty chains as filters: [], filter_count: 0.  
4. Aggregation, this raises a a good point.  It should be resolvable per-listener as in large topologies overhead is compiling/processing meta data on to be discarded is inefficient.  It's also a question of speedy debug/support resolution.   A format that supports  JSONPath supportive format is sufficient.  Using serde_json should be ideal and sufficient.  In short, if we can refine a target at a low cost then it saves scrolling and parsing megabytes of output.

twghu, line 1: Some low cost considerations:
  1. Route /api/pipelines into the existing admin dispatch. Admin endpoints (/healthy, /metrics, /ready, /api/kv/*) all live in one PingoraAdminService::response match statement (protocol/src/http/pingora/health/service.rs), bound to a single loopback-only port. A pipelines route slots in the same way — no new listener, no new service plumbing.
  2. Reuse the existing access model as-is. There's no auth today — access control is just "admin bind address must be loopback unless insecure_options.allow_public_admin: true." That's already a considered, existing security boundary. 
  3. Admin surface naturally excludes itself from the dump. AdminConfig is a separate top-level field from config.listeners and never enters ListenerPipelines/ArcSwap. Sourcing /api/pipelines from ListenerPipelines::listener_names() means the admin listener just doesn't show up — no special-casing needed to keep it out.

docs/proposals/00796_pipeline-info-admin-endpoint.md
```diff
@@ -0,0 +1,133 @@
...
+open where they are not already constrained by the goals below.
+
+### Goals
+
+- Expose a read-only admin API that returns resolved pipelines
+  per listener
+- Include, for each listener, address, protocol, and TLS status
+- Include the resolved filter chain: filter identity/type, phase
+  hooks that apply, conditions, and body mode
+- Include branch points: condition, target chain, and rejoin point
```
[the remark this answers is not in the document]
henschwartz: done

docs/proposals/00796_pipeline-info-admin-endpoint.md
henschwartz: added in the Decisions section

henschwartz: added under the decisions section and under goals

henschwartz:
Rerun CI after a flaky Integration failure unrelated to this doc change.

leseb (commented):
LGTM, just one nit, I found the "admin-API" a little confusing in the sense that "admin" commonly implies authenticated, privileged access. So can we add a note like:

> Unauthenticated operator endpoint on the admin listener

Thanks!

shaneutt (approved):

PR #825, opened 2026-07-21
feat(metrics): expand Prometheus metrics surface (#794)

### LINKED ISSUE #794: Expand Prometheus metrics surface
## Summary

Praxis exposes only two Prometheus metrics today: `praxis_http_requests_total` (counter) and `praxis_http_request_duration_seconds` (histogram). This is insufficient for production monitoring. Add metrics covering request/response sizes, connection state, upstream performance, health status, and config reload lifecycle.

## New metrics

Request/response sizing:
- `praxis_http_request_body_bytes` (histogram)
- `praxis_http_response_body_bytes` (histogram)

Connection state:
- `praxis_connections_active` (gauge) — concurrent connections per listener

Upstream performance:
- `praxis_upstream_connect_duration_seconds` (histogram) — time to establish upstream connection

Health:
- `praxis_upstream_health_status` (gauge, per cluster/endpoint) — 1 healthy, 0 unhealthy

Config reload:
- `praxis_config_reload_total` (counter, labels: result=success|failure)
- `praxis_config_reload_last_success_timestamp` (gauge)

All metrics use the existing `metrics` crate facade and `metrics-exporter-prometheus` backend.

## Acceptance criteria

- [ ] All listed metrics implemented and emitting
- [ ] Labels follow existing conventions (method, status_class, cluster)
- [ ] Unit tests verifying metric recording
- [ ] Integration test: send requests, scrape `/metrics`, verify presence and correctness
- [ ] No measurable latency regression from additional instrumentation

## Related issues

- #9 Per-filter metrics (`praxis_filter_duration_seconds`)
- #383 access_log: response_body_bytes always 0

henschwartz:
## Summary
- Implements the #794 / #822 baseline on admin `/metrics`: body size histograms, `connections_active` (HTTP+TCP), overload rejects, upstream connect duration/failures/retries, circuit-breaker open gauge, LB panic mode (HTTP+TCP), cluster health aggregates + transitions (active+passive), and config reload success/failure + last-success timestamp.
- Populates the existing `route` label from router `path_match` (bounded cardinality) and hooks instrumentation into existing runtime paths rather than adding parallel subsystems.
- Adds unit coverage plus an integration scrape suite (`prometheus_metrics.rs`); includes reload safety (cancel in-flight probes, clear/seed health gauges, CB gauge cleanup on drop).

## Explicit follow-ups (out of scope here)
- Connect-failure `reason` taxonomy
- `praxis_filter_rejections_total`
- Process uptime gauge
- Per-family `MetricsConfig` opt-in flags beyond recorder/`admin` gating

## Test plan
- [x] `make test`
- [x] `make lint`
- [ ] CI green on this PR

Closes #794


shaneutt:
Keep the branch, but let's consider this closed for the moment while we're still doing the proposal and re-open accordingly afterwards.

benchmarks/microbenchmarks/common.rs
twghu, line 1: - Order of insertion of new field in struct.
- Does the new field even belong in the struct?  Seems this isn't the place to store the cluster name.

RouterFilter doesn't hold "many contexts" — it holds one static routes: Vec<ResolvedRoute> (router/mod.rs:108), built once in resolve_routes() at filter construction (from YAML config), and every request's on_request just picks one existing entry out of it via match_route(). It's a small, fixed table, not something that grows per request.

ResolvedRoute (router/mod.rs:113) already precomputes one derived value at construction time for exactly this reason — wildcard_suffix, whose doc comment literally says "pre-computed... " so it doesn't get recomputed per request. metrics_route should get the same treatment: add a metrics_label: ::metrics::SharedString field to ResolvedRoute, compute it once in resolve_routes() from route.path_match.value(), and then on_request becomes ctx.metrics_route = Some(route.metrics_label.clone()) — a cheap SharedString clone (Arc-style refcount bump), not the current route.path_match.value().to_owned() + SharedString::from(), which allocates a fresh String on every single matching request, forever, for a value that's identical every time that route matches.

filter/ext-proc/src/tests.rs
twghu, line 1: Same comment re placement of field in the common.rs

filter/src/builtins/http/traffic_management/circuit_breaker/mod.rs
twghu, line 1: One real thing worth checking here though: *is `self.cluster.clone()` on SharedString actually cheap?* state.rs:83 builds it via `SharedString::from(cluster.to_owned())` — same `.to_owned()` allocation pattern flagged earlier for router/mod.rs:395 and the bot's record_lb_panic_mode finding, except here it only runs once at construction (not per-request), so it's much less costly — construction happens once per cluster per filter build, not once per request. Not worth a finding on its own, but it's the same shape of issue, just at a low-frequency call site so it doesn't rise to the same severity.

filter/src/builtins/http/traffic_management/circuit_breaker/state.rs
twghu, line 1: The definition of =pub(super) fn new(cluster: &str, threshold: u32, recovery_window_secs: u64, half_open_timeout_secs: u64) -> Self= would be more readable if the name first param was "=cluster_name=".  As it stands, it is confusing and open to interpretation as an actual cluster (Line 42 and 76).

Would =set_circuit_breaker_open()= be better named with a =_guage= in the name?

Line 220 =crate::metrics::set_circuit_breaker_open(self.cluster.clone(), false);=
The same problem as before, =cluster_name= is more insightful.    Also relates to =src/metrics.rs=.

The method might be better named =set_circuit_breaker_state= for consistency and
extension of states.

The following is not in scope of PR but should be followed up:

This predates the PR (the validation and the state machine here are unchanged by this diff — the only new code is the gauge-publish wrapping), so it's out of scope for this review, not something to block #825 on. But it's a real gap worth a follow-up issue if you want to flag it separately.

There is an unenforced invariant (four independent pieces of state that have to be kept in sync by convention, not by the type system) :

— an enum like =Open { opened_at: Instant }= would make the invalid state unrepresentable, but that's a larger pre-existing structural choice, not something this PR touches. - Concurrency: advance_circuit_state only ever runs with =self.inner='s mutex held (verified in check()), so the "exactly one caller gets the Open→HalfOpen transition and thus the probe" behavior is race-free — the mutex serializes it.

Where I'd actually put a sanity check, though: half_open_timeout_secs isn't validated to be nonzero. mod.rs:97-110 explicitly rejects
consecutive_failures == 0 and recovery_window_secs == 0 at from_config, but there's no equivalent check for half_open_timeout_secs (config.rs:76,
plain u64 with a default = 30, not a try_from-constrained type — so an operator can explicitly write half_open_timeout_secs: 0). If that happens:
in the HalfOpen branch, half_opened_at.elapsed() >= half_open_timeout becomes elapsed >= 0, which is always true. Any second request landing in
HalfOpen — even microseconds after the probe was admitted — immediately snaps inner.state back to Open. Since record_success/record_failure match on inner.state directly rather than on some token tied to "this specific probe," the real probe's eventual success arrives to find state == Open
and hits the Open => {} no-op branch — its success is silently dropped. Under any concurrent traffic to that cluster, the circuit could never
actually close.

filter/src/builtins/http/traffic_management/circuit_breaker/tests.rs
twghu, line 1: Checked metrics_circuit_breaker_open_gauge (prometheus_metrics.rs:265) against what CircuitBreaker's new Drop impl (state.rs:216) actually does, and there's a coverage gap worth closing.

The only circuit-breaker gauge test trips the breaker and asserts praxis_circuit_breaker_open{cluster="backend"} 1. Two things aren't exercised anywhere, unit or integration:

  1. The gauge resetting to 0 on recovery (HalfOpen → Closed) — publish_open_gauge(Closed) is reachable from record_success, but no test scrapes for praxis_circuit_breaker_open{cluster="backend"} 0 after a successful probe.
  2. The Drop-triggered cleanup on hot reload — this is the one I'd actually block on. The PR description calls out "CB gauge cleanup on drop" as a deliberate feature, and it's real: state.rs:216-222 clears the gauge specifically so a cluster removed from config doesn't leave a stale open=1 series behind after reload. metrics_config_reload_success_and_failure (prometheus_metrics.rs:425) reloads config and checks praxis_config_reload_total, but never removes a circuit-breaker cluster in the process — so the actual behavior the Drop impl exists for has no test proving it works. If that cleanup path ever regressed (e.g. the reload path swaps the breaker map in a way that doesn't actually drop the old entries), nothing here would catch it.

  Suggest one integration test: start a proxy with a circuit_breaker filter configured for a cluster, trip it (gauge should read 1), reload with
  that cluster/filter removed from config, and assert the series is gone from the scrape (or reset to 0, whichever the intended contract is — worth
  being explicit about which, since "removed filter" and "cluster still present but not tripped" are different guarantees).

filter/src/builtins/http/traffic_management/load_balancer/mod.rs
twghu, line 1: The diff itself: the switch from ctx.cluster.as_deref() (→ Option<&str>) to ctx.cluster.as_ref() + cluster.as_ref() (→ keeps &Arc<str> around alongside the derived &str) exists solely so record_lb_panic_mode can be called with SharedString::from(Arc::clone(cluster)) — a refcount bump, not a heap allocation. The warn! log on the line above is pre-existing (unchanged in the diff) and already runs per-request for as long as all endpoints stay unhealthy; the new metrics call mirrors that same per-request cadence, so praxis_lb_panic_mode_total counts "requests served under panic mode," not "panic-mode episodes" — consistent with the existing log behavior it sits next to, not a new judgment call. No issues here.

The bigger thing this surfaced: this is exactly the caller-side fix for the bot's Medium finding on record_lb_panic_mode from the stale review (commit 8a1e7ee) — and checking further, I think that whole review is now stale. PR #825 currently has a single commit (309556f, matching what we're reviewing locally) — the branch history was rewritten since the bot ran. I checked all three of its findings against current HEAD:

  1. Histogram buckets (protocol/src/http/pingora/metrics.rs:123-133) — set_buckets_for_metric() with BODY_SIZE_BUCKETS_BYTES is now wired into
  install_prometheus_recorder(). Fixed.
  2. SharedString allocation (filter/src/metrics.rs) — record_lb_panic_mode now takes SharedString directly, and this file's caller passes it via
  cheap Arc::clone. Fixed.
  3. Missing retry-success test — metrics_upstream_retry_success_after_transient_connect_failure exists in prometheus_metrics.rs:188, asserting
  praxis_upstream_retries_total{cluster="backend",result="success"} via a real transient-then-recovered backend. Fixed.

filter/src/builtins/http/traffic_management/router/mod.rs
twghu, line 1: Two functional/test items and two smaller notes from going through the diff file-by-file.

  1. Circuit-breaker Drop cleanup has no test proving it works

  CircuitBreaker's new Drop impl (state.rs:216-222) clears praxis_circuit_breaker_open on drop specifically so a cluster removed from config doesn't leave a stale open=1 series behind after a hot reload — called out in the PR description as deliberate. But metrics_circuit_breaker_open_gauge (prometheus_metrics.rs:265) only asserts the gauge hits 1 after tripping, and metrics_config_reload_success_and_failure (prometheus_metrics.rs:425) reloads config but never removes a circuit-breaker cluster in the process. So the one behavior this Drop impl exists for has zero coverage — if the reload path ever stopped actually dropping old breaker entries, nothing here would catch it. Suggest an integration test: trip a breaker, reload with that cluster/filter removed, assert the series is gone (or reset — worth being explicit about which contract is intended). Would also be good to assert the gauge goes back to 0 on HalfOpen → Closed recovery, which isn't covered either.

  2. router/mod.rs:395 allocates a new String on every matching request

  ctx.metrics_route = Some(::metrics::SharedString::from(route.path_match.value().to_owned()));

  route.path_match.value() is static per configured route — it never changes after resolve_routes() builds the route table at filter construction. .to_owned() here re-allocates that same string on every single request that matches the route, forever. ResolvedRoute (router/mod.rs:113) already precomputes one derived value at construction for exactly this reason (wildcard_suffix); suggest adding a metrics_label: ::metrics::SharedString field there, computed once in resolve_routes(), and cloning it (cheap, Arc-backed) in on_request instead. Same shape as the metrics_cluster_shared construction two lines below cluster selection (request_filter/mod.rs:240, SharedString::from(Arc::clone(c))) — this is the one call site in the PR that didn't get the cheap-clone treatment.

  3. Minor: metrics_route field placed out of order

  filter/src/context.rs:113 — the surrounding block (body_done_indices → peer_identity) is alphabetical; metrics_route was inserted after cluster but alphabetically belongs after downstream_tls, before peer_identity.

  4. Minor: exact and prefix routes produce identical route labels

  PathMatch::value() (core/src/config/route.rs:108) returns the same string for Exact and Prefix variants. A path: /api/v1 exact route and a path_prefix: /api/v1 prefix route emit the identical route="/api/v1" label despite matching different traffic — worth a * suffix or similar on

filter/src/builtins/tcp/traffic_management/tcp_load_balancer.rs
twghu, line 1: Structurally identical to load_balancer/mod.rs — same as_ref() + Arc::clone(cluster) refactor, correctly applying the cheap-SharedString pattern for record_lb_panic_mode. No issues with the mechanics; it mirrors the HTTP-side fix exactly.

One gap, though: the new unit test in this file, panic_mode_still_routes_when_all_unhealthy (line 295), only asserts ctx.upstream_addr.is_some() — it tests that panic-mode routing still selects an endpoint, which is pre-existing behavior unrelated to this PR. It never asserts anything about record_lb_panic_mode actually firing. That would pass identically whether or not the new metrics line (tcp_load_balancer.rs:147) exists at all.

And the integration side doesn't cover it either: metrics_lb_panic_mode_when_all_unhealthy (prometheus_metrics.rs:361) only exercises the HTTP
path — router + load_balancer filters over an HTTP listener. There's no TCP-listener equivalent. So record_lb_panic_mode has two independent call sites (HTTP load_balancer/mod.rs:147, TCP tcp_load_balancer.rs:147) feeding the same counter, but only one of them is verified anywhere — unit or integration. If the TCP call site were ever dropped, malformed, or diverged in its label, nothing in this PR's test suite would catch ito.

Same category as the circuit-breaker Drop gap from before — worth adding to the same review comment: either extend the TCP unit test to assert the counter incremented (would need a lightweight metrics assertion, which — as we found earlier — this codebase doesn't have a pattern for at the unit level), or add a TCP-listener integration test alongside the existing HTTP one.

filter/src/context.rs
twghu, line 1: Finding — context.rs:113, field out of alphabetical order. The surrounding block (body_done_indices, branch_iterations, client_addr, cluster,
  current_filter_id, downstream_tls, peer_identity) is strictly alphabetical, per CLAUDE.md's "Struct fields: alphabetical" convention.
  metrics_route was inserted right after cluster, but alphabetically it belongs between downstream_tls and peer_identity. Trivial fix — move the
  field down four lines. Every call site that constructs this struct literal (benchmarks/microbenchmarks/common.rs, filter/src/lib.rs:402,
  protocol/src/http/pingora/context.rs:469, and the macro-generated copy at context.rs:285) currently mirrors this same out-of-order position, so
  they'd all move together, but none of them break either way since struct literals aren't order-sensitive in Rust — purely a readability/convention
  nit, Minor severity.

filter/src/lib.rs
twghu, line 1: Same out-of-order position as context.rs:113 (inserted after cluster, before current_filter_id), which will self-correct if that field gets moved.

protocol/src/http/pingora/handler/no_body.rs
twghu, line 1: praxis_connections_active measures in-flight requests, not open connections, under HTTP/1.1 keep-alive

  ActiveConnectionGuard::acquire (with_body.rs:149) runs inside early_request_filter, which fires via new_ctx(). Checked against the vendored
  Pingora fork directly (pingora-proxy/src/lib.rs:1170): new_ctx() runs once per request, including requests served over a reused keep-alive
  connection — the surrounding code explicitly threads prev_user_ctx through on_connection_reuse to let implementations carry state across requests
  on the same connection. Praxis doesn't implement that hook (uses Pingora's no-op default at proxy_trait.rs:500-506), so nothing carries the
  ActiveConnectionGuard forward.

  Practical effect: on a persistent HTTP/1.1 connection serving N sequential keep-alive requests, each request gets its own fresh ctx, acquires its
  own guard (gauge +1), and releases it when that single request's ctx drops at end-of-request (gauge −1). The gauge blips once per request rather
  than holding steady at 1 for the connection's lifetime. Since keep-alive is the common case for real HTTP/1.1 traffic, this metric ends up
  tracking something close to "concurrent in-flight requests" — largely redundant with what a praxis_http_requests_total rate already shows — rather
  than "concurrent open sockets," which is the saturation signal the original design doc justified it against (Envoy's downstream_cx_active, which
  is genuinely connection-scoped and stays incremented regardless of request pacing on the connection).

  Not blocking — the counter/gauge codef is correct, this is a semantics gap, not a bug in the code as written. But worth deciding
  explicitly: either (a) document that connections_active means "in-flight requests" and rename/relabel accordingly so it isn't read as a
  socket-level signal, or (b) move the guard acquire/release to wherever Pingora exposes true connection-level lifecycle (if it does — may not be
  reachable from ProxyHttp at all, in which case (a) is probably the honest answer).

protocol/src/http/pingora/handler/with_body.rs
twghu, line 1: Unrelated to this PR, flagging since I was in this file: PingoraHttpHandlerNoBody appears to be dead code.

=load_http_handler (handler/mod.rs:106)= unconditionally constructs
PingoraHttpHandler (with_body.rs), with an explicit comment explaining why:
"Always use the body-capable handler: a reload may add body filters, and
compression init is one-shot in Pingora."

I couldn't find any call site anywhere in the repo that constructs PingoraHttpHandlerNoBody::new — it's pub use'd from handler/mod.rs:44 and has its own doc comment describing an intended use case ("used when no filter in the pipeline declares body access"), but nothing currently selects it.

This PR duplicates the new metrics wiring (ActiveConnectionGuard, record_overload_reject, connected_to_upstream) into this file to keep it in sync
with with_body.rs, which is reasonable if the intent is to keep it revivable — but if it's actually been superseded and isn't coming back, it's
~280 lines of parallel ProxyHttp implementation with no test coverage proving it still works (no construction site means nothing exercises it
end-to-end). Worth a separate follow-up to decide: keep it maintained as a real second code path, or remove it.

protocol/src/http/pingora/handler/upstream_peer.rs
twghu, line 1: upstream_peer.rs:39-41 introduces a hardcoded 50ms retry-backoff sleep — this looks like scope creep from a test-timing need, not an intentional
  feature of this PR

  pub(super) async fn execute(ctx: &mut PingoraRequestCtx) -> Result<Box<HttpPeer>> {
      if ctx.retries > 0 {
          tokio::time::sleep(Duration::from_millis(50)).await;
      }
      ctx.upstream_connect_start = Some(Instant::now());
      ...

The upstream_connect_start timestamp is the legitimate metrics addition here — it's what feeds praxis_upstream_connect_duration_seconds via connected_to_upstream, and it's correctly placed after the sleep so the artificial delay doesn't pollute that histogram.

The sleep above it is a separate, real change to retry behavior: every idempotent request that fails to connect once now waits an extra 50ms before its retry, unconditionally, in every deployment. Tracing why it's here — the new integration test metrics_upstream_retry_success_after_transient_connect_failure (prometheus_metrics.rs:188) has a comment reading "a later retry (with the short reconnect pause) can succeed," which suggests this exists to make that test's timing deterministic (giving a backend thread time to bind before the retry fires) rather than being something #794/#822 asked for.

A few things make this stand out as unintentional scope creep rather than a deliberate design decision:

  - No config control. Every other timing tunable touched by this PR's neighborhood (recovery_window_secs, half_open_timeout_secs) is per-cluster
  config. This is a bare literal.
  - No test asserts the delay itself. Nothing verifies retries actually wait ~50ms — it's only implicitly required for one unrelated test's timing
  to pass reliably.
  - Not mentioned in the PR description, either in the summary or the "explicit follow-ups" list, despite being a change to request-handling
  latency, not to observability.
  - It adds a real latency floor to production retry paths as a side effect of a metrics PR — a behavior change like this should get its own PR, its
  own tests for the backoff specifically, and ideally a config option (even a fixed internal constant should at least be named and documented with
  the actual rationale, not just "so a transient outage can recover without burning the retry budget").

Suggest either pulling this out into its own PR with proper scoping, or — if the test genuinely can't be made deterministic without it — using a test-only synchronization primitve (e.g. have the test wait on a signal that the backend is bound, rather than the proxy code guessing at a delay) instead of changing production behavior.

protocol/src/http/pingora/metrics.rs
twghu, line 1: The record_helpers_noop_without_recorder test-order hazard (shared OnceLock racing against four sibling tests that call install_prometheus_recorder()), and the pub vs pub(crate) visibility nit on the label constants. Neither depended on how reload.rs calls into this file — both stand on their own.

1. record_helpers_noop_without_recorder doesn't reliably test what it claims to. This test exists to prove the guarded functions don't panic when install_prometheus_recorder() was never called. But PROMETHEUS_HANDLE: OnceLock<PrometheusHandle> is a single process-wide static with no reset mechanism, and four other new tests in this same file — overload_reject_reasons_appear_in_scrape, body_size_histograms_use_byte_buckets, clear_stale_upstream_health_gauges_zeros_removed_clusters, seed_upstream_health_gauges_publishes_registry_counts — all call install_prometheus_recorder() themselves. Checked main before this PR: none of the existing tests in this file ever called it, so this is a hazard this PR introduces fresh, not one it inherited. cargo test runs tests within a binary in parallel with no ordering guarantee, so whether record_helpers_noop_without_recorder actually exercises the "recorder absent" path depends on whether it happens to run before any of those four siblings — not guaranteed, not detectable from the test result either way (it "passes" regardless of which case it hit). Suggest moving this assertion to its own test binary/file so it gets a guaranteed-unset OnceLock, or restructuring so the recorder-present tests share one installation and this one is verified some other way.

NOTE: We can skip this.
2. Minor: label constants are pub, broader than they need to be. OVERLOAD_REASON_*, RETRY_RESULT_*, HEALTH_RESULT_*, RELOAD_RESULT_* are declared pub, but every call site referencing them (no_body.rs, with_body.rs, handler/mod.rs, runner.rs) is within the protocol crate. Every function in this same file that's internal-only is correctly pub(crate) — these constants are the one place that doesn't match. pub(crate) would be tighter and consistent with the rest of the file.

Positive: clear_stale_upstream_health_gauges + seed_upstream_health_gauges correctly solve a real Prometheus gotcha (stale series lingering after a cluster is removed from config) — traced the full call path through reload.rs and confirmed the argument order matches the function signature and the registry is freshly built from the new config before either is called. Good design, correctly wired.

protocol/src/tcp/proxy.rs
twghu, line 1: praxis_connections_active{listener="..."} misattributes traffic for grouped TCP listeners — fixable, not an inherent limitation

group_tcp_listeners (tls_setup.rs:29) merges multiple differently-named listeners sharing the same (upstream, cluster, timeout_ms, max_duration_secs) into a single Service/PingoraTcpProxy. This isn't hypothetical — the file's own test, group_tcp_listeners_groups_by_upstream_and_timeout, sets up exactly this: db1 (:5432) and db2 (:5433) pointed at the same upstream produce one group. validate_tcp_group_consistency's doc comment already acknowledges the sharing directly and validates that filter_chains/max_connections stay consistent across it — but nothing validates or accounts for listener name, which is always distinct by definition.

mod.rs:67-70 derives listener_name from listeners.first() once per group and passes it into PingoraTcpProxy::new, which caches it on the struct (listener_name: ::metrics::SharedString, proxy.rs:72). Every connection handled by that PingoraTcpProxy then reports the first listener's name via ActiveConnectionGuard::acquire(self.listener_name.clone()) (proxy.rs:262ish) — regardless of which of the group's actual bind addresses the connection came in on. For the db1/db2 example, all traffic on :5433 (db2) would silently show up as listener="db1" in the gauge.

The reason this is worth fixing rather than just documenting: the data needed to disambiguate is already available. proxy.rs:237 extracts local_addr (the real per-connection local socket address) via extract_addrs(&session), in the same function that later uses self.listener_name.clone() for the metric. So the fix is contained:

  - In mod.rs, instead of computing a single listener_name from listeners.first(), build a small map from each grouped listener's bind address to
  its SharedString name (group sizes are just the number of distinct addresses sharing one upstream — typically 1-3) and pass that into
  PingoraTcpProxy::new instead of one value.
  - In proxy.rs, look up local_addr in that map at connection time instead of cloning the cached self.listener_name.

Not blocking — the metric is still useful for the common case (one listener per group), and this doesn't affect correctness of anything except the listener label on grouped setups. But worth a fix or, at minimum, a doc-comment caveat on CONNECTIONS_ACTIVE (metrics.rs) so it doesn't quietly mislead anyone running grouped TCP listeners.

protocol/src/tcp/mod.rs
twghu, line 1: See comment re proxy.rs

tests/integration/tests/suite/prometheus_metrics.rs
twghu, line 1: Misattached /// doc comment in prometheus_metrics.rs:155-160

  /// Memory and global-connection overload reasons share the same
  /// `early_request_filter` call sites as `listener_connections`, but cannot
  /// be exercised safely end-to-end in this suite: both use process-wide
  /// `OnceLock` limits that would poison later tests. Label emission for
  /// `reason=memory` and `reason=global_connections` is covered by
  /// `overload_reject_reasons_appear_in_scrape` in `protocol` metrics unit tests.

  #[test]
  fn metrics_upstream_connect_failure_and_retry_exhausted() {

  This /// doc comment attaches to whatever item follows it in Rust regardless of the blank line — here, that's
  metrics_upstream_connect_failure_and_retry_exhausted, a test about connect failures and retry exhaustion. But the comment's content is about why
  memory/global_connections overload reasons aren't tested end-to-end — it clearly belongs to the preceding test,
  metrics_overload_rejects_listener_connections (the listener_connections case it's contrasting against), not the one it's actually attached to.
  cargo doc or an IDE hover tooltip would show this explanation on the wrong function.

  The reasoning itself holds up — praxis_core::memory::is_exceeded() and crate::connections::try_acquire_global()'s global OnceLock-backed semaphore really are process-wide, unresettable state that would poison every test running afterward in the same binary, so the decision to cover those two reasons at the unit level (overload_reject_reasons_appear_in_scrape in protocol's metrics.rs) instead is sound. Just needs to move (or become a plain // comment) so it's not misread as documenting an unrelated test.

twghu (commented):
I saw the PR had been closed due to How/Why etc note on the PR 822 but as I had reviewed the code in this PR it seemed sensible to include the review for any pending adjustment/rework from PR 822.

filter/ext-proc/src/tests.rs
henschwartz: Done. Same field-order fix as the other struct literals.

benchmarks/microbenchmarks/common.rs
henschwartz: Done. Precomputed metrics_label on ResolvedRoute (clone on match). metrics_route kept on request ctx and ordered alphabetically.

filter/src/builtins/http/traffic_management/circuit_breaker/mod.rs
henschwartz: Acknowledged. Construction-time SharedString::from(...to_owned()) left as is (once per cluster build, not per request).

filter/src/builtins/http/traffic_management/circuit_breaker/state.rs
henschwartz: Done for naming: cluster_name + set_circuit_breaker_state. half_open_timeout_secs: 0 left as a follow-up (out of scope here).

filter/src/builtins/http/traffic_management/circuit_breaker/tests.rs
henschwartz: Done. Added scrape coverage for gauge → 0 on recovery and Drop cleanup after reload removes the breaker.

filter/src/builtins/http/traffic_management/load_balancer/mod.rs
henschwartz: Done. Cheap SharedString via Arc::clone. bot findings addressed on current HEAD.

filter/src/context.rs
henschwartz: done

filter/src/lib.rs
henschwartz: done

protocol/src/http/pingora/handler/no_body.rs
henschwartz: done, chose a documented as in-flight HTTP requests

protocol/src/http/pingora/handler/upstream_peer.rs
henschwartz: Kept upstream_connect_start metrics wiring.
Removed production 50ms sleep.
Retry-success test uses a test-only sync gate instead.

protocol/src/http/pingora/metrics.rs
henschwartz: Noop-without-recorder race skipped.
pub(crate) is done.

protocol/src/tcp/proxy.rs
henschwartz: done, per-connection local-addr → listener_name map for grouped TCP listeners.

protocol/src/tcp/mod.rs
henschwartz: Same grouped-listener label fix.

tests/integration/tests/suite/prometheus_metrics.rs
henschwartz: done, changed to plain // on the overload test

shaneutt:
@henschwartz, this needs conflict resolution when you get a second please :vulcan_salute:

protocol/src/http/pingora/metrics.rs
```diff
@@ -120,41 +184,234 @@ pub fn method_label(method: &str) -> &'static str {
...
         "status_class" => labels.status_class,
-        "route" => labels.route,
+        "route" => route,
         "cluster" => cluster
     )
     .record(duration_secs);
 }
 
+/// Record HTTP request and response body size histograms.
+pub(crate) fn record_body_size_metrics(
```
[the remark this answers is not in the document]
henschwartz: Done — configured byte/duration buckets via PrometheusBuilder::set_buckets_for_metric in install_prometheus_recorder()

filter/src/metrics.rs
```diff
@@ -43,3 +49,15 @@ pub(crate) fn record_filter_duration(
...
+/// Set the circuit-breaker open gauge for a cluster.
+///
+/// Half-open is treated as open (`1.0`).
+pub(crate) fn set_circuit_breaker_open(cluster: SharedString, open: bool) {
+    gauge!(CIRCUIT_BREAKER_OPEN, "cluster" => cluster).set(if open { 1.0 } else { 0.0 });
+}
+
+/// Increment the load-balancer panic-mode counter for a cluster.
+pub(crate) fn record_lb_panic_mode(cluster: &str) {
+    counter!(LB_PANIC_MODE_TOTAL, "cluster" => SharedString::from(cluster.to_owned())).increment(1);
```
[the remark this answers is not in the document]
henschwartz: Done — takes SharedString; HTTP/TCP LB callers pass Arc-cloned labels.

tests/integration/tests/suite/prometheus_metrics.rs
```diff
@@ -0,0 +1,428 @@
...
+    let (status, _) = http_get(proxy.addr(), "/api/", None);
+    assert_eq!(status, 502, "dead upstream should fail the request");
+
+    let metrics = scrape_metrics(&admin);
+    assert!(
+        metrics.contains("praxis_upstream_connect_failures_total{cluster=\"backend\"}"),
+        "expected connect failure counter:\n{metrics}"
+    );
+    assert!(
+        metrics.contains("praxis_upstream_retries_total{cluster=\"backend\",result=\"exhausted\"}"),
```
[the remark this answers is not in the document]
henschwartz: Done — added scrape test with a test-only retry gate (no production sleep)

filter/src/builtins/http/traffic_management/router/mod.rs
henschwartz: Done —
1) Added integration coverage for CB gauge recovery to 0 and Drop cleanup after reload removes the breaker.
2) Precomputed ResolvedRoute.metrics_label; on_request clones the SharedString (no per-request String alloc).
3) metrics_route field order fixed.
4) Exact uses bare path; Prefix appends `*` so match kinds stay distinguishable.

filter/src/builtins/tcp/traffic_management/tcp_load_balancer.rs
henschwartz: Done — added TCP scrape coverage (`metrics_tcp_lb_panic_mode_when_all_unhealthy`) so the TCP call site of `record_lb_panic_mode` is verified, not only the HTTP path.

protocol/src/http/pingora/handler/with_body.rs
henschwartz: Done — removed the unused `PingoraHttpHandlerNoBody` path. `load_http_handler` already always constructs the body-capable handler, so keeping a parallel untested ProxyHttp impl wasn’t worth it.

henschwartz:
@shaneutt ready for review :)

protocol/src/http/pingora/handler/mod.rs
```diff
@@ -240,38 +242,59 @@ fn adjust_compression(
 /// Body-mutating filters can change payload length after `request_filter`,
 /// so the retry guard uses the larger of the original and mutated lengths.
 fn handle_connect_failure(ctx: &mut PingoraRequestCtx, e: Box<pingora_core::Error>) -> Box<pingora_core::Error> {
```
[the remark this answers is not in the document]
henschwartz: Resolved handle_connect_failure` now records `praxis_upstream_connect_duration_seconds` from `upstream_connect_start` before incrementing the failure counter, so failed/timed-out connects are included in the histogram.

shaneutt (approved):

Issue #317, opened 2026-05-15
Access log trace ID correlation


shaneutt:
Inject trace ID into structured JSON access log entries for log-to-trace correlation.

## Requirements

- Include `trace_id` field in JSON access log entries
- Include `span_id` field for the root request span
- Trace ID format: 32-character hex string (W3C format)
- Field omitted when tracing is not enabled
- Correlate with OTLP exported traces

## Implementation Notes

- Trace ID sourced from current span context
- Enables log-to-trace correlation in observability backends

Parent epic: #10

Ladas:
Implemented in b988e6b (feat(filter): access log trace ID correlation), pushed directly to main as part of bulk merge.

Issue #313, opened 2026-05-15
TCP connection lifecycle spans


shaneutt:
Spans for TCP proxy connection lifecycle including byte counts and duration.

## Requirements

- Span per TCP connection lifecycle
- Attributes: client address, upstream address, bytes sent, bytes received
- Attributes: connection duration, TLS handshake duration (if applicable)
- Span events for connection open, close, error
- Support for TcpFilter pipeline tracing

## Implementation Notes

- TCP spans are independent root spans (not children of HTTP spans)
- Byte counts updated at connection close

Parent epic: #10

Ladas:
Implemented in 16f0c62 (feat(protocol,filter): tracing spans (TCP, upstream exchange, OTel semconv, span events, per-filter)), pushed directly to main as part of bulk merge.

Issue #311, opened 2026-05-15
Configurable head-based sampling with always-sample-on-error


shaneutt:
Configurable trace sampling rate with always-sample-on-error override.

## Requirements

- Head-based sampling with configurable rate (0.0 to 1.0)
- Default: sample 1% of requests (0.01)
- Always-sample-on-error: override sampling decision if request results in error
- Respect incoming sampling decision from traceparent
- Parent-based sampling: inherit parent's sampling decision

## Implementation Notes

- Use OpenTelemetry `Sampler` trait
- Composite sampler: rate-based + error-based override
- Error detection at response phase (5xx status codes)

Parent epic: #10

Issue #309, opened 2026-05-15
Span events for rejection, timeout, retry, error


shaneutt:
Record span events at key decision points in request processing.

## Requirements

- Event on request rejection (with reason, rejecting filter)
- Event on upstream timeout (with timeout duration, upstream address)
- Event on retry attempt (with attempt number, reason)
- Event on error (with error type, message)
- Event on circuit breaker trip

## Implementation Notes

- Events are lightweight annotations on existing spans
- Include relevant context as event attributes

Parent epic: #10

Ladas:
Implemented in 16f0c62 (feat(protocol,filter): tracing spans (TCP, upstream exchange, OTel semconv, span events, per-filter)), pushed directly to main as part of bulk merge.

PR #836, opened 2026-07-22
feat(core): wire tracing subscriber with fmt + OTLP layers

### LINKED ISSUE #315: Wire tracing subscriber: fmt + OTLP layers
Compose tracing subscriber with both fmt (console/file logging) and OTLP (trace export) layers sharing context.

## Requirements

- Single `tracing` subscriber combining fmt and OTLP layers
- Shared span context between logging and tracing
- Trace ID available in log output (fmt layer)
- Configurable: OTLP layer opt-in, fmt layer always present
- Graceful initialization: OTLP layer skipped if not configured

## Implementation Notes

- Use `tracing-subscriber` `Registry` with layered approach
- `tracing-opentelemetry` for OTLP layer
- Replace existing fmt-only subscriber setup

Parent epic: #10

Ladas:
## What

Replace the fmt-only tracing subscriber with a layered `tracing_subscriber::Registry` combining:

- **fmt layer** (always) — stdout text or JSON logging
- **OTLP layer** (opt-in) — span export to OTel Collector via gRPC

## Why

This is the foundational issue for all Praxis observability work. Issues #299, #301, #296, #305, #307, #309, #311, #313, #317 and all AI observability (ai#92, ai#141) are blocked on this.

The `tracing` crate was already used throughout for structured logging. This PR adds an optional OTLP export layer behind the `otel` cargo feature flag, so the default binary is unchanged — zero new dependencies without the feature.

## How

- Add `opentelemetry` 0.32, `opentelemetry_sdk` 0.32, `opentelemetry-otlp` 0.32, `tracing-opentelemetry` 0.33 as workspace dependencies
- Add `registry` feature to `tracing-subscriber`
- Add `otel` feature to `core` and `server` crates (opt-in, not default)
- New `TelemetryConfig` with `otlp_endpoint` field (`deny_unknown_fields`); `OTEL_EXPORTER_OTLP_ENDPOINT` env var as fallback
- Refactor `init_tracing` to use `Registry` + layers, return `TracingGuard` (RAII shutdown)
- Add example config `examples/configs/observability/tracing-otlp.yaml`
- Add feature-gated integration test

**Verified:** tonic 0.14.6 unifies with Pingora (single version in lockfile). Confirmed by [Pingap](https://github.com/vicanso/pingap) (production Pingora proxy using same OTel versions).

## Review Response

| Finding | Response |
|---------|----------|
| Hoist `otel_layer` above `if json` branch | **Won't fix** — does not compile. `OpenTelemetryLayer<S>` requires `S` to match the composed subscriber type. JSON and text fmt produce different `S` types, preventing a shared binding. Added comment explaining the constraint. |
| CI coverage for `otel` feature | **Acknowledged** — the `otel` feature-gated code is not tested in CI (same gap as `ext-proc` after its removal from default features). Follow-up PR will add a CI job running `cargo test --features otel` and `cargo clippy --features otel`. |
| Env var test race condition | **Fixed** — `TelemetryConfig::resolve()` snapshots env var once at startup. Tests use `TelemetryConfig::resolved()` with explicit parameters — zero unsafe, zero env var manipulation. |

## Follow-up

- CI job for `--features otel` testing (addresses reviewer finding on CI coverage)
- Feature flag design discussion: the issue acceptance criteria describe runtime opt-in ("OTLP layer skipped if not configured"), not compile-time gating. The feature flag adds complexity (two code paths, CI gap) but keeps the default binary lean. Open to removing the feature flag in favor of always-compiled `Option<Layer>` if maintainers prefer.

Closes #315

### DIFF
--- a/core/src/config/telemetry.rs
+++ b/core/src/config/telemetry.rs
@@ -6,6 +6,13 @@
 use serde::{Deserialize, Serialize};
 
 // -----------------------------------------------------------------------------
+// Constants
+// -----------------------------------------------------------------------------
+
+/// OTel-standard environment variable for the OTLP exporter endpoint.
+const OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
+
+// -----------------------------------------------------------------------------
 // TelemetryConfig
 // -----------------------------------------------------------------------------
 
@@ -41,14 +48,26 @@ pub struct TelemetryConfig {
 }
 
 impl TelemetryConfig {
-    /// Resolve the effective OTLP endpoint from config or environment.
+    /// Snapshot telemetry settings by merging config with environment.
     ///
-    /// Returns the config value if set, otherwise checks
-    /// `OTEL_EXPORTER_OTLP_ENDPOINT`. Returns `None` if neither is set.
-    pub(crate) fn resolved_otlp_endpoint(&self) -> Option<String> {
-        self.otlp_endpoint
-            .clone()
-            .or_else(|| std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT").ok())
+    /// Config values take precedence over environment variables.
+    /// Call once at startup — the returned config should be stored
+    /// rather than re-evaluated per request.
+    pub(crate) fn resolve(&self) -> Self {
+        Self {
+            otlp_endpoint: self
+                .otlp_endpoint
+                .clone()
+                .or_else(|| std::env::var(OTLP_ENDPOINT_ENV_VAR).ok()),
+        }
+    }
+
+    /// Build from explicit values (for testing without env var mutation).
+    #[cfg(test)]
+    fn resolved(config_endpoint: Option<&str>, env_endpoint: Option<&str>) -> Self {
+        Self {
+            otlp_endpoint: config_endpoint.or(env_endpoint).map(ToOwned::to_owned),
+        }
     }
 }
 
@@ -62,8 +81,7 @@ impl TelemetryConfig {
     clippy::unwrap_used,
     clippy::expect_used,
     clippy::indexing_slicing,
-    unsafe_code,
-    reason = "tests use unwrap/expect/indexing and unsafe env var manipulation"
+    reason = "tests use unwrap/expect/indexing for brevity"
 )]
 mod tests {
     use super::*;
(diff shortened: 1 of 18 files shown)


core/src/config/telemetry.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl TelemetryConfig {
+    /// Resolve the effective OTLP endpoint from config or environment.
+    ///
+    /// Returns the config value if set, otherwise checks
+    /// `OTEL_EXPORTER_OTLP_ENDPOINT`. Returns `None` if neither is set.
+    pub(crate) fn resolved_otlp_endpoint(&self) -> Option<String> {
+        self.otlp_endpoint
+            .clone()
+            .or_else(|| std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT").ok())
```
alexsnaps: Could we have this be resolved once, preferably as soon as possible?
Somewhat of an issue I have with environment variables, is accessing them "over time". Or is it considered "a feature" to be able to change them and trigger a reload of the config for the new props to be applied? Not a big fan in "global mutable state" in general... which is what environment vars effectively are 😭 
I find it easier to reason about if they all get "snapshot"'ed upon process start up, thoughts?

core/src/config/telemetry.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    #[test]
+    fn resolved_endpoint_env_var_fallback_and_unset() {
+        let telemetry = TelemetryConfig { otlp_endpoint: None };
+
+        // SAFETY: env var mutation is not thread-safe. This test is the only
+        // one touching OTEL_EXPORTER_OTLP_ENDPOINT, consolidated into a single
+        // function to avoid races with parallel test execution.
+        unsafe {
```
alexsnaps: So, the issue is that this "safety" guarantee can't be... well, guaranteed. Especially as test runs are multithreaded, it's just a matte of time until this assumption flies out the window. 

I'm generally torn here. Get rid of the test? Add a coordination mechanism around dealing with env variables (e.g. `.with_env_var(FOO, "bar", || assert_something)`? Or... just get rid of the environment variable entirely? Do we really ~want~ need those as a mean to configure praxis?

core/src/config/telemetry.rs
Ladas: yes that makes sense, moved under init_tracing() in second commit I've sent, so it gets resolved one at startup

core/src/config/telemetry.rs
Ladas: ok, looking at other tests workign with env vars, seems like we don't test the env vars themselves (std::env::var() call itself), just testing the precedence of config over env var and then just testing with values we need. That gets rid of needing to mutate env var for test and test can be clean. (no unsafe)

I think we should support also env vars as that is usually a standard way to configure the operator? With config file values taking precedence.

shaneutt (approved):

core/src/config/telemetry.rs
shaneutt: Let's do a follow-up for this.

PR #847, opened 2026-07-22
fix(filter): reconcile security filter classification

### LINKED ISSUE #841: fix(filter): reconcile security filter classification sources
## Summary

Two independent sources of truth disagree about which filters are security-critical:

1. **Hardcoded `SECURITY_FILTERS`** in `filter/src/pipeline/checks.rs` — used by `check_skip_to_bypasses_security` and `check_conditional_security`
2. **`SecurityClass` in the registry** (`filter/src/registry.rs`) — used by `is_security_filter()` and `security_filters()`

Discrepancies:
- `rate_limit`: in `SECURITY_FILTERS` but registered as `SecurityClass::Standard`
- `policy`: registered as `SecurityClass::Security` but absent from `SECURITY_FILTERS`

If future checks use the registry instead of the hardcoded list (or vice versa), these filters would be silently unprotected or over-protected.

## Expected

Use a single source of truth. Either:
- Remove the hardcoded list and derive it from registry metadata, or
- Keep both in sync with a test that asserts equivalence

Also decide: should `rate_limit` be `Security` class? Should `policy` be in `SECURITY_FILTERS`?

## Files

- `filter/src/pipeline/checks.rs` (`SECURITY_FILTERS`)
- `filter/src/registry.rs` (security class registrations)

AnayGarodia:
## Summary
- classify `rate_limit` as security-critical, matching both its protective role and the existing pipeline safeguards
- include feature-gated `policy` in pipeline security checks
- add an equivalence regression test so the hardcoded pipeline list cannot drift from registry metadata again

## Verification
- `cargo test -p praxis-proxy-filter security_filter_list_matches_registry_metadata`
- `cargo test -p praxis-proxy-filter registry::tests::`
- `cargo test -p praxis-proxy-filter --features cpex-policy-engine security_filter_list_matches_registry_metadata`
- `cargo clippy -p praxis-proxy-filter --all-targets --all-features -- -D warnings`
- `cargo +nightly fmt --all -- --check`

Closes #841

### DIFF
--- a/filter/src/pipeline/checks.rs
+++ b/filter/src/pipeline/checks.rs
@@ -20,6 +20,8 @@ const SECURITY_FILTERS: &[&str] = &[
     "forwarded_headers",
     "guardrails",
     "ip_acl",
+    #[cfg(feature = "cpex-policy-engine")]
+    "policy",
     "rate_limit",
 ];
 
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
Good stuff, thank you!

Issue #549, opened 2026-06-10
Quota exhaustion failover


shaneutt:
## What

Automatic traffic shifting when a backend's token quota is exhausted. When a rate limiter returns 429 for a specific backend, route the request to an alternative backend rather than rejecting it.

## Why

In multi-provider deployments, quota exhaustion on one provider shouldn't mean request failure when capacity exists elsewhere. Envoy AI Gateway implements this: exhausted backends cause traffic to shift to alternatives in the route rule it uses this to shift between provisioned throughput and on-demand capacity tiers.

## Goals

- Detect quota exhaustion per backend (429 from rate limiter, not from upstream)
- Re-route to next-priority backend in the route configuration
- Apply appropriate schema translation for the fallback backend
- Metrics for quota exhaustion events per backend

shaneutt:
This can be achieved with existing branch chains. No dedicated mechanism needed.

Issue #826, opened 2026-07-21
core: add a hardened SubRequestClient over SubRequestConnector


leseb:
## Summary

Add a reusable, hardened `SubRequestClient` in `praxis-core` on top of
Pingora's `SubRequestConnector` so filters and API adapters can execute HTTP
sub-requests without rebuilding transport safety and lifecycle handling.

PR #849 removed the reqwest-based `CalloutClient` and introduced the
Pingora-native sub-request path used by `iterative_request_router`:

- a process-shared `SubRequestConnector` with connection pooling and optional
  concurrency admission;
- a bounded, deadline-aware buffered executor;
- typed internal transport errors;
- request/response header sanitization and correct message framing;
- framework-owned iterative-depth propagation; and
- tracing around iterative steps.

The executor and its error types currently remain private to the filter crate.
PR #868 moves the transport-neutral `SubRequest` and `SubResponse` data types
into core while preserving the filter crate's public re-exports. The next step
is to promote the reusable execution boundary into core as a safe client API.

This continues the safety requirements described in #28, particularly bounded
time and memory, cancellation, concurrency limits, and observability.

## Current Gap

`SubRequestConnector` exposes the raw Pingora connector and an admission
semaphore, while the hardened exchange logic lives in
`filter/src/pipeline/subrequest.rs`. A new consumer can therefore use the core
connector and types without automatically receiving the executor's deadline,
body-limit, framing, sanitization, and typed-error guarantees.

Concurrency admission also has a lifecycle mismatch today: runtime
documentation says permit acquisition is bounded by the step timeout, but the
executor awaits the semaphore permit before starting its timeout. Saturated
admission can therefore wait beyond the configured step deadline and has no
explicit overload outcome.

## Proposed Boundary

`praxis-core` should own a `SubRequestClient` that combines the shared
connector with the complete HTTP exchange lifecycle. Callers provide an
already selected/validated target, the request, and per-call limits; the client
enforces transport invariants and returns either a `SubResponse` or a typed
transport error.

The client should not become an application or routing abstraction. JSON
serialization, resource-path construction, target selection, API-specific
header allowlists, HTTP-status policy, domain error mapping, fail-open/fail-
closed behavior, and circuit breaking remain caller or filter responsibilities.

Iterative depth and iteration policy also remain owned by
`iterative_request_router`. Core should prevent arbitrary transport/internal
headers from bypassing its boundary, but should not turn router-specific loop
state into a generic client concern.

## Success Criteria

[...]

PR #849, opened 2026-07-23
iterative_request_router: failover and agentic loop support

### LINKED ISSUE #786: Spike: Agentic Loop Support

shaneutt:
### What does this PR do?

This adds an HTTP client harness from Pingora to be used for sub-request callouts (see https://github.com/orgs/praxis-proxy/discussions/87), and adds a new `iterative_request_router` filter which enables failover and agentic loop facilities.

Fixes #786 

### Related issues

Supports:
- https://github.com/praxis-proxy/praxis/discussions/777 - Pipeline Continuations
- https://github.com/orgs/praxis-proxy/discussions/87 - Sub-requests
- https://github.com/praxis-proxy/praxis/issues/549 - Quota exhaustion failover
- https://github.com/praxis-proxy/praxis/issues/358 - HTTP Callout Filter (now deferred)
- https://github.com/praxis-proxy/ai/discussions/287 - Provider fallback
- https://github.com/praxis-proxy/ai/issues/26 - Agentic loop filter
- https://github.com/praxis-proxy/ai/issues/93 - Responses API and Agentic Loop epic
- https://github.com/praxis-proxy/ai/issues/134 - LLM Provider Failover
- https://github.com/praxis-proxy/ai/issues/425 - MCP tool resolve rewriting


leseb (commented):
This is what my AI tell me based on my work in agentic_loop filter:

Concrete blockers:

- Step pipelines only execute request headers—no request body, response headers, or response body.
- Steps use a built-ins-only registry, so agentic_loop, tool_parse, MCP, and other Praxis AI filters cannot load.
- The cross-step state types are crate-private.
- The existing “agentic loop” test merely routes on HTTP 202; it never passes a tool result back to the model.

I'll pass a test soon that you can include here to valid the behavior i need

leseb:
@shaneutt acceptance tests we can add to validate the behaviour https://gist.github.com/leseb/0b051e3545d93fb73442274b00fed99c

docs/proposals/00358_http-callout-filter.md
```diff
@@ -16,6 +17,17 @@ stakeholders:
 
 # HTTP Callout Filter
 
+> **Status: Deferred.** The sub-request capability this
+> proposal required is now provided by the
+> `iterative_request_router` filter
+> ([proposal 00786](00786_iterative-request-router.md)),
+> which uses Pingora's native Connector for sub-request
+> execution with connection pooling, HTTP/2, and TLS.
+> The reqwest-based `CalloutClient` has been removed.
```
usize, line 26: Callout client isn't used here, but is in the ai project. 

Would it make sense to move the `CalloutClient` removal to its own PR?

We don't have much traffic now, so if you want to keep the removal over here it's fine. We'll just have to follow up with replacement in the AI repo quickly.

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
```diff
@@ -0,0 +1,1067 @@
...
+    /// Transition to the named step.
+    Next(Arc<str>),
+
+    /// No transition matched.
+    NoMatch,
+}
+
+/// Convert transport failures into gateway responses so ordinary status
+/// transitions can drive provider failover.
+fn transport_failure_status(error: &subrequest::SubRequestError) -> u16 {
```
usize: `transport_failure_status` collapses `Connect`, `Io`, and `ResponseTooLarge` into 502 and `DeadlineExceeded` into 504. 

Operators may want to express "fail over only if we never connected." It may be the case, for example, that an operator finds that falling back to the upstream provider doesn't play nice in workflows that have already had a few turns. Not a blocker, but worth calling out as something that could come up and may be worth adding an issue for -- new contributors are onboarding, so it's fine to have some things like this set aside for them maybe? c:

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
```diff
@@ -0,0 +1,1067 @@
...
+    match error {
+        subrequest::SubRequestError::DeadlineExceeded => 504,
+        subrequest::SubRequestError::Connect(_)
+        | subrequest::SubRequestError::Io(_)
+        | subrequest::SubRequestError::ResponseTooLarge { .. } => 502,
+    }
+}
+
+/// Convert a nested filter's local response into transition input.
+fn subresponse_from_rejection(rejection: Rejection) -> SubResponse {
```
usize, line 737: For the quota-exhaustion failover case (#549) a `rate_limit` filter inside a step chain that rejects with 429 will match a `status: [429]` transition and fail over to the fallback provider even though the upstream was never contacted. Is falling back from our own rate limit correct here? It seems iffy to me.

1provider_failover_on_local_step_rejection` exists we could mark the origin (e.g., a reserved result key such as `iterative_request_router.origin = local|upstream` that way we could only fallback on an upstream 429?

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
```diff
@@ -0,0 +1,1067 @@
...
+                    debug!(
+                        from = current_step.as_ref(),
+                        to = next_step.as_ref(),
+                        "transitioning to next step"
+                    );
+                    let next_body = next_iteration_body.map_or_else(|| current_request.body.clone(), |b| b.0);
+                    current_request = SubRequest {
+                        method: current_request.method.clone(),
+                        uri: current_request.uri.clone(),
+                        headers: HeaderMap::new(),
```
usize, line 664: On transition, the next step starts with `HeaderMap::new()` while the *first* step inherits the full client header set. The credential-isolation rationale is sound (and `credentials_do_not_cross_step_boundaries` tests it), but the asymmetry is easy to get wrong: fallback/agentic steps must re-add `Content-Type`, `Accept`, and any auth, or LLM providers will 400/415. We should take care to document this very explicitly for operators.

docs/filters/http/traffic_management/iterative_request_router.md
```diff
@@ -0,0 +1,68 @@
+<!-- Generated by: cargo xtask generate-filter-docs -->
+<!-- Do not edit manually -->
+
+# `iterative_request_router`
+
+Framework-level filter for iterative sub-request execution.
+
+## Configuration Notes
+
+Holds named steps, each backed by a pre-built sub-pipeline. During request processing, runs an iteration loop: execute each step's request filters, make the HTTP call via Pingora's `Connector`, execute its response filters, evaluate transition rules, and continue or return the final response.
```
usize, line 10: The proposal lists "streaming intermediate responses" as a non-goal, but the full consequence for AI users isn't stated anywhere: the sub-request executor buffers the entire upstream response (`max_response_bytes`, default 10 MiB), the *final* response is also returned fully buffered to the client via the rejection path, and the default 30s overall deadline will kill long-lived SSE streams.

These are big limitations. We should propose following up with a plan for supporting streaming fallback responses.

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
```diff
@@ -0,0 +1,1067 @@
...
+
+            let transitions = self
+                .step_transitions
+                .get(&current_step)
+                .map_or(&[][..], |v| v.as_slice());
+
+            match evaluate_transitions(transitions, &response, &step_filter_results) {
+                TransitionResult::Done => {
+                    debug!(step = current_step.as_ref(), "iteration complete, returning response");
+                    return Ok(FilterAction::Reject(build_response_rejection(
```
usize: Because of `FilterAction::Reject`, we'd not run e.g., a token counting filter after this right? Is it possible for us to end up in a state where some tokens were produced but we never count them if we don't order this filter at the end of the pipeline?

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
```diff
@@ -0,0 +1,1067 @@
...
+                "executing step"
+            );
+
+            let mut sub_headers = current_request.headers.clone();
+            strip_reserved_headers(&mut sub_headers);
+            // Praxis ingress deliberately rejects this reserved marker from
+            // every network peer. That prevents spoofing and terminates a
+            // cross-listener cycle at its first hop; do not weaken the ingress
+            // boundary merely to permit recursive proxy topologies.
+            if let Ok(depth_val) = http::HeaderValue::from_str(&(depth + 1).to_string()) {
```
usize: The framework-owned invariant from #826 ("depth cannot be supplied or overridden through ordinary request paths") argues for re-asserting the value after sanitization — e.g., strip depth from filter mutations entirely and inject it at dispatch time (or immediately re-insert it after line 413).

Different approach from what's here it seems, and has better correctness guarantees.

core/src/subrequest.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl SubRequestConnector {
+    /// Create a connector with the given keepalive pool size.
+    ///
+    /// ```
+    /// use praxis_core::subrequest::SubRequestConnector;
+    ///
+    /// let connector = SubRequestConnector::new(64);
+    /// ```
+    pub fn new(keepalive_pool_size: usize) -> Self {
```
usize: Automated review highlights that `keepalive_pool_size` bounds *idle* connections only. We should follow with a limit for live connections (as a followup).

examples/configs/traffic-management/iterative-request-router-failover.yaml
```diff
@@ -0,0 +1,53 @@
...
+                routes:
+                  - path_prefix: "/"
+                    cluster: primary-backend
+              - filter: load_balancer
+                clusters:
+                  - name: primary-backend
+                    endpoints:
+                      - "127.0.0.1:3000"
+            on_result:
+              - status: [502, 503, 504]
```
usize, line 36: Nit: The PR description cites #549 (quota-exhaustion failover) and there's a `failover_429_quota_exhaustion` integration test, but the headline example only transitions on `[502, 503, 504]`. Adding 429 here (with a comment about `Retry-After` handling and the local-vs-upstream 429 caveat from `subresponse_from_rejection`) would make the example match the advertised use case.

usize (approved):
Of course, this is a huge PR. I used Claude to help me zoom into relevant sections. I checked out what it said. Much of it boils down to decisions that have been made e.g., compressing `transport_failure_status` into a single HTTP response code with non-obvious implications for operators.

Some other things are questions I have, based on the original review that the model made. Because they are also decisions that I think could be valid but that have tricky implications.

None of it struck me as a blocker per se. Useful followups and issues and some changes to documentation should cover it in general.

I appreciate you taking this on. Using reqwest for the callout client seemed like the thing to do at the time, but this is clearly better.

docs/filters/http/traffic_management/iterative_request_router.md
leseb: Yeah streaming is a must-have for the ai project.

core/src/config/runtime.rs
```diff
@@ -113,6 +114,18 @@ pub struct RuntimeConfig {
...
+    /// used by `iterative_request_router` step chains.
+    ///
+    /// ```
+    /// use praxis_core::config::RuntimeConfig;
+    ///
+    /// let cfg = RuntimeConfig::default();
+    /// assert_eq!(cfg.subrequest_pool_size, Some(128));
+    /// ```
+    #[serde(default = "default_subrequest_pool_size")]
+    pub subrequest_pool_size: Option<usize>,
```
leseb, line 147: call this `subrequest_keepalive_pool_size` instead?

examples/configs/traffic-management/iterative-request-router-failover.yaml
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

core/src/subrequest.rs
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

docs/filters/http/traffic_management/iterative_request_router.md
leseb: Decision is to move subrequest types to core and add SubRequestClient in a followup PR.

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
leseb: https://github.com/praxis-proxy/praxis/pull/849/commits/c74afc3b828e4823c38ebdb0e02bde528560e104

docs/proposals/00358_http-callout-filter.md
leseb: Decision is to followup on that.

leseb (approved):

PR #852, opened 2026-07-23
docs(proposals): add initial proposal for state

### LINKED ISSUE #147: Spike: Storage + State
Due to the nature of AI inference, agents, etc there are lots of places where integration with storage (e.g. files API, etc) needs to be evaluated and designed. The purpose of this spike is to do the initial validation, and get an initial proposal down which covers the "What?" and "Why?" describing what we want to do at a high level, and the motivation for doing so.

The most pressing use cases is to determine how we want to manage state for responses, e.g. praxis-proxy/ai#95

shaneutt:
### What does this PR do?

This adds the initial "What?" and "Why?" with regard to the interfaces and management of State (ephemeral/storage) in Praxis.

Partially resolves https://github.com/praxis-proxy/ai/issues/147

### DIFF
--- a/docs/proposals/00412_unified-state-interface.md
+++ b/docs/proposals/00412_unified-state-interface.md
@@ -1,0 +1,183 @@
+---
+issue: https://github.com/praxis-proxy/praxis/issues/412
+discussion: https://github.com/praxis-proxy/praxis/issues/412
+status: proposed
+authors:
+  - shaneutt
+graduation_criteria:
+  - State type hierarchy agreed by stakeholders
+  - Scoping model agreed
+  - How? section with requirements and design
+  - Determine whether Ephemeral and Persistent storage really need to be separate and resolve
+stakeholders:
+  - shaneutt
+  - nerdalert
+  - rikatz
+  - leseb
+  - twghu
+---
+
+# Unified State Interface
+
+## What?
+
+Create a standard interface and machinery for State
+in Praxis core. State is the single top-level
+abstraction. It has two categories:
+
+- **Ephemeral state** covers backends like Valkey
+  and in-memory stores. Data may be lost on restart.
+  Suitable for caches, counters, and session
+  affinity across requests and replicas.
+
+- **Storage state** covers persistent backends like
+  PostgreSQL, SQLite, and file stores. Data survives
+  restarts. Suitable for conversation history,
+  response records, and durable business data.
+
+Filters and other consumers (such as probes)
+declare their state needs in configuration. Praxis
+manages backend lifecycle, configuration, and
+scoped access. State backends are scoped to a
+specific filter, a named chain, or made globally
+available with explicit opt-in.
+
+This proposal complements proposal 00099 (Stateful
+Proxy State Management). Proposal 00099 defines the
+state *model*: which classes of state exist and what
+conventions govern each. This proposal defines the
+concrete *interfaces*: traits, configuration schema,
+scoping rules, and backend implementations that
+make that model operational.
+
+### Goals
+
+- A single interface that covers both ephemeral
+  and persistent state under one abstraction.
+- Config-driven backend lifecycle: operators
+  declare backends in YAML, Praxis provisions
+  them at startup.
+- Scoped access by default: backends are bound to
+  a filter or chain. Global access requires
+  explicit configuration.
+- Pluggable backends: core defines traits and ships
+  common defaults (in-memory, Valkey, SQLite,
+  OpenAI Files API). External crates provide
+  additional
+  backends (e.g. PostgreSQL) without modifying
+  core.
+- Accessible to filters and to probes (background
+  processes), decoupled from the HTTP request
+  lifecycle.
+
+## Why?
+
+### Ephemeral State
+
+Praxis needs ephemeral state to function as a
+stateful proxy.
+
+**Multi-instance deployments.** Distributed
+ephemeral state (Valkey) enables session affinity,
+rate limiting, and feature flags that are consistent
+across proxy replicas. Without a standard backend,
+each feature that needs distributed ephemeral state
+would build its own Redis client, key format,
+timeout policy, and failure semantics.
+
+**Stateful protocols.** MCP uses session IDs that
+bind follow-up requests to the backend that owns
+the session context. A2A tracks long-running tasks
+across multiple request/response exchanges.
+WebSocket and gRPC streaming sessions need
+connection-to-session mapping. Without shared
+ephemeral state, these protocols cannot be proxied
+correctly across multiple requests.
+
+**Cross-request state.** Rate limiters track
+request counts across calls. Circuit breakers
+accumulate failure counts over time. Health
+snapshots inform load balancing decisions. These
+patterns need state that persists across requests
+and survives configuration reloads, but does not
+need to survive a process restart.
+Today each builds its own in-process store
+(DashMap, atomics) with ad-hoc capacity limits
+and no shared lifecycle management.
+
+### Storage State
+
+Praxis needs persistent storage for the Responses
+API, Conversations API, and the agentic loop.
+
+**Responses and Conversations APIs.** The OpenAI
+Responses API is stateful by design. Each response
+can reference a `previous_response_id` to continue
+a conversation. The Conversations API maintains
+accumulated message history across turns. Both
+must persist records so that subsequent requests
+can rehydrate context from earlier turns. This
+already works today via `ResponseStore` and
+`ConversationItemStore` with SQLite and PostgreSQL
+backends, but the implementations are
+self-contained in the ai/ repository with their
+own traits, registries, and configuration. Nothing
+else can reuse them.
+
+**Agentic loop orchestration.** The agentic loop
+executes multiple inference rounds, accumulating
+tool call results, conversation messages, and token
+usage across iterations. When `store: true` is set,
+the full conversation state must be persisted
+durably so that clients can retrieve or continue it
+later. Conversation compare-and-swap (already
+implemented in the ai/ repo) shows the need for
+concurrent-safe persistent state operations.
+
+**Near-term consumers.** MCP session persistence
+requires durable session-to-backend mappings that
+survive proxy restarts and are shareable across
+replicas. A2A task state tracks long-running task
+lifecycle (submitted, working, completed, failed)
+across multiple request/response exchanges. Both
+are protocol-level concerns, not OpenAI-specific.
+The Files API needs local file storage for
+development environments (praxis-proxy/ai#494).
+Each of these would otherwise build its own
+trait, registry, and configuration - the same
+scaffolding that `ResponseStore` and
+`ConversationItemStore` already duplicated.
+
+**Shared need.** Both ephemeral and storage state
+need the same infrastructure: named backends
+declared in configuration, scoped access control,
+managed lifecycle, pluggable implementations, and
+availability outside the HTTP request path (for
+probes and background processes). Building this
+once as a unified interface avoids duplicating the
+machinery for each new feature.
+
+### User Stories
+
+- As a proxy operator, I want to declare state
+  backends in YAML so that I manage connectivity,
+  credentials, and TLS centrally instead of per
+  filter.
+- As a filter author, I want to request a named
+  state backend through the filter context so that
+  I do not manage backend lifecycle or connection
+  pooling myself.
+- As a platform engineer, I want state scoped to
+  specific filters by default so that one filter
+  cannot corrupt another's state.
+- As an AI gateway operator, I want the same
+  interface for conversation persistence, file
+  caching, and session tracking so that each
+  feature does not build its own storage layer.
+- As a Praxis developer, I want to add a new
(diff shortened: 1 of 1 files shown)


docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+
+**Multi-instance deployments.** Distributed
+ephemeral state (Valkey) enables session affinity,
+rate limiting, and feature flags that are consistent
+across proxy replicas. Without a standard backend,
+each feature that needs distributed ephemeral state
+would build its own Redis client, key format,
+timeout policy, and failure semantics.
+
+### Storage State
```
leseb, line 108: Conversations API is the second concrete consumer

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+Praxis needs persistent storage for the Responses
+API and the agentic loop.
+
+**Responses API.** The OpenAI Responses API is
+stateful by design. Each response can reference a
+`previous_response_id` to continue a conversation.
+The proxy must persist response records and
+conversation history so that subsequent requests
+can rehydrate context from earlier turns. This
+already works today via `ResponseStore` with SQLite
```
leseb: Yes but what are the new consumers that justify this unified layer?

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+in Praxis core. State is the single top-level
+abstraction. It has two categories:
+
+- **Ephemeral state** covers backends like Valkey
+  and in-memory stores. Data may be lost on restart.
+  Suitable for caches, counters, session affinity,
+  and cross-filter coordination within a pipeline.
+
+- **Storage state** covers persistent backends like
+  PostgreSQL, SQLite, and filesystems. Data survives
```
leseb: I think filesystems is too broad here, remove it?

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+## What?
+
+Create a standard interface and machinery for State
+in Praxis core. State is the single top-level
+abstraction. It has two categories:
+
+- **Ephemeral state** covers backends like Valkey
+  and in-memory stores. Data may be lost on restart.
+  Suitable for caches, counters, session affinity,
+  and cross-filter coordination within a pipeline.
```
leseb: I don't think we use any in-memory store for "cross-filter coordination within a pipeline". `ResponsesState` should be sufficient.

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+**Stateful protocols.** MCP uses session IDs that
+bind follow-up requests to the backend that owns
+the session context. A2A tracks long-running tasks
+across multiple request/response exchanges.
+WebSocket and gRPC streaming sessions need
+connection-to-session mapping. Without shared
+ephemeral state, these protocols cannot be proxied
+correctly across multiple requests.
+
+**Cross-filter coordination.** The iterative
```
leseb: Same comment about why ResponsesState is not enough here.

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+request router executes multiple sub-requests in a
+loop, accumulating state between iterations.
+Branch chains evaluate filter results to choose
+pipeline paths. Rate limiters track request counts
+across calls. These patterns all need state that
+persists across filter invocations within or across
+requests, but does not need to survive a restart.
+
+**Multi-instance deployments.** Distributed
```
leseb: This is a valid justification for not using ResponsesState, maybe the only one? If so, I guess this should be the main driver for having ephemeral state in praxis core. Within-request coordination should not be part of the justification.

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+once as a unified interface avoids duplicating the
+machinery for each new feature.
+
+### User Stories
+
+- As a proxy operator, I want to declare state
+  backends in YAML so that I manage connectivity,
+  credentials, and TLS centrally instead of per
+  filter.
+- As a filter author, I want to request a named
```
leseb, line 166: Yes!

leseb (changes requested):
I generally agree about the proposal, I think this is something Praxis should have eventually. However, as written the proposal does not justify a broad unified persistent-store abstraction since all consumers are OpenAI API related.
Do we have any new consumer that require this unified abstraction?

Also, I still think we should have 2 proposals one for ephemeral and one for persistent storage.

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+  and persistent state under one abstraction.
+- Config-driven backend lifecycle: operators
+  declare backends in YAML, Praxis provisions
+  them at startup.
+- Scoped access by default: backends are bound to
+  a filter or chain. Global access requires
+  explicit configuration.
+- Pluggable backends: core defines traits and ships
+  common defaults (in-memory, Valkey, SQLite,
+  filesystem). External crates provide additional
```
leseb: filesystem consumer is Files API (for the local dev env) https://github.com/praxis-proxy/ai/issues/494

docs/proposals/00412_unified-state-interface.md
shaneutt:
```suggestion
  PostgreSQL, SQLite, and file stores. Data survives
```

docs/proposals/00412_unified-state-interface.md
shaneutt:
```suggestion
  Files API). External crates provide additional
```

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+  and persistent state under one abstraction.
+- Config-driven backend lifecycle: operators
+  declare backends in YAML, Praxis provisions
+  them at startup.
+- Scoped access by default: backends are bound to
+  a filter or chain. Global access requires
+  explicit configuration.
+- Pluggable backends: core defines traits and ships
+  common defaults (in-memory, Valkey, SQLite,
+  Files API). External crates provide additional
```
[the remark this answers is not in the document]
franciscojavierarceo: from it's OpenAI Files API i think most users of this repo know that fwiw

franciscojavierarceo:
I definitely agree with the direction of collapsing this into one State subsystem. I don’t think ephemeral and persistent state need separate top-level abstractions since most of the machinery around configuration, lifecycle, scoping, health, credentials, and observability is the same.

I think we just need to be clear that one State subsystem doesn’t mean every backend has to support the same operations.

The way I’m thinking about it is: 
- A provider/backend is the actual implementation: Valkey, PostgreSQL, SQLite, filesystem, S3, Infinispan, Elastic, Milvus, Weaviate, Qdrant, ChromaDB, etc.
- A named state resource says how we use that backend: capability, namespace, scope, timeouts, failure behavior, and so on.
- A consumer is something like a filter, probe, MCP handler, Responses API, Files API, or Vector Stores API that asks for that named resource.

This would let us configure one Valkey backend and use it for both mcp-sessions and routing-cache, while still giving them different namespaces and failure policies.

I think the existing KvStoreRegistry should also become part of this subsystem rather than remaining separate. KV would be one capability alongside things like sessions, response storage, blobs/files, and vector storage.

We should also make sure the model is expressive enough for future database providers. Vector stores are a good example: Milvus, Weaviate, Qdrant, ChromaDB, or Elastic could provide vector-search capabilities, while the Vector Stores API or a retrieval filter would consume them. A provider may expose more than one capability too—PostgreSQL with pgvector could provide relational, transactional, and vector operations.

Provider-specific configuration is probably the complicated part. IMO Praxis should own the common structure and lifecycle, but each provider needs to own and validate its connection, TLS, pooling, topology, migration, and engine-specific configuration. We shouldn’t try to normalize all database configuration into one lowest-common-denominator schema.
Also, filesystem or S3 would be providers for file storage; the Files API itself is a consumer. I think making that terminology clear would provide some clarity.

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+ephemeral state, these protocols cannot be proxied
+correctly across multiple requests.
+
+**Cross-filter coordination.** The iterative
+request router executes multiple sub-requests in a
+loop, accumulating state between iterations.
+Branch chains evaluate filter results to choose
+pipeline paths. Rate limiters track request counts
+across calls. These patterns all need state that
+persists across filter invocations within or across
```
aslakknutsen: "filter invocations and filter instance". Not sure if it needs the additional context, however it might not require to survive a restart, but it might require to survive a (filter) reload.

docs/proposals/00412_unified-state-interface.md
shaneutt: :+1: updated in 016a5e9

docs/proposals/00412_unified-state-interface.md
shaneutt: :+1: updated in 016a5e9 - added a 'Near-term consumers' paragraph covering MCP session persistence, A2A task state, and Files API local storage. These are protocol-level concerns, not OpenAI-specific. Also called out the scaffolding duplication problem (ResponseStore and ConversationItemStore already duplicated the trait+registry+config pattern).

docs/proposals/00412_unified-state-interface.md
shaneutt: :+1: updated in 016a5e9 - removed 'cross-filter coordination within a pipeline' from the ephemeral state description. Narrowed to cross-request and cross-instance use cases, which are the actual gaps. Note that `ResponsesState` and similar patterns are expected to become implementations of the core-level state traits, not alternatives to them.

docs/proposals/00412_unified-state-interface.md
shaneutt: :+1: updated in 016a5e9 - rewrote this section as 'Cross-request state' focused on rate limiters, circuit breakers, and health snapshots. These need state across requests, which `ResponsesState` (request-scoped) doesn't cover. Removed the iterative router and branch chain references since those are within-request mechanisms already solved.

docs/proposals/00412_unified-state-interface.md
shaneutt: :+1: updated in 016a5e9 - reordered to lead with multi-instance deployments as the primary driver. Within-request coordination removed from the justification.

docs/proposals/00412_unified-state-interface.md
```diff
@@ -0,0 +1,155 @@
...
+  - How? section with requirements and design
+stakeholders:
+  - shaneutt
+  - nerdalert
+  - rikatz
+  - leseb
+  - twghu
+---
+
+# Unified State Interface
```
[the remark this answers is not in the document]
shaneutt: :+1: updated in 016a5e9 - added a relationship note. This proposal complements 00099: 00099 defines the state *model* (which classes of state exist), this proposal defines the concrete *interfaces* (traits, config, scoping) that make that model operational. We may want to merge the two proposals at some point once the direction solidifies.

docs/proposals/00412_unified-state-interface.md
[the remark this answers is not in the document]
shaneutt: :+1: updated in 016a5e9 - clarified to 'OpenAI Files API' to remove ambiguity. It refers to the consumer (OpenAI Files API needing local file storage), not a backend type.

docs/proposals/00412_unified-state-interface.md
shaneutt: :+1: updated in 016a5e9. Added 'and survives configuration reloads' to the description. This is a real concern today (e.g. rate limiter state resets on hot reload).

shaneutt:
> I don’t think ephemeral and persistent state need separate top-level abstractions since most of the machinery around configuration, lifecycle, scoping, health, credentials, and observability is the same.

I'm open to collapsing them into one. The purpose in having them differentiated was wanting to make a clear dilineation between optimization data, and data that requires retention. I would like to leave it and tease it out in the "How?" if that's agreeable? I've added it to graduation criteria so it is not forgotten.

shaneutt:
> I think we just need to be clear that one State subsystem doesn’t mean every backend has to support the same operations.

Agreed :+1:

shaneutt:
> Do we have any new consumer that require this unified abstraction?

Added.

> I generally agree about the proposal, I think this is something Praxis should have eventually. However, as written the proposal does not justify a broad unified persistent-store abstraction since all consumers are OpenAI API related.

If we do not do this now, the current implementation gets more and more settled in, and more difficult to change. Even with just one user right now, I would still want to have this as part of the core framework so that the path we intend to take is made clear, for us when we add more state options in the future (seems very likely) or users of the framework.

leseb (approved):

franciscojavierarceo (approved):

PR #857, opened 2026-07-23
fix(server): race condition between config file and watcher setup/loop

gdozortsev:
### Summary
Fixes failing hot_reload test cases 

### Root Cause
Race condition between the file watcher thread startup and the HTTP server becoming ready. When integration tests write a new config immediately after wait_for_http returns, the watcher hasn't started watching yet, so the filesystem event is missed entirely. 

### Fix
1. add initial_content_hash to WatcherParams before the watcher thread spawns. This ensures the watcher's baseline hash reflects what the server actually loaded, not whatever the file contains when the watcher eventually starts.
2. add a check for a new config before the watcher loop starts (when the watcher is ready but hasn't started yet). If the config was changed in startup it will trigger a reload, if nothing changed its no-op. 

##Verification
(these were all failing before)
`cargo test -p praxis-tests-integration --test suite -- hot_reload `

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -168,9 +168,11 @@ fn spawn_watcher(
     state: ServerState,
 ) -> Option<std::thread::JoinHandle<()>> {
     let path = config_path?;
+    let initial_content_hash = std::fs::read_to_string(&path).map_or(0, |c| crate::watcher::hash_content(&c));
     let handle = crate::watcher::spawn_config_watcher(crate::watcher::WatcherParams {
         config_path: path,
         health_shutdown: state.health_shutdown,
+        initial_content_hash,
         initial_config: config,
         kv_stores: state.kv_stores,
         pipelines: state.pipelines,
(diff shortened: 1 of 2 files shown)


alexsnaps:
Thanks for the PR, can you please fix the commit to be signed off?

Not that I think this fix is wrong, but I'm thinking we might be capable of getting rid of the race entirely by having the watcher thread being the one loading the config _always_. Not asking any changes per se _right now_, but I'll look into what 🔥 the race (rather than dealing with it) would mean, as soon as I get out of meeting hell! I might file an issue to follow up based on my findings there.

gdozortsev:
> Thanks for the PR, can you please fix the commit to be signed off?

fixed!

> 
> Not that I think this fix is wrong, but I'm thinking we might be capable of getting rid of the race entirely by having the watcher thread being the one loading the config _always_. Not asking any changes per se _right now_, but I'll look into what 🔥 the race (rather than dealing with it) would mean, as soon as I get out of meeting hell! I might file an issue to follow up based on my findings there.

Makes sense! I don't know the repo well enough yet for that but was just trying to fix the failing test cases :)

alexsnaps (approved):
> Makes sense! I don't know the repo well enough yet for that but was just trying to fix the failing test cases :)

And it is very much appreciated! 🙏 
I'll merge this in any case (the additional field is absolutely worth it, certainly better than the _losing_ the race!). But as someone once said: less is more or something ;)

shaneutt:
Please create a follow-up issue for this interim fix :+1:

PR #897, opened 2026-08-02
feat(load-balancer): support upstream authority override

nerdalert:
## Summary

  Adds an optional per-cluster HTTP authority override to the Praxis load_balancer.

  When a cluster configures authority, Praxis replaces the downstream Host header before sending the request upstream. TLS SNI remains independently configured through tls.sni.

```yaml
  clusters:
    - name: openai-api
      endpoints:
        - api.openai.com:443
      authority: api.openai.com
      tls:
        sni: api.openai.com
        verify: true
```

  This allows Praxis to proxy through internal routing layers while still presenting the authority expected by the final upstream service.

  ## Why This Is Needed

  A downstream request may enter Praxis using an internal or public Grid hostname:

```
  client
    Host: api.grid.example.com
      |
      v
  consumer gateway
      |
      | authenticated provider hop
      | Host may identify the provider gateway
      v
  provider gateway
      |
      | TLS SNI: api.openai.com
      | Host: api.openai.com
      v
  OpenAI
```

  Selecting api.openai.com:443 and setting TLS SNI are not sufficient by themselves. Without an authority override, the HTTP request can retain a downstream hostname such as:

```
  Host: api.grid.example.com
```

  or:
```
  Host: east-provider.grid.internal
```

  The TLS handshake may succeed because SNI is api.openai.com, but the HTTP request still names a different virtual host. Virtual-hosted upstream services can reject that request or route it
  incorrectly.

  The new field makes the final upstream request internally consistent:

```
  TCP endpoint: api.openai.com:443
  TLS SNI:     api.openai.com
  HTTP Host:   api.openai.com
```

  ## Why This Belongs In load_balancer

  intelligent_route selects a configured cluster. It does not construct the upstream HTTP request or own transport behavior.

```
  request classification
    -> intelligent_route selects "openai-api"
    -> load_balancer resolves the cluster endpoint
    -> load_balancer constructs the upstream
    -> protocol handler prepares the upstream HTTP request
```

  The authority override is a property of the selected upstream cluster, alongside its endpoints, connection settings, TLS configuration, and SNI. The load-balancing path is where that
  cluster configuration becomes a concrete upstream connection and request.

  Implementing this in intelligent_route would create several problems:

  - It would make generic HTTP upstream behavior depend on an AI-specific routing filter.
  - Routes that use router and load_balancer without intelligent_route would not receive the override.
  - Another filter selecting the same cluster could behave differently.
  - intelligent_route would need knowledge of HTTP transport details that it does not otherwise own.
  - The authority value would need to travel through request metadata instead of remaining part of validated cluster state.
  - Downstream-supplied Host values could be handled inconsistently across routing paths.

  Keeping the feature in load_balancer preserves the existing ownership boundary:

  - Routing filters select a cluster.
  - Cluster configuration describes how to reach that upstream.
  - The load balancer resolves the selected cluster.
  - The protocol layer applies the selected upstream’s HTTP authority.

  There is no Grid-specific API or runtime dependency in this change.

  ## Configuration Contract

  The new cluster field is optional:

```
  authority: api.example.com
```

  Supported forms include:

```
  api.example.com
  api.example.com:8443
  localhost
  localhost:8080
  [2001:db8::1]
  [2001:db8::1]:8443
```

  Configuration validation rejects:

  - Empty values
  - Control characters or whitespace
  - URI schemes
  - Paths
  - Query strings
  - Fragments
  - User information
  - Invalid DNS labels
  - Unbracketed IPv6 addresses
  - Empty, zero, malformed, or out-of-range ports
  - DNS hostnames over 253 characters (the optional port is validated separately)

  An invalid authority fails filter construction. Praxis does not silently ignore the value and fall back to forwarding the downstream Host.

  ## Runtime Behavior

  When authority is configured:

  1. The value is validated during configuration loading.
  2. It is parsed into an HTTP HeaderValue while building the cluster entry.
  3. The parsed value is carried with the selected Upstream.
  4. The upstream request handler replaces the Host header after reserved and hop-by-hop header stripping.
  5. Absolute-form request URIs also have their authority component replaced.
  6. Origin-form request URIs retain their existing path and query representation.

  When authority is absent, existing behavior is unchanged and the downstream Host continues through the normal request path.

  Praxis currently uses HTTP/1.1 for upstream connections, so the active behavior is an upstream Host replacement. The implementation also keeps URI authority handling consistent if an HTTP/2
  upstream path is added later.

  ## TLS SNI Independence

  authority does not modify TLS SNI.

  This separation is intentional because the connection identity and HTTP virtual host can differ:

```
  authority: tenant-api.example.com
  tls:
    sni: shared-edge.example.net
```

  Operators must configure both values when the upstream requires both a specific TLS server name and a specific HTTP virtual host.

  ## Security Properties

  - Invalid authority values fail during startup.
  - Downstream-controlled Host values are replaced when an override is configured.
  - The override is applied after reserved and hop-by-hop header stripping.
  - Authority is cluster-local and cannot leak between clusters.
  - Absolute-form URIs cannot retain a conflicting downstream authority.
  - No routing metadata or caller header can dynamically choose the override.
  - The change does not weaken TLS certificate or SNI verification.

  This supports external provider routing without making caller-controlled headers authoritative at the final upstream hop.

  ## Performance

  The expected request-path cost is small and applies only to clusters that configure authority.

  The implementation avoids repeated parsing:

  - Configuration stores the source value once.
  - Cluster construction parses it into HeaderValue once.
  - Requests reuse the cached value.
  - Cloning HeaderValue is inexpensive because its backing storage is shared.
  - Normal origin-form requests perform one header replacement and do not rebuild the URI.
  - Absolute-form requests rebuild the URI to replace its authority; these are uncommon in the normal reverse-proxy path.
  - Clusters without authority take an early return and preserve the existing request behavior.

  Memory overhead is one optional cached HeaderValue per configured cluster plus the inexpensive value carried by the selected per-request Upstream.

  The implementation does not add:

  - Per-request DNS work
  - File or environment reads
  - Kubernetes or Grid queries
  - New shared locks
  - Unbounded caches
  - Request-body copies
  - Credential processing
  - String-to-header parsing on every request

  ## Testing

  Coverage includes:

  - Valid hostname, port, localhost, and bracketed IPv6 forms
  - Invalid DNS labels and ports
  - Schemes, paths, fragments, user information, whitespace, and control characters
  - Schema acceptance and rejection
  - Serialization round trips
  - Independence between authority and tls.sni
  - Startup failure for semantically invalid embedded and programmatic authority values
  - Host replacement for HTTP/1.1 requests
  - HTTP/2 downstream requests forwarded to an HTTP/1.1 upstream
  - Downstream Host replacement
  - Absolute-form URI normalization
  - Origin-form URI preservation
  - Isolation between clusters with and without an override
  - Stability across sequential requests
  - Functional proxy traffic through the example configuration

  The Grid integration was also validated with a live OpenAI request:

  client
    -> Grid consumer gateway
    -> authenticated provider gateway
    -> credential replacement
    -> OpenAI

  Observed evidence included:

  - HTTP 200 from the OpenAI Responses API
  - OpenAI response-object validation
  - Selected Grid edge
  - Authenticated provider gateway
  - Exact OpenAI candidate
  - Provider-local openai-api cluster
  - Exact serving overlay revision
  - Successful replacement of deliberately invalid caller authorization

  ## Validation

  - make lint
  - make test
  - make doc
  - cargo check --workspace
  - cargo clippy --workspace --all-targets --all-features -- -D warnings
  - RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
  - cargo xtask lint-filter-docs
  - cargo xtask lint-example-tests
  - cargo xtask sync-example-readme
  - git diff --check
  - Cold Grid quick demo with a live OpenAI provider
  - Cold normal-mode Grid regression with no external-provider resources

  ## Scope

  This PR adds a generic per-cluster HTTP authority override.

  It does not add:

  - Grid-aware behavior to Praxis core
  - AI model or provider selection
  - Provider credentials
  - API translation
  - Dynamic cluster creation
  - DNS or Kubernetes lookups
  - HTTP/2 upstream transport
  - Automatic inference-provider configuration

  Grid and Praxis AI consume this generic cluster capability in their own configuration and routing layers.

### DIFF
--- a/core/src/config/cluster/mod.rs
+++ b/core/src/config/cluster/mod.rs
@@ -21,6 +21,8 @@ pub use retry_policy::{
 };
 use serde::{Deserialize, Serialize};
 
+use crate::errors::ProxyError;
+
 // -----------------------------------------------------------------------------
 // Cluster
 // -----------------------------------------------------------------------------
@@ -46,6 +48,36 @@ use serde::{Deserialize, Serialize};
 pub struct Cluster {
     /// Unique name for the cluster.
     pub name: Arc<str>,
+
+    /// Override the upstream HTTP `Host` header. **HTTP only** — this
+    /// field has no effect on TCP load balancers, which do not process
+    /// HTTP headers.
+    ///
+    /// When set, the proxy rewrites the `Host` header sent to the
+    /// upstream instead of forwarding the downstream value. Upstream
+    /// connections use HTTP/1.1; the `:authority` pseudo-header on
+    /// the downstream HTTP/2 side is not forwarded upstream. TLS SNI
+    /// remains independent — configure `tls.sni` separately when
+    /// needed.
+    ///
+    /// Must be a valid HTTP authority: a hostname with an optional
+    /// port, or a bracketed IPv6 address with an optional port. URI
+    /// schemes, paths, userinfo, and fragments are rejected.
+    ///
+    /// ```
+    /// # use praxis_core::config::Cluster;
+    /// let yaml = r#"
+    /// name: "api"
+    /// endpoints: ["10.0.0.1:443"]
+    /// authority: "api.example.com"
+    /// tls:
+    ///   sni: "api.example.com"
+    /// "#;
+    /// let cluster: Cluster = serde_yaml::from_str(yaml).unwrap();
+    /// assert_eq!(cluster.authority.as_deref(), Some("api.example.com"));
+    /// ```
+    #[serde(default)]
+    pub authority: Option<Arc<str>>,
 
     /// TCP connection timeout in milliseconds.
     ///
@@ -138,6 +170,20 @@ pub struct Cluster {
 }
 
 impl Cluster {
+    /// Validate the optional upstream HTTP authority override.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ProxyError::Config`] when the authority is not a
+    /// supported hostname with an optional port or a bracketed IPv6
+    /// address with an optional port.
+    pub fn validate_authority(&self) -> Result<(), ProxyError> {
+        let Some(authority) = self.authority.as_deref() else {
+            return Ok(());
+        };
+        super::validate::cluster::validate_authority(authority, &self.name)
+    }
+
     /// Build a cluster with only a name and endpoints; all other
     /// fields use their defaults (no timeouts, no TLS, no health
     /// check, `round_robin` strategy).
@@ -157,6 +203,7 @@ impl Cluster {
     pub fn with_defaults(name: &str, endpoints: Vec<Endpoint>) -> Self {
         Self {
             name: Arc::from(name),
+            authority: None,
             connection_timeout_ms: None,
             endpoints,
             health_check: None,
--- a/core/src/config/validate/cluster/authority.rs
+++ b/core/src/config/validate/cluster/authority.rs
@@ -1,0 +1,414 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Praxis Contributors
+
+//! HTTP authority validation for per-cluster upstream authority override.
+
+use crate::errors::ProxyError;
+
+/// Maximum DNS hostname length without a trailing root label.
+const MAX_HOSTNAME_LEN: usize = 253;
+
+/// Validate a per-cluster authority override value.
+///
+/// The supported form is `host [ ":" port ]`, where `host` is an
+/// ASCII DNS hostname or bracketed IPv6 address. Schemes, paths,
+/// userinfo, query strings, and fragments are rejected.
+pub(super) fn validate_authority(authority: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    if authority.is_empty() {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority must not be empty"
+        )));
+    }
+
+    reject_control_chars(authority, cluster_name)?;
+    reject_whitespace(authority, cluster_name)?;
+    reject_uri_components(authority, cluster_name)?;
+    validate_host_port(authority, cluster_name)
+}
+
+/// Reject ASCII control characters (C0 range and DEL).
+fn reject_control_chars(authority: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    if authority.bytes().any(|b| b < 0x20 || b == 0x7F) {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority contains control characters"
+        )));
+    }
+    Ok(())
+}
+
+/// Reject space and tab characters.
+fn reject_whitespace(authority: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    if authority.bytes().any(|b| b == b' ' || b == b'\t') {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority contains whitespace"
+        )));
+    }
+    Ok(())
+}
+
+/// Reject URI components that are not part of a bare authority.
+fn reject_uri_components(authority: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    if authority.contains("://") {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority must not contain a URI scheme"
+        )));
+    }
+    if authority.contains('/') {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority must not contain a path"
+        )));
+    }
+    if authority.contains('@') {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority must not contain userinfo"
+        )));
+    }
+    if authority.contains('#') {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority must not contain a fragment"
+        )));
+    }
+    if authority.contains('?') {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority must not contain a query string"
+        )));
+    }
+    Ok(())
+}
+
+/// Validate the host and optional port components.
+fn validate_host_port(authority: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    if let Some(rest) = authority.strip_prefix('[') {
+        validate_bracketed_ipv6(rest, cluster_name)
+    } else {
+        validate_hostname_port(authority, cluster_name)
+    }
+}
+
+/// Validate `[ipv6]:port` or `[ipv6]` form.
+fn validate_bracketed_ipv6(rest: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    let Some(bracket_end) = rest.find(']') else {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority has unclosed IPv6 bracket"
+        )));
+    };
+    let ipv6 = rest.get(..bracket_end).unwrap_or_default();
+    if ipv6.parse::<std::net::Ipv6Addr>().is_err() {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority contains invalid IPv6 address"
+        )));
+    }
+    let after = rest.get(bracket_end + 1..).unwrap_or_default();
+    if after.is_empty() {
+        return Ok(());
+    }
+    if let Some(port_str) = after.strip_prefix(':') {
+        validate_port(port_str, cluster_name)
+    } else {
+        Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority has unexpected characters after IPv6 address"
+        )))
+    }
+}
+
+/// Validate `hostname:port` or `hostname` form.
+fn validate_hostname_port(authority: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    let (host, port) = match authority.rsplit_once(':') {
+        Some((h, p)) if p.bytes().all(|b| b.is_ascii_digit()) && !p.is_empty() => (h, Some(p)),
+        _ => (authority, None),
+    };
+
+    if host.is_empty() {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority hostname is empty"
+        )));
+    }
+
+    for b in host.bytes() {
+        if !b.is_ascii_alphanumeric() && b != b'-' && b != b'.' {
+            return Err(ProxyError::Config(format!(
+                "cluster '{cluster_name}': authority hostname contains invalid characters"
+            )));
+        }
+    }
+
+    validate_dns_labels(host, cluster_name)?;
+
+    if let Some(p) = port {
+        validate_port(p, cluster_name)?;
+    }
+
+    Ok(())
+}
+
+/// Validate DNS label rules (RFC 1035 §2.3.1).
+///
+/// Each label must be 1-63 characters and must not start or end with a hyphen.
+fn validate_dns_labels(host: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    let hostname = host.strip_suffix('.').unwrap_or(host);
+    if hostname.is_empty() {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority hostname is empty"
+        )));
+    }
+    if hostname.len() > MAX_HOSTNAME_LEN {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority hostname exceeds {MAX_HOSTNAME_LEN} characters"
+        )));
+    }
+    for label in hostname.split('.') {
+        if label.is_empty() {
+            return Err(ProxyError::Config(format!(
+                "cluster '{cluster_name}': authority hostname contains an empty label"
+            )));
+        }
+        if label.len() > 63 {
+            return Err(ProxyError::Config(format!(
+                "cluster '{cluster_name}': authority hostname label exceeds 63 characters"
+            )));
+        }
+        if label.starts_with('-') || label.ends_with('-') {
+            return Err(ProxyError::Config(format!(
+                "cluster '{cluster_name}': authority hostname label must not start or end with a hyphen"
+            )));
+        }
+    }
+    Ok(())
+}
+
+/// Validate the port number (1..=65535).
+fn validate_port(port_str: &str, cluster_name: &str) -> Result<(), ProxyError> {
+    if port_str.is_empty() {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority has empty port"
+        )));
+    }
+    let port: u32 = port_str
+        .parse()
+        .map_err(|_e| ProxyError::Config(format!("cluster '{cluster_name}': authority has invalid port")))?;
+    if port == 0 || port > 65535 {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': authority port out of range"
+        )));
+    }
+    Ok(())
+}
+
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+#[cfg(test)]
+#[expect(clippy::allow_attributes, reason = "blanket test suppressions")]
+#[allow(
+    clippy::panic,
+    clippy::unwrap_used,
+    clippy::expect_used,
+    clippy::needless_raw_strings,
+    clippy::needless_raw_string_hashes,
+    reason = "tests use unwrap/expect/panic/raw strings for brevity"
+)]
+mod tests {
+    use super::*;
+
+    fn ok(authority: &str) {
+        validate_authority(authority, "test").unwrap_or_else(|e| panic!("expected Ok for {authority:?}, got: {e}"));
+    }
+
+    fn err(authority: &str) -> String {
+        validate_authority(authority, "test").unwrap_err().to_string()
+    }
+
+    #[test]
+    fn accept_simple_hostname() {
+        ok("api.example.com");
+    }
+
+    #[test]
+    fn accept_hostname_with_port() {
+        ok("api.example.com:443");
+    }
+
+    #[test]
+    fn accept_localhost() {
+        ok("localhost");
+    }
+
+    #[test]
+    fn accept_localhost_with_port() {
+        ok("localhost:8080");
+    }
+
+    #[test]
+    fn accept_single_label() {
+        ok("backend");
+    }
+
+    #[test]
+    fn accept_bracketed_ipv6() {
+        ok("[::1]");
+    }
+
+    #[test]
+    fn accept_bracketed_ipv6_with_port() {
+        ok("[::1]:8443");
+    }
+
+    #[test]
+    fn accept_full_ipv6() {
+        ok("[2001:db8::1]:443");
+    }
+
+    #[test]
+    fn reject_empty() {
+        assert!(err("").contains("must not be empty"));
+    }
+
+    #[test]
+    fn reject_control_char() {
+        assert!(err("api\x00.example.com").contains("control characters"));
+    }
+
+    #[test]
+    fn reject_newline() {
+        assert!(err("api\nexample.com").contains("control characters"));
+    }
+
+    #[test]
+    fn reject_space() {
+        assert!(err("api example.com").contains("whitespace"));
+    }
+
+    #[test]
+    fn reject_tab() {
+        assert!(err("api\texample.com").contains("control characters"));
+    }
+
+    #[test]
+    fn reject_scheme() {
+        assert!(err("https://api.example.com").contains("URI scheme"));
+    }
+
+    #[test]
+    fn reject_path() {
+        assert!(err("api.example.com/v1").contains("path"));
+    }
+
+    #[test]
+    fn reject_userinfo() {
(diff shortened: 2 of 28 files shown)


core/src/config/cluster/mod.rs
```diff
@@ -40,6 +42,36 @@ pub struct Cluster {
...
+    /// endpoints: ["10.0.0.1:443"]
+    /// authority: "api.example.com"
+    /// tls:
+    ///   sni: "api.example.com"
+    /// "#;
+    /// let cluster: Cluster = serde_yaml::from_str(yaml).unwrap();
+    /// assert_eq!(cluster.authority.as_deref(), Some("api.example.com"));
+    /// ```
+    #[serde(default)]
+    pub authority: Option<Arc<str>>,
```
shaneutt, line 73: The `Cluster` struct has so far been stepped in L3/L4 semantics, as it's about connectivity and transport security. I can see how the Host header would correspond to the latter, but I think it still blurs concerns.

Let's look at other options for where we can put this, and compare/constrast.

core/src/config/validate/cluster/authority.rs
```diff
@@ -0,0 +1,414 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Praxis Contributors
+
+//! HTTP authority validation for per-cluster upstream authority override.
+
+use crate::errors::ProxyError;
+
```
shaneutt, line 7: +separator-comments

core/src/config/validate/cluster/authority.rs
```diff
@@ -0,0 +1,414 @@
...
+    #[test]
+    fn accept_label_at_63_chars() {
+        let label = "a".repeat(63);
+        ok(&format!("{label}.example.com"));
+    }
+
+    #[test]
+    fn accept_hyphen_in_middle() {
+        ok("my-api.example.com");
+    }
```
shaneutt, line 413: These tests freak me out.

</img>

filter/src/builtins/http/observability/access_log.rs
```diff
@@ -434,6 +434,10 @@ mod tests {
     }
 
     #[tokio::test]
+    #[expect(
+        clippy::too_many_lines,
+        reason = "test fixture exercises the complete populated context"
```
shaneutt, line 439: Reason "tests" in fine in tests.

```suggestion
        reason = "tests"
```

shaneutt (changes requested):
Thanks @nerdalert!

I would like to consider a different place to inject the authority, LMKWYT?

nerdalert:
@shaneutt thanks for entertaining this. Did some research with GPT-5.6 SOL that gave a good even swap on some approaches and what I think is a reasonable recommendation to take a peak at your leisure.

# Assessment

I agree with Shane’s architectural concern, while also thinking the PR’s **runtime behavior is fundamentally correct**.

The real issue is not whether the feature belongs somewhere near the load balancer. It does. The issue is that the PR currently puts HTTP request semantics into two types that are presented as shared connectivity abstractions:

* `praxis_core::config::Cluster`
* `praxis_core::connectivity::Upstream`

The PR adds `authority` directly to both.

`Cluster` is consumed by both the HTTP and TCP load-balancer filters today. The TCP load balancer deserializes the same `Cluster` object even though it cannot use an HTTP authority.

That validates Shane’s concern:

> The selected authority belongs with the selected upstream service, but it should remain inside an HTTP-specific representation of that service.

I would **not** move this to `intelligent_route`, and I would **not** solve it with the generic headers filter. My recommendation is to keep the good runtime behavior but refactor the configuration and selected-upstream types around an HTTP-specific wrapper.

---

# What the current proposal gets right

The PR correctly identifies a real distinction:

```text
TCP endpoint:  api.openai.com:443
TLS SNI:       api.openai.com
HTTP authority: api.openai.com
```

Those values are related but not necessarily identical. Deriving the HTTP authority from SNI or from the selected endpoint would be wrong in cases where:

* The endpoint is an internal IP or routing layer.
* The TLS endpoint serves multiple HTTP virtual hosts.
* SNI names a shared edge while HTTP authority names a tenant.
* A provider gateway connects through an internal address to a public API origin.

The runtime implementation also has several strong properties:

* Authority is parsed once during cluster construction.
* It follows the selected cluster.
* It is retained with retry state.
* It is applied after internal and hop-by-hop headers are stripped.
* It replaces both `Host` and an absolute-form URI authority.
* Clusters without the setting preserve existing behavior.

Those behaviors should be preserved regardless of where the configuration field ultimately lives.

---

# One nuance about “L3/L4”

I would not argue too literally about the OSI labels.

The existing `Cluster` is not completely protocol-neutral already: its health-check configuration supports HTTP paths and expected status codes, in addition to TCP and future gRPC probes.

However, there is still an important distinction:

* An HTTP health check is an **endpoint eligibility mechanism** attached to cluster management.
* Rewriting authority changes the **identity of the live application request**.

So Shane’s separation concern remains valid even though `Cluster` is not mathematically pure L3/L4.

The better framing is:

> `Cluster` is currently shared HTTP/TCP connectivity and endpoint-management configuration. Authority override is per-request HTTP behavior.

---

# Compare and contrast

| Placement                                   | Selected-cluster aware | Works with `intelligent_route` |     Retry safe |  HTTP/TCP separation | Config duplication | Assessment                                      |
| ------------------------------------------- | ---------------------: | -----------------------------: | -------------: | -------------------: | -----------------: | ----------------------------------------------- |
| Current `Cluster.authority`                 |                    Yes |                            Yes |            Yes |                 Poor |               None | Correct behavior, wrong shared type             |
| `Cluster.http.authority`                    |                    Yes |                            Yes |            Yes | Better, not complete |               None | Reasonable pragmatic compromise                 |
| HTTP load-balancer cluster wrapper          |                    Yes |                            Yes |            Yes |               Strong |               None | **Recommended**                                 |
| Dedicated authority filter                  |       Yes, if after LB |                            Yes |         Can be |               Strong |        Usually yes | Viable second choice                            |
| Route-level rewrite                         |  Yes for static routes |                        Awkward |            Yes |               Strong |   Potentially high | Good generally, weak for dynamic AI routing     |
| Generic headers filter                      |          Not naturally |                           Poor | Not inherently |               Strong |               High | Insufficient                                    |
| Endpoint-level authority                    |                    Yes |                            Yes |            Yes |               Strong |               None | Wrong default abstraction                       |
| Derive from TLS SNI                         |                    Yes |                            Yes |            Yes |                Mixed |               None | Semantically incorrect                          |
| Derive from endpoint hostname               |                    Yes |                            Yes |            Yes |                Mixed |               None | Fails for IP/internal endpoints                 |
| Generic typed protocol options on `Cluster` |                    Yes |                            Yes |            Yes |    Good if validated |               None | Stronger long-term abstraction, larger refactor |

## 1. Keep `Cluster.authority`

### Advantages

* Smallest conceptual configuration.
* Authority naturally follows cluster selection.
* Works with router, `intelligent_route`, retries and failover.
* No second cluster-to-authority lookup.
* Current implementation is already tested.

### Problems

* TCP cluster configuration accepts an HTTP-only field.
* Generic `Upstream` now contains `HeaderValue`.
* Future HTTP options will likely accumulate at the top level:

```yaml
authority:
http_version:
path_prefix:
request_headers:
grpc:
```

* A TCP caller can construct a `Cluster` with meaningful-looking HTTP fields that are silently ignored.
* It weakens crate boundaries by putting the `http` crate’s header representation into connectivity state.

### Verdict

I would not keep the current top-level placement.

---

## 2. Nest it under `Cluster.http`

```yaml
clusters:
  - name: openai-api
    endpoints:
      - api.openai.com:443
    tls:
      sni: api.openai.com
      verify: true
    http:
      authority: api.openai.com
```

### Advantages

* Makes the protocol boundary visible.
* Maintains one logical cluster object.
* Retains selected-cluster and retry behavior.
* Provides room for future upstream HTTP settings.
* Similar in spirit to Envoy’s namespaced upstream HTTP protocol options within cluster configuration. Envoy keeps protocol-specific cluster settings in a typed HTTP options extension rather than mixing every field into the cluster’s top level. ([Envoy Proxy][1])

### Problems

* The core shared `Cluster` still owns L7 settings.
* TCP must either reject `http` or silently ignore it.
* The generic `Upstream` runtime type still needs to be fixed.
* It can become a miscellaneous bucket unless its scope is narrowly defined.

### Verdict

A respectable compromise, particularly if Praxis wants `Cluster` to evolve into a general logical upstream service definition.

It does not fully address Shane’s concern, but it is materially better than a top-level field.

---

## 3. HTTP-specific load-balancer cluster wrapper

```yaml
filter: load_balancer
clusters:
  - name: openai-api
    endpoints:
      - api.openai.com:443
    tls:
      sni: api.openai.com
      verify: true
    http:
      authority: api.openai.com
```

Conceptually:

```rust
struct HttpClusterConfig {
    cluster: Cluster,
    http: HttpUpstreamConfig,
}

struct HttpUpstreamConfig {
    authority: Option<ValidatedAuthority>,
}
```

The HTTP load balancer already has its own config wrapper and `ClusterEntry`, separate from the TCP load balancer. That makes it a natural place to compile HTTP-specific per-cluster state.

The runtime representation should follow the same separation:


```rust
struct HttpUpstream {
    transport: Upstream,
    request: HttpUpstreamRequestPolicy,
}

struct HttpUpstreamRequestPolicy {
    authority: Option<ValidatedAuthority>,
}
```

Then:

```text
Cluster
  -> address, TLS, connection options

HttpClusterConfig
  -> Cluster + HTTP request policy

HttpUpstream
  -> selected transport upstream + selected HTTP request policy
```

### Advantages

* Strong HTTP/TCP separation.
* Authority remains tied to the dynamically selected cluster.
* No separate lookup table.
* Retry state can clone one complete `HttpUpstream`.
* TCP never sees or silently ignores the option.
* `intelligent_route` remains responsible only for selecting a cluster.
* The protocol handler can keep the current correct application logic.
* Creates an extensible home for genuinely upstream-service-wide HTTP options.

Traefik uses a similar conceptual boundary: `passHostHeader` belongs to its HTTP service load-balancer model, while HTTP and TCP services are separate configuration concepts. ([Traefik Docs][2])

### Problems

* Requires a small type and config refactor.
* The public `LoadBalancerFilter::new(&[Cluster])` API needs a compatibility path.
* Shared parsing/schema code becomes slightly more involved.
* Care is needed with Serde flattening and `deny_unknown_fields`.

### Verdict

**Best match for Praxis’s current architecture and the Grid use case.**

This is my recommendation.

---

## 4. Dedicated `upstream_authority` filter

```yaml
filters:
  - filter: intelligent_route

  - filter: load_balancer
    clusters:
      - name: openai-api
        endpoints:
          - api.openai.com:443

  - filter: upstream_authority
    clusters:
      openai-api: api.openai.com
```

The filter would run after `load_balancer`, inspect the selected cluster or selected upstream, and store a validated HTTP request policy for the protocol handler.

### Advantages

* Very clean L7 ownership.
* Reusable independently from the load balancer.
* Fits Praxis’s composable-filter philosophy.
* Core connectivity types stay clean.
* Can eventually support authority selected from trusted metadata.

### Problems

* Duplicates cluster names and configuration.
* Requires strict filter ordering.
* A missing filter silently changes upstream behavior unless pipeline validation catches it.
* Retry state must snapshot the filter’s selected value.
* The same cluster can behave differently depending on which chain happened to include the filter.
* A configuration intended to describe one logical upstream is split across two filters.

This is viable if the project has a strong rule that all request mutation must be a filter. I still prefer the HTTP cluster wrapper because authority is a stable property of the logical upstream in this use case, not a general per-route transformation.

---

## 5. Route-level authority rewrite

```yaml
routes:
  - path_prefix: /
    cluster: openai-api
    rewrite:
      authority: api.openai.com
```

This is where several other systems place host rewriting:

* Gateway API exposes hostname rewriting through the HTTPRoute `URLRewrite` filter. ([Gateway API][3])
* Envoy’s host rewrite controls are part of its HTTP route action rather than its base cluster definition. ([Envoy Gateway][4])
* NGINX configures `proxy_set_header Host` within HTTP `http`, `server`, or `location` proxy configuration rather than its upstream server group. ([Nginx][5])
* HAProxy models Host replacement as an `http-request` rule that can be applied in an HTTP backend. ([HAProxy Technologies][6])

### Advantages

* Clearly L7.
* Allows the same endpoint pool to serve different authorities on different routes.
* Aligns well with common proxy APIs.
* Makes authority a request-routing action rather than connectivity state.

### Problems for Praxis/Grid

Praxis’s current route object names one static target cluster.

Grid’s path is different:

```text
classify request
    -> intelligent_route chooses one of many clusters dynamically
    -> load_balancer selects endpoint
```

There may be no static route corresponding to the final cluster. Putting authority on the route would either:

* Duplicate it across many routes.
* Require `intelligent_route` candidates to carry HTTP rewrite details.
* Require another cluster-to-authority lookup after dynamic selection.

### Verdict

Good for conventional static routing and perhaps worth supporting eventually as an override. Not the best primary mechanism for Grid-selected clusters.

A future precedence model could be:

```text
explicit route override
    > HTTP cluster default
    > preserve downstream authority
```

That does not need to be part of this PR.

---

## 6. Generic headers filter

Praxis’s current headers filter queues static request header additions, removals and replacements.

```yaml
filter: headers
request_set:
  - name: Host
    value: api.openai.com
```

### Advantages

* Already exists.
* Clearly HTTP/L7.
* Easy for one static upstream.

### Problems

* Static for the entire filter instance.
* Not naturally tied to a cluster selected later.
* Requires branch chains or duplicated filters for multiple clusters.
* Only changes `Host`; it does not inherently normalize absolute-form URI authority.
* Ordering relative to cluster selection becomes critical.
* It is easy for generic header configuration and special authority semantics to diverge.

### Verdict

Not sufficient as currently designed.

Extending it until it becomes selected-cluster-aware and URI-aware would effectively recreate a dedicated authority feature under a less precise name.

---

## 7. Endpoint-level authority

```yaml
endpoints:
  - address: 10.0.0.1:443
    authority: api-a.example.com
  - address: 10.0.0.2:443
    authority: api-b.example.com
```

### Advantages

* Supports endpoints with genuinely different HTTP virtual hosts.
* Naturally follows endpoint retries.

### Problems

* Authority usually identifies the logical origin/service, not one IP.
* Duplicates the same authority across every endpoint.
* Makes endpoint rotation or discovery more complicated.
* Encourages treating address identity and application origin as the same thing.

Envoy allows an endpoint hostname to participate in automatic host rewriting, showing that endpoint-derived behavior can be useful, but that is an explicit mode rather than the universal representation. ([Envoy Proxy][7])

### Verdict

Possibly useful as a future advanced override, but wrong as the default configuration model.

---

# Validation should also be simplified

The current PR adds a 414-line custom authority validator and roughly three dozen individual parsing tests.

I suspect this is part of what Shane meant when he said the tests “freak me out.”

The validator is not merely validating HTTP syntax. It defines a narrower Praxis policy:

* ASCII DNS names.
* Bracketed IPv6.
* Strict DNS-label lengths.
* No underscore.
* No general URI registered-name characters.

RFC 9110 defines `Host` as `uri-host` with an optional port, and the referenced URI host grammar includes IP literals, IPv4 and registered names. ([RFC Editor][8])

A smaller implementation would use a typed newtype around the existing HTTP parser:

```rust
struct ValidatedAuthority {
    authority: http::uri::Authority,
    header_value: HeaderValue,
}
```

Construction could:

1. Parse with `http::uri::Authority`.
2. Reject userinfo explicitly.
3. Apply any Praxis-specific port policy.
4. Build and cache `HeaderValue`.
5. Preserve the parsed `Authority` for URI reconstruction.

The Rust `http` crate already provides a dedicated `Authority` representation and parser. ([Docs.rs][9])

That would substantially reduce:

* Custom parser code.
* RFC interpretation burden.
* Security-sensitive edge cases.
* The number of bespoke tests.
* The PR’s size.

When replacing the URI authority, use URI parts rather than string formatting:

```rust
let mut parts = req.uri.clone().into_parts();
parts.authority = Some(validated.authority.clone());
req.set_uri(Uri::from_parts(parts)?);
```

That avoids inventing a default scheme and preserves the original URI representation more directly.

---

# Recommended design

```yaml
filter: load_balancer
clusters:
  - name: openai-api
    endpoints:
      - api.openai.com:443
    tls:
      sni: api.openai.com
      verify: true
    http:
      authority: api.openai.com
```

Conceptually:

```rust
// Shared HTTP/TCP connectivity definition.
struct Cluster {
    name: Arc<str>,
    endpoints: Vec<Endpoint>,
    connection_timeout_ms: Option<u64>,
    tls: Option<ClusterTls>,
    load_balancer_strategy: LoadBalancerStrategy,
    // ...
}

// Used only by the HTTP load_balancer.
struct HttpClusterConfig {
    cluster: Cluster,
    http: HttpUpstreamConfig,
}

struct HttpUpstreamConfig {
    authority: Option<ValidatedAuthority>,
}

// Selected HTTP request target.
struct HttpUpstream {
    transport: Upstream,
    request_policy: HttpUpstreamRequestPolicy,
}

struct HttpUpstreamRequestPolicy {
    authority: Option<ValidatedAuthority>,
}
```

Runtime:

```text
intelligent_route
    -> sets ctx.cluster

HTTP load_balancer
    -> selects ClusterEntry
    -> selects endpoint
    -> creates HttpUpstream {
           transport,
           request_policy
       }

Pingora protocol handler
    -> builds/reuses transport peer
    -> applies HTTP authority policy
    -> retains complete HttpUpstream for retries
```

This keeps all of the current PR’s useful runtime behavior while restoring the HTTP/TCP boundary.

---

## Conclusion/Recommendation

  The authority must follow the dynamically selected cluster, but it is HTTP request semantics rather than shared connectivity or transport state. Adding it directly to the shared Cluster and Upstream types blurs the existing L3/L4 boundary and exposes an HTTP-only option to TCP consumers.

  The alternatives have meaningful drawbacks:

  - Generic headers filter: Clearly L7, but static for the filter chain. It is not naturally associated with the cluster selected by router or intelligent_route, and rewriting Host alone does not normalize an absolute-form URI authority.

  - Route-level configuration: Clean for static routes, but unsuitable when intelligent_route chooses the final cluster dynamically. It would duplicate authority configuration or push HTTP behavior into routing candidates.

  - Dedicated post-load-balancer filter: Preserves separation, but introduces filter-order coupling and requires another cluster-to-authority mapping.
  - Endpoint-level configuration: Would follow endpoint retries, but authority identifies the logical upstream service, not usually an individual network address.
  - Cluster.http on the shared type: Makes the intent clearer, but the shared HTTP/TCP Cluster would still own L7 behavior.

  The preferred design is an HTTP-specific load-balancer cluster wrapper:

```yaml
  clusters:
    - name: openai-api
      endpoints:
        - api.openai.com:443
      tls:
        sni: api.openai.com
        verify: true
      http:
        authority: api.openai.com

  Conceptually:

  struct HttpClusterConfig {
      #[serde(flatten)]
      cluster: Cluster,

      #[serde(default)]
      http: HttpUpstreamConfig,
  }

  struct HttpUpstreamConfig {
      authority: Option<ValidatedAuthority>,
  }
```

  The shared Cluster and transport Upstream remain HTTP-agnostic. The HTTP load balancer validates and compiles the authority when loading configuration, then associates it with the selected cluster in HTTP-specific request state. The protocol handler applies that policy when constructing the upstream request.

  This preserves the required behavior:

  - Authority follows the dynamically selected cluster.
  - Retries retain the selected logical authority.
  - Parsing and validation happen during configuration loading.
  - The downstream Host header is replaced.
  - Absolute-form URI authority is normalized consistently.
  - TLS SNI remains independently configurable.
  - intelligent_route remains unaware of HTTP request mutation.
  - TCP configuration never encounters or silently ignores an HTTP-only field.

  The validator can also be simplified by wrapping `http::uri::Authority` and caching its corresponding HeaderValue, with only the additional Praxis-specific restrictions enforced explicitly. Those restrictions should document decisions around userinfo, empty hosts, port zero, and whether Praxis permits the full registered-name syntax accepted by URI authority parsing or intentionally limits values to DNS-style hostnames.

  This keeps cluster selection and request mutation connected without making shared connectivity types responsible for L7 behavior. The HTTP-specific load-balancer wrapper is  therefore the strongest boundary among the considered options.

## Research Refs

- https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto.html "Cluster configuration (proto) — envoy 1.40.0-dev-e4532b documentation"
- https://gateway-api.sigs.k8s.io/guides/user-guides/http-redirect-rewrite/ "HTTP path redirects and rewrites | Gateway API"
- https://gateway.envoyproxy.io/latest/api/extension_types/ "Gateway API Extensions | Envoy Gateway"
- https://nginx.org/en/docs/http/ngx_http_proxy_module.html "Module ngx_http_proxy_module"
- https://www.haproxy.com/documentation/haproxy-configuration-tutorials/proxying-essentials/http-rewrites/ "HTTP rewrites | HAProxy config tutorials"
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto "Endpoints (proto) — envoy 1.40.0-dev-fd5261 documentation"
- https://www.rfc-editor.org/rfc/inline-errata/rfc9110.html "rfc9110"
- https://docs.rs/http/latest/http/uri/struct.Authority.html "Authority in http::uri - Rust"

shaneutt:
Thank you for your contribution! This has been merged into `main` as part of a bulk merge, see: https://github.com/praxis-proxy/praxis/commit/ace2e13ae6322e51ad23aa315f7342c10b163095

PR #906, opened 2026-08-03
feat(core): OTLP exporter configuration

### LINKED ISSUE #299: OTLP exporter configuration
OTLP (OpenTelemetry Protocol) exporter with configurable endpoint, batching, and TLS.

## Requirements

- OTLP gRPC exporter endpoint configuration
- Optional OTLP HTTP exporter as alternative
- Configurable headers (e.g. API keys for hosted backends)
- Batch size and export interval configuration
- TLS options for exporter connection
- Resource attributes: `service.name`, `service.version`, `deployment.environment`

## Implementation Notes

- Use `opentelemetry-otlp` crate
- Sensible defaults: 512 batch size, 5s export interval
- Graceful shutdown: flush pending spans on proxy shutdown

Parent epic: #10

Ladas:
## What
Extend TelemetryConfig with configurable OTLP exporter settings: headers, batch size/interval, resource attributes (service.name, service.version, environment).

## Why
#315 hardcodes `service.name = "praxis"` and uses default batch config. Production deployments need custom headers (API keys), tuned batch sizes, and proper resource attributes for trace identification.

## How
- Add 6 new fields to TelemetryConfig (all Optional with serde defaults)
- Env var fallbacks per OTel spec: OTEL_SERVICE_NAME, OTEL_RESOURCE_ATTRIBUTES
- Refactor build_otel_provider into focused functions: build_span_exporter, build_batch_processor, build_otel_resource
- 20 new tests

**Depends on** #315 (PR #836)

Closes #299

### DIFF
--- a/core/src/config/telemetry.rs
+++ b/core/src/config/telemetry.rs
@@ -3,14 +3,30 @@
 
 //! OpenTelemetry and distributed tracing configuration.
 
+use std::{collections::HashMap, fmt};
+
 use serde::{Deserialize, Serialize};
 
 // -----------------------------------------------------------------------------
 // Constants
 // -----------------------------------------------------------------------------
 
+/// OTel-standard environment variable for resource attributes
+/// (format: `key1=value1,key2=value2`).
+const OTEL_RESOURCE_ATTRIBUTES_ENV_VAR: &str = "OTEL_RESOURCE_ATTRIBUTES";
+
+/// OTel-standard environment variable for the service name.
+const OTEL_SERVICE_NAME_ENV_VAR: &str = "OTEL_SERVICE_NAME";
+
 /// OTel-standard environment variable for the OTLP exporter endpoint.
 const OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
+
+/// Standard `OTel` environment variable for OTLP exporter headers.
+const OTLP_HEADERS_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_HEADERS";
+
+/// OTel-standard environment variable for the OTLP exporter protocol.
+#[cfg(feature = "otel")]
+pub(crate) const OTLP_PROTOCOL_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_PROTOCOL";
 
 // -----------------------------------------------------------------------------
 // TelemetryConfig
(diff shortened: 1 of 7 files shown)


shaneutt:
conflict resolution and some comments resolved here @Ladas :vulcan_salute:

core/src/logging.rs
```diff
@@ -226,6 +222,116 @@ fn build_otel_provider(
...
+/// Build the OTLP span exporter with endpoint and optional gRPC headers.
+#[cfg(feature = "otel")]
+fn build_span_exporter(
+    endpoint: &str,
+    headers: &Option<std::collections::HashMap<String, String>>,
+) -> Result<opentelemetry_otlp::SpanExporter, ProxyError> {
+    use opentelemetry_otlp::{WithExportConfig as _, WithTonicConfig as _};
+
+    let mut builder = opentelemetry_otlp::SpanExporter::builder()
+        .with_tonic()
```
nerdalert, line 257: This always selects OTLP over gRPC. However, the configuration uses the standard `OTEL_EXPORTER_OTLP_ENDPOINT` variable, which users may set to either a gRPC or HTTP/protobuf collector endpoint. If they provide an HTTP endpoint, Praxis may start normally but fail to export traces.

Could we either honor `OTEL_EXPORTER_OTLP_PROTOCOL` and select the matching exporter, or explicitly support only `grpc`, document that limitation, and reject any other configured protocol during startup? The important part is to fail clearly during configuration instead of discovering the mismatch later when traces do not appear.

core/src/config/telemetry.rs
```diff
@@ -36,41 +45,163 @@ const OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
...
     ///
     /// When set, enables the OTLP trace exporter (requires the `otel`
     /// feature). Falls back to `OTEL_EXPORTER_OTLP_ENDPOINT` env var
     /// if not set in config.
     pub otlp_endpoint: Option<String>,
+
+    /// Custom headers for the OTLP exporter (e.g. API keys).
+    ///
+    /// Sent as gRPC metadata on every export request.
+    pub otlp_headers: Option<HashMap<String, String>>,
```
nerdalert, line 88: These headers are expected to contain API keys and authorization tokens, but `TelemetryConfig` derives `Debug` and `Serialize`, so their values could be exposed through diagnostics or configuration dumps. The current design also encourages putting credentials directly in YAML because it does not read the standard `OTEL_EXPORTER_OTLP_HEADERS` environment variable.

Could we support `OTEL_EXPORTER_OTLP_HEADERS` for production credentials and ensure debug output redacts all header values? File-based headers can remain supported, but their secret values should never appear in logs or diagnostics.

core/src/config/telemetry.rs
```diff
@@ -36,41 +45,163 @@ const OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
...
+    let attrs = std::env::var(OTEL_RESOURCE_ATTRIBUTES_ENV_VAR).ok()?;
+    extract_resource_attribute(&attrs, key)
+}
+
+/// Extract a single attribute value from an `OTel` resource-attributes string.
+///
+/// Parses the `key1=value1,key2=value2` format used by
+/// `OTEL_RESOURCE_ATTRIBUTES`. Returns the trimmed value for the first
+/// entry whose trimmed key matches `target_key`.
+fn extract_resource_attribute(attrs: &str, target_key: &str) -> Option<String> {
```
nerdalert, line 248: The OpenTelemetry Rust SDK already provides `opentelemetry_sdk::resource::EnvResourceDetector` for reading `OTEL_RESOURCE_ATTRIBUTES`. Could we use that here instead of maintaining a partial parser?

The SDK detector handles the OpenTelemetry format, including percent-encoded commas and equals signs and malformed input. This implementation currently splits the string but leaves encoded values such as `%2C` unchanged. Explicit YAML values can still be layered over the detected resource so configuration keeps precedence.

core/src/config/telemetry.rs
Ladas: - Removed Debug from derive, added custom impl fmt::Debug that shows otlp_headers: Some("<2 header(s) redacted>") instead of actual values
- Added OTEL_EXPORTER_OTLP_HEADERS env var support in resolve() — parses key1=value1,key2=value2 format. Config values take precedence over env var
- #[serde(skip_serializing)] was already on the field (prevents JSON/YAML serialization)

core/src/logging.rs
Ladas: yes, this is still a gap, let me extend it to support http and add tests

core/src/config/telemetry.rs
Ladas: Changed build_otel_resource() to use ResourceBuilder::with_detector(Box::new(EnvResourceDetector::new())). Now all OTEL_RESOURCE_ATTRIBUTES flow into the OTel Resource — not just the 3 we manually extracted. Config values are layered on top via .with_attribute() so they still take precedence.

nerdalert:
@Ladas once the batch queue-size bot comment is resolved we are good to go, ty!! 🙏

nerdalert (approved):
LGTM, thanks @Ladas 

I think you said https://github.com/praxis-proxy/praxis/pull/906#discussion_r3769404141 was a false positive from bot. If you don't mind resolving it if so. Ty!

Issue #301, opened 2026-05-15
Root span per request with standard attributes


shaneutt:
Create a root span for each request with standard HTTP semantic convention attributes.

## Requirements

- Root span created at request ingress
- Attributes: HTTP method, path, status code, client IP, request ID
- Attributes: upstream cluster name, upstream address
- Span name: `{method} {path}` (e.g. `GET /api/v1/models`)
- Span kind: SERVER
- Duration covers full request lifecycle
- Request ID must be a recorded span field (not just
  an attribute) so it propagates into every `tracing`
  log line emitted within the span tree — filter
  execution, upstream connect, error events, etc.
  This enables grep-by-request-ID across all log
  output, not just access logs.

## Implementation Notes

- OpenTelemetry HTTP semantic conventions for attribute names
- Request ID from `RequestIdFilter` output

Parent epic: #10

PR #907, opened 2026-08-03
feat(core): configurable head-based trace sampling

### LINKED ISSUE #311: Configurable head-based sampling with always-sample-on-error

Ladas:
## What
Add sampling_rate field to TelemetryConfig (0.0 to 1.0). When set, configures a ParentBased sampler wrapping TraceIdRatioBased on the SdkTracerProvider.

## Why
Production deployments need to control trace volume. AlwaysOn sampling generates too much data; configurable rates (1-10%) with always-on-error preserve visibility while managing cost.

## How
- New `sampling_rate: Option<f64>` on TelemetryConfig
- Validation at config load (reject < 0.0 or > 1.0)
- ParentBased(TraceIdRatioBased) when rate is set
- TODO: always-sample-on-error requires #301 for error detection at span close
- 15 new tests

**Depends on** #315 (PR #836)

Closes #311

### DIFF
--- a/core/src/config/telemetry.rs
+++ b/core/src/config/telemetry.rs
@@ -28,6 +28,7 @@ const OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
 ///
 /// let telemetry = TelemetryConfig::default();
 /// assert!(telemetry.otlp_endpoint.is_none());
+/// assert!(telemetry.sampling_rate.is_none());
 ///
 /// let telemetry: TelemetryConfig =
 ///     serde_yaml::from_str("otlp_endpoint: \"http://localhost:4317\"").unwrap();
@@ -36,7 +37,7 @@ const OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
 ///     Some("http://localhost:4317")
 /// );
 /// ```
-#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct TelemetryConfig {
     /// OTLP collector endpoint (e.g. `http://localhost:4317`).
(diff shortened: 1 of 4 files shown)


core/src/config/telemetry.rs
```diff
@@ -45,6 +46,19 @@ pub struct TelemetryConfig {
...
+    /// Head-based trace sampling rate between `0.0` (drop all) and `1.0`
+    /// (sample all).
+    ///
+    /// When set, configures a `ParentBased(TraceIdRatioBased(rate))`
+    /// sampler: root spans are sampled at the given rate while child
+    /// spans respect the parent's sampling decision (W3C `traceparent`).
+    ///
+    /// When `None` (the default), the `OTel` default sampler is used
+    /// (`ParentBased(AlwaysOn)`), preserving backward compatibility.
+    // TODO(#311): always-sample-on-error requires #301 (root span) to
```
nerdalert: The head-sampling implementation looks fine, but it cannot guarantee that errors are sampled because the decision happens before the response is known.

Please rename the PR to `feat(core): configurable head-based trace sampling` and remove the "always sample on error" wording and TODO. The sampling code itself does not need to change.

core/src/config/telemetry.rs
```diff
@@ -45,6 +46,19 @@ pub struct TelemetryConfig {
     /// feature). Falls back to `OTEL_EXPORTER_OTLP_ENDPOINT` env var
     /// if not set in config.
     pub otlp_endpoint: Option<String>,
+
+    /// Head-based trace sampling rate between `0.0` (drop all) and `1.0`
+    /// (sample all).
+    ///
+    /// When set, configures a `ParentBased(TraceIdRatioBased(rate))`
+    /// sampler: root spans are sampled at the given rate while child
+    /// spans respect the parent's sampling decision (W3C `traceparent`).
```
nerdalert: `ParentBased` preserves a parent decision only when an OpenTelemetry parent context has actually been attached to the span. Registering `TraceContextPropagator` alone does not extract an inbound `traceparent`, and Praxis does not currently perform that extraction.

Could we describe this as local parent/child inheritance for now and defer the W3C remote-parent claim until propagation is implemented?

Suggested change here:

```diff
-    /// spans respect the parent's sampling decision (W3C `traceparent`).
+    /// locally-created child spans inherit their parent's sampling decision.
```

Apply the equivalent wording to the comment in `core/src/logging.rs`. The unused `set_text_map_propagator(...)` call should also remain deferred to the propagation PR, where extraction and injection are added together.

nerdalert:
Tiny nits and then it looks like this PR needs to be rebased onto #906 after it merges. Ty!

nerdalert (approved):

PR #909, opened 2026-08-03
ci: add experimental-features job with cargo-hack

Ladas:
## What

Add an `experimental-features` job to the Tests workflow that
verifies non-default feature combinations lint and pass tests
using `cargo-hack --each-feature`.

## Why

Code behind `#[cfg(feature = "...")]` is invisible to the default
CI pipeline — the compiler and clippy never see it. A regression
in any feature-gated path goes undetected until someone manually
runs `cargo clippy --features otel`.

## How

Two steps, each using `cargo-hack --each-feature`:

1. **Clippy** — lint each non-default feature in isolation
2. **Test** — run tests per non-default feature

Scoped to the 4 crates that have non-default features:
- `praxis-proxy-core` (otel)
- `praxis-proxy-filter` (basic-auth-filter, cpex-policy-engine)
- `praxis-proxy-tls` (hot-reload)
- `praxis-proxy` (otel, dev, basic-auth-filter, cpex-policy-engine, experimental)

Flags:
- `--exclude-no-default-features` skips the zero-features run
  (already covered by `make test-unit`)
- `--exclude-features default` skips the default-features run
  (also already covered)
- Result: only non-default features are tested, no redundant work

Non-voting (`continue-on-error: true`) until all feature-gated
paths stabilize.

Uses `cargo install cargo-hack --locked` with `actions/cache`,
matching the existing cargo-machete pattern in the lint job.

## Review Response

| Finding | Response |
|---------|----------|
| Critical: otel feature missing on main | Resolved — #836 merged |
| SHA pin mismatch | Fixed — uses same SHA as existing jobs |
| Test only core crate | Fixed — scoped to all 4 crates with non-default features |
| Use generic approach | Fixed — cargo-hack auto-discovers features per crate |
| h2spec failure from conformance crate | Fixed — scoped with -p, no longer tests irrelevant crates |
| Redundant check step | Removed — cargo test already compiles |


shaneutt (changes requested):
Thanks @Ladas 🫡 

Is this something we could change from "Otel testing" to "experimental feature testing" and cover ALL flags that aren't on by default?

Ladas:
@shaneutt ok, let me look if this can be nicely made generic

Ladas:
@shaneutt  Tests / experimental-features tests 9 crate+feature combinations (only the non default) as lint+tests and takes 19 minutes. So maybe we should split it per crate? And maybe splitting the non-default lint as another check too?

And do we want this as required check? All seems to be passing, so it can be required. (right now I added it as optional, so failures don't block PR merge)

Experimenting with parallel run per crate in https://github.com/praxis-proxy/praxis/pull/942, this PR can be closed if that will work. (so we had it much faster run)

shaneutt (approved):

Ladas:
@shaneutt  replaced by https://github.com/praxis-proxy/praxis/pull/942, that do feature tests per crate in parallel

PR #912, opened 2026-08-04
fix(policy): honor route-scoped identity inside policy engine.

terylt:
## What does this PR do?

The `policy` security filter invoked the `identity.resolve` hook with
`Extensions::default()`, so the policy engine never received the request's route coordinates
(`entity_type` / `entity_name`). Without them the identity hook can't tell which
route it is resolving for, so it runs **every** registered identity resolver
instead of the route's own `authentication:` list — route-scoped identity
**silently no-op'd**. This is the fix: pass the route coordinates so the hook
scopes resolution to the route.

This stamps a `MetaExtension { entity_type, entity_name }` onto the `Extensions`
passed to the hook, at both resolution sites:

- **HTTP authz phase** — `ENTITY_HTTP` / `ENTITY_NAME_GLOBAL`
- **MCP body phase** — the matched route's `entity_type` / `entity_name`

## Why

Route-scoped `authentication:` is a supported CPEX config surface, but praxis
wasn't handing the policy engine the route identity, so it fell back to global resolution. A
route that authenticates solely by, say, a workload SVID —

```yaml
authentication:
  replace_inherited: true
  steps: [jwt-workload]
```

— had its scoping ignored, and every registered resolver ran anyway.

## Tests

Three unit tests in `security/policy/tests.rs`:

- `route_authentication_scopes_identity_to_its_resolver` — the route's resolver runs for that route
- `route_replace_inherited_excludes_global_resolver` — `replace_inherited` drops the inherited global resolver
- `route_without_authentication_inherits_global_resolver` — routes with no block still inherit the global resolver

## Notes

- **No new dependencies** — imports `MetaExtension` from the existing `cpex` dep.
- `#[expect(clippy::large_stack_frames, …)]` on `resolve_identity` (async handler over the large CMF/pipeline types).
- Full `make lint` clean (clippy `-D warnings`, nightly `fmt --check`, `cargo machete`, example-test-coverage, README sync, filter-docs); DCO signed off.

### Checklist

- [x] Tests added or updated

No, it shouldn't.  This wires up the routing extensions in the policy engine.

Note:  This will enable the identity/delegation use cases in https://github.com/contextforge-org/cpex/pull/131.  Updates to demos to reflect these will be next.


shaneutt (approved):
Bot comments need to be resolved, once they are LGTM.

araujof (approved):
LGTM

The bot suggestions seem pretty minor.

PR #943, opened 2026-08-09
feat: move the policy filter onto the Praxis Policy Engine; hot reload referenced documents

### LINKED ISSUE #900: Config watcher ignores files referenced by the config, so filter-owned documents never hot-reload
## Summary

The config watcher reloads on changes to `praxis.yaml` itself, but never to files that `praxis.yaml` *references*. Filters that load an external document (a policy file, a rule set, anything behind a `config_path`-style key) therefore never pick up edits to that document, even though the operator's mental model is "edit config, it reloads".

Worse, the failure is silent and asymmetric: if the referenced file lives in the same directory as `praxis.yaml`, the watcher **does** wake up, then suppresses the reload because a *different* file's hash is unchanged.

**Version:** `praxis-proxy` at rev `ed46eb5347d99b7aaf1fe67fa40f8c9178b7aa88`.

## Root cause

Two things combine, both in `server/src/watcher.rs`.

**1. The watch is directory-scoped, but the read is single-file.** `setup_watcher` watches the *directory* containing the config, non-recursively:

```rust
watcher.watch(watch_dir, RecursiveMode::NonRecursive)?;
```

So a sibling file's modification does fire an event. Referenced files in any *other* directory are not observed at all.

**2. The reload is gated on the hash of `config_path` alone.** `handle_reload` reads only the main config and returns early when its content is unchanged (`server/src/watcher.rs:210-227`):

```rust
let content = match std::fs::read_to_string(config_path) { ... };

let new_hash = hash_content(&content);
if new_hash == *content_hash {
    tracing::debug!("config file content unchanged, skipping reload");
    return true;
}
```

A referenced document is never read and never hashed. So editing only that document leaves `praxis.yaml` byte-identical, the guard trips, and the pipeline is never rebuilt — the filter keeps serving the document it loaded at startup. The `debug!` is the only trace, and it says the config is unchanged, which is true but misleading: something the config *depends on* did change.

## Impact

Any filter whose behaviour is defined by an external file is effectively frozen until something unrelated perturbs `praxis.yaml` or the process restarts. For policy documents this is the difference between "policy updated" and "policy silently still the old one", which is a security-relevant failure mode rather than a convenience one — the operator has positive confirmation the file on disk changed, and no signal that the running process disagrees.

## Reproduction

1. Configure a filter that reads an external document via its own config key.
2. Start the server; confirm the filter's behaviour reflects the document.
3. Edit **only** the referenced document, leaving `praxis.yaml` byte-identical.
4. Behaviour is unchanged. With the document in the same directory you can see the event arrive and then `config file content unchanged, skipping reload`.
5. Touch `praxis.yaml` in any way that alters its bytes; the rebuild now picks up the new document.

## Workaround (for others hitting this)

We embed a SHA-256 digest of the referenced documents as a comment line in the generated `praxis.yaml`:

```
# cpex-documents-sha256: <hex>
```

Any change to a referenced document changes the digest, which changes `praxis.yaml`'s bytes, which defeats the hash gate and forces the rebuild — while genuinely no-op renders stay byte-identical and correctly do **not** trigger a spurious reload. It works, but it requires whoever writes `praxis.yaml` to know about the limitation and to be in a position to inject a comment, which is not true for hand-authored configs.

## Possible directions

1. **Let filters declare dependent paths.** A filter that loads a file could register it, and the watcher would include those paths in both the watch set and the change detection. Most precise, needs a small trait/registration surface.
2. **Hash the transitive set rather than one file.** Keep the current shape but fold referenced files into the hash, so the dedupe question becomes "did anything the config depends on change?".
3. **An explicit reload trigger.** A signal or admin endpoint to force a rebuild regardless of hashing. This overlaps #785 (control-plane config push) and the closed #129, and would also serve operators who keep referenced files outside the config directory.

Option 2 is the smallest change that removes the silent-staleness; option 1 is the one that also fixes the "referenced file is in another directory" case.

## Related

- #11 Dynamic Configuration Reloading (closed) — established the reload contract this gap quietly breaks.
- #183 Config file watcher fails with bare filename (closed) — same `watch_dir_for_path` area.
- #785 Spike: gRPC/HTTP control plane API for runtime config push — an explicit trigger would subsume direction 3.

araujof:
Points the `policy` filter at the new [praxis-proxy/policy](https://github.com/praxis-proxy/policy) crates instead of the external `cpex` dependency, lets a host register its own policy plugins, and makes the config watcher notice documents the config references.

The engine is published, so this depends on `praxis-policy = "0.1.0"` from crates.io rather than a git ref. `allow-git` in `deny.toml` is empty again, which stops an unreviewed git dependency from slipping back in.

### Engine swap

- Swaps the engine dependency and renames the cargo feature `cpex-policy-engine` to `policy-engine`. This is a breaking change for anything that enabled the old name.
- Filter config keys do not change, `config_path` included. Existing operator configs keep working, and no fixture needed editing.
- Adds a behavior corpus under `filter/tests/`. Each document is parsed and fingerprinted, so the swap has to keep the same plugin names, kinds and route count.

### Host plugin registration

The engine bundles some plugins, but a host had no way to add one of its own. `PolicyFilter::new` built the engine and installed the bundled set, and that was the end of it. An unrecognised `kind` in a policy document is a startup error, so anything the engine did not ship simply could not be used.

The filter crate now exports:

```rust
register_policy_plugin_factory(kind, make);
```

`make` is a closure that returns a fresh factory each time it is called, rather than a factory value. Two reasons: the engine takes the factory by value, and `PolicyFilter::new` runs more than once.

That second point is the important one. The server rebuilds filter pipelines on every config change, so the filter is constructed again on each reload. The registry is therefore read on every build and never emptied. If it handed its entries out once, the first reload would fail with "no factory registered" for a config that had worked at startup.

Host registrations are applied after the bundled ones, so registering a `kind` the engine already ships replaces it. That is the useful direction, and a test pins the order so the opposite cannot creep in.

An unknown `kind` is still a hard startup error naming the kind, so a forgotten registration fails loudly instead of leaving a plugin that quietly never runs.

This is what let the PII scanner and the audit logger leave the engine's published set. They are now unpublished reference implementations, and the demo gateway registers them itself. Neither was fit to ship as supported: the scanner is regex matching with no Luhn check, and the logger writes to stderr.

### Reload fixes

- **Closes #900.** The reload gate hashed `praxis.yaml` alone, so a filter that loaded a separate policy document never noticed edits to it. The gate now covers every referenced document, the event filter accepts those paths, and each of their directories is watched. Paths are hashed alongside content, so two documents swapping contents still counts as a change, and an unreadable document hashes distinctly so deleting one counts too.
- Fixes a second bug in the same function. The content hash was recorded before parsing, so a reload that failed still marked the new content as seen and every retry was skipped, discarding the operator's edit for good. The hash now advances only on success, so the existing backoff can retry.

### Verification

- 1855 filter tests with `policy-engine` on, 1413 across core, protocol and server. `make lint` and `make audit` clean.
- All four corpus fingerprints are unchanged across the swap.
- The `cpex` demo runs end to end: the walkthrough, the PII deny, the session-taint deny, and the Cedar, CEL and OPA decision points.
- Hot reload is checked against the running gateway, since that is the path a drain-once registry would break. Editing the policy document reloads cleanly, the host factories re-register on every rebuild, and the PII scenario still denies afterwards.

### Known limits

Filters nested in branch chains are not walked for referenced files, matching the existing `apply_insecure_options` walk. The referenced set is collected at startup, which is enough because it comes from the main config, and a change to that always passes the gate on its own.


filter/src/pipeline/checks.rs
```diff
@@ -36,7 +36,7 @@ const SECURITY_FILTERS: &[&str] = &[
     "guardrails",
     "ip_acl",
     "peer_identity_trust",
-    #[cfg(feature = "cpex-policy-engine")]
+    #[cfg(feature = "policy-engine")]
```
shaneutt, line 39: Let's make sure that we create a Github issue for making this feature on-by-default, and assign it to milestone `v0.6.0`.

shaneutt (changes requested):
I think we may be missing a couple of unit tests for some of the added functions. Maybe do a quick sweep to double check, and then LGTM. 👍

filter/src/pipeline/checks.rs
araujof: Issue created: https://github.com/praxis-proxy/praxis/issues/987

araujof:
@shaneutt added additional tests to cover the missing bits. Also opened an issue to track the feature gate removal of the policy engine: https://github.com/praxis-proxy/praxis/issues/987

N.B.: also fixed a flaky test from the last merge from main.

shaneutt (approved):

shaneutt:
Thanks @araujof!

PR #954, opened 2026-08-11
feat(filter): expose credential-free AuthenticatedIdentity extension

leseb:
### What does this PR do?

Adds a public, read-only `AuthenticatedIdentity` request extension containing subject ID, typed roles/teams, and validated custom claims. The CPEX policy filter populates it after successful identity resolution so downstream filters can consume trusted identity without accessing JWTs, credentials, or provider payload types.

Key behaviors:
- Route-scoped policies publish identity only after authoritative resolution (not from the unscoped early gate)
- Body-first pre-read authenticates before downstream body consumers can execute
- Admission state is isolated per policy instance to prevent cross-policy overwrites
- Missing entity metadata fails closed instead of silently passing

N/A

### Checklist

- [x] Tests added or updated
- [ ] Documentation updated (if applicable)

Yes (minor). For entity-aware policies with `require_protocol_metadata: true` (the default), a request carrying `mcp.method` but missing `mcp.name` now rejects with HTTP 500 (`config.missing_protocol_metadata`) instead of silently passing through to the upstream. This is an intentional fail-closed hardening: the previous behavior admitted requests without entity-level authorization when the classifier wrote an incomplete metadata set. Operators who rely on the previous pass-through behavior can set `require_protocol_metadata: false`.


leseb:
@alexsnaps would like your opinion on this approach.

filter/src/extensions.rs
```diff
@@ -19,7 +19,85 @@
...
+    }
+
+    /// Provider-mapped custom claims from the validated identity.
+    ///
+    /// Registered JWT claims and claims promoted to the typed fields above are
+    /// excluded by the CPEX claim mapper. Values use CPEX's normalized string
+    /// representation rather than exposing the raw decoded JWT claim set.
+    /// Wire adapters must explicitly allowlist and bound claims before
+    /// serializing them outside Praxis.
+    pub fn custom_claims(&self) -> &BTreeMap<String, String> {
```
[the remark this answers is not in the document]
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/d7b57fcf74ef30edd1bf747f88a77e296ce5db10

filter/src/builtins/http/security/policy/tests.rs
```diff
@@ -1922,12 +2341,12 @@ async fn session_taint_is_isolated_across_principals() {
...
-/// Non-EOS chunks must pass through untouched — CMF dispatch waits
-/// for the full body so the upstream protocol classifier filter has finished parsing
-/// and writing metadata. Pins the streaming-chunk fast path.
+/// Entity-routed policy waits for the full body so the upstream protocol
+/// classifier has finished parsing and writing route metadata. Non-entity
+/// policies intentionally authenticate on the first chunk instead.
 #[tokio::test(flavor = "multi_thread")]
-async fn on_request_body_continues_on_partial_chunks() {
-    let (_dir, path) = write_single_plugin_config();
+async fn entity_on_request_body_continues_on_partial_chunks() {
```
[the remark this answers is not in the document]
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/d7b57fcf74ef30edd1bf747f88a77e296ce5db10

filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -588,23 +765,23 @@ impl HttpFilter for PolicyFilter {
...
         let Some(entity_name) = ctx.get_metadata("mcp.name").map(str::to_owned) else {
-            tracing::debug!(
+            tracing::error!(
                 target: "policy.filter",
                 protocol_method = %method,
-                "JSON-RPC method missing mcp.name metadata; skipping CMF dispatch",
+                "entity-bound JSON-RPC method is missing mcp.name metadata; denying fail-closed",
             );
-            return Ok(FilterAction::BodyDone);
+            return Ok(FilterAction::Reject(missing_protocol_metadata_rejection()));
```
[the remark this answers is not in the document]
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/d7b57fcf74ef30edd1bf747f88a77e296ce5db10

filter/src/builtins/http/security/policy/tests.rs
```diff
@@ -732,6 +795,216 @@ fn build_filter(config_path: String) -> PolicyFilter {
...
+async fn pre_read_authenticates_before_identity_only_body_consumer() {
+    let (_dir, path) = write_single_plugin_config();
+    assert_pre_read_admission(&path, Method::POST).await;
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn pre_read_authorizes_before_pure_l7_body_consumer() {
+    let (_dir, path) = write_l7_global_config();
+    assert_pre_read_admission(&path, Method::GET).await;
+}
```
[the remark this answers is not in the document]
leseb: StreamBuffer reports EOS only after read_request_body() returns None; a non-empty single-chunk request still enters the tested non-EOS path first. The existing test would also catch an EOS guard incorrectly moved before admission.

filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -331,11 +354,135 @@ impl PolicyFilter {
         ext
     }
 
+    /// Publish the raw-credential-free subject projection for downstream filters.
+    ///
+    /// CPEX may authenticate a client or workload without resolving a user
```
shaneutt:
```suggestion
    /// Praxis Policy Engine (PPE) may authenticate a client or workload without resolving a user
```

filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -331,11 +354,135 @@ impl PolicyFilter {
...
+    /// subject. In that case no `AuthenticatedIdentity` is published; a
+    /// consumer that requires a user principal must fail closed when the
+    /// extension is absent.
+    fn publish_authenticated_identity(ctx: &mut HttpFilterContext<'_>, identity: &IdentityPayload) {
+        Self::publish_identity_projection(ctx, Self::authenticated_identity(identity));
+    }
+
+    /// Build the public raw-credential-free projection from CPEX's private
```
shaneutt:
```suggestion
    /// Build the public raw-credential-free projection from PPE's private
```

filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -331,11 +354,135 @@ impl PolicyFilter {
...
+
+    /// Remove and return this policy instance's temporary early-gate result.
+    fn take_gated_identity(ctx: &mut HttpFilterContext<'_>) -> GatedIdentity {
+        ctx.get_filter_state_mut::<AdmissionState>()
+            .map_or(GatedIdentity::NotRun, |state| std::mem::take(&mut state.gated_identity))
+    }
+
+    /// Run the unscoped identity hook and return only its raw-credential-free
+    /// projection. Route-aware callers defer publication until classification.
+    #[expect(clippy::large_stack_frames, reason = "async CPEX identity payload")]
```
shaneutt:
```suggestion
    #[expect(clippy::large_stack_frames, reason = "async PPE identity payload")]
```

filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -349,19 +496,20 @@ impl PolicyFilter {
...
-            return Ok(FilterAction::Reject(auth_rejection(result.violation.as_ref())));
+        let authenticated = match self.resolve_gated_identity(ctx).await {
+            Ok(authenticated) => authenticated,
+            Err(rejection) => {
+                tracing::debug!(target: "policy.filter", "identity deny (on_request)");
+                return Ok(FilterAction::Reject(rejection));
+            },
+        };
+        // Entity-routed policies resolve again once classifier metadata gives
+        // CPEX the authoritative route coordinates. Publishing this unscoped
```
shaneutt:
```suggestion
        // PPE the authoritative route coordinates. Publishing this unscoped
```

shaneutt (changes requested):
We are in the midst of CPEX becoming Praxis Policy Engine (PPE), so it would be fair to update the nomenclature throughout.

Also, would like @araujof to give this one a look (please!)

filter/src/builtins/http/security/policy/filter.rs
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/a055de413c5826d4375f00acb9fc98879e3527eb

filter/src/builtins/http/security/policy/filter.rs
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/a055de413c5826d4375f00acb9fc98879e3527eb

filter/src/builtins/http/security/policy/filter.rs
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/a055de413c5826d4375f00acb9fc98879e3527eb

filter/src/builtins/http/security/policy/filter.rs
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/a055de413c5826d4375f00acb9fc98879e3527eb

araujof:
Thanks for this PR! 

I was reading the complete proposal that depends on this PR, and thought of suggesting a few things that could be used to better support the use case. Please feel free to disagree. Happy to help support this! 

1. Based on my understanding of the proposal, the policy filter already does almost exactly this job for delegation flows, where [attach_delegated_tokens](https://github.com/praxis-proxy/praxis/blob/c9c2a46898ebd47f58cffde5865f9e976078fa6e/filter/src/builtins/http/security/policy/filter.rs#L444) takes engine-derived state, sorts it, and writes it onto the upstream request. Identity is the same operation on the same data at the same point in the request. So instead of a new extension plus a downstream filter that reads it, this could be a config block on the policy filter itself (header names in, projection out). That reduces a filter, and it sidesteps ordering entirely, since header mutations don't end up in the request until the pipeline finishes anyway.

2. The same config surface could cover what shouldn't propagate (e.g., JWTs, raw credentials, provider payload types). We could leverage the already existent capability gating we have in PPE. In this PR, that's encoded in the shape of AuthenticatedIdentity, which means the policy is in the type. As config it's something an operator can see and audit, and you don't need a new type per consumer that wants a slightly different slice.

3) Lastly, on our side: we've got the claims data-model fix up as [#9](https://github.com/praxis-proxy/policy/pull/9), so subject claims keep their JSON shape instead of being stringified. The remaining half is making the claim map configurable, which is what actually completes IdP coverage (nested Keycloak roles, Auth0's namespaced claims, etc.).

Worth noting, if you go the config route, custom_claims() would want to hand back structured values rather than strings, since after [#9](https://github.com/praxis-proxy/policy/pull/9) there's real structure to hand back. 

Thoughts?

cc: @shaneutt @terylt

filter/src/extensions.rs
```diff
@@ -19,7 +19,88 @@
...
+
+    /// Teams assigned to the authenticated subject.
+    pub fn teams(&self) -> &BTreeSet<String> {
+        &self.teams
+    }
+
+    /// Provider-mapped custom claims from the validated identity.
+    ///
+    /// Registered JWT claims and claims promoted to the typed fields above are
+    /// excluded by the CPEX claim mapper. Values use CPEX's normalized string
```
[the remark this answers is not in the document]
leseb: https://github.com/praxis-proxy/praxis/pull/954/commits/18c0ef0679af637025849f2420fd00396b318593

alexsnaps:
> @alexsnaps would like your opinion on this approach.

You are back, and I've not come around to this 🤦 
Sorry about that!

alexsnaps:
I feel like I might be missing an obvious piece here, but I'm concerned/surprised to find `mcp` or any other "AI-centric" wording in _this filter_ or _in this code base in general_. But most was already there... I'm wondering how much of these concerns shouldn't be extracted in the policy itself and the filter be agnostic of these? 

But, yeah... Time might be of the essence here? So, feel free to ignore me. Tho I'd appreciate the background (and possibly where we see this going moving forward, if only far off in a distant future...)

shaneutt:
Thank you for your contribution! This has been merged into `main` as part of a bulk merge, see: https://github.com/praxis-proxy/praxis/commit/1d17fd3aee27f9108e27230713358cc8b61239dc

PR #908, opened 2026-08-03
feat(protocol): root span per request with HTTP semantic conventions

Ladas:
## What
Create a root tracing span for each HTTP request covering the full lifecycle from request_filter through logging, with OTel HTTP semantic convention attributes.

## Why
Without root spans, there are no traces to view in Tempo. This is the prerequisite for all downstream tracing work (filter spans, upstream spans, trace context propagation).

## How
- Root `info_span!("http_request")` created in request_filter phase
- OTel attributes: http.request.method, url.path, http.response.status_code, server.address, client.address, network.protocol.version
- request_id as recorded span field (propagates to all log lines within span)
- All ProxyHttp hooks instrumented with the span (.instrument for async, span.enter for sync)
- Status code and upstream info recorded in logging phase
- Span dropped at end of logging for accurate duration
- 10 new tests

**Depends on** #315 (PR #836)

Closes #301

PR #957, opened 2026-08-12
feat(filter): trace_context filter for W3C traceparent propagation

### LINKED ISSUE #296: trace_context filter: parse/inject traceparent and tracestate
W3C Trace Context propagation filter that parses incoming and injects outgoing traceparent and tracestate headers.

## Requirements

- Parse `traceparent` header per W3C Trace Context specification
- Parse `tracestate` header, preserve vendor entries
- Generate new trace ID if no incoming traceparent
- Join existing trace (create child span) if traceparent present
- Inject `traceparent` and `tracestate` into upstream request
- Validate traceparent format, ignore malformed headers

## Implementation Notes

- Implement as `HttpFilter` under `filter/src/builtins/http/observability/`
- W3C Trace Context Level 2 compliance
- Optional: opt-in b3 and Jaeger propagation format support

Parent epic: #10

Ladas:
## Summary

- Add `trace_context` filter that parses incoming W3C `traceparent` headers and injects them on upstream requests
- Propagates trace context across service boundaries for distributed tracing
- Supports `tracestate` header pass-through

Closes #296

## Dependencies

Depends on #908 (root span per request) — first commit is the #301 dependency.

## Test plan

- [ ] Verify traceparent header parsed from incoming requests
- [ ] Verify traceparent injected on upstream requests
- [ ] Verify tracestate passed through
- [ ] Unit tests for W3C traceparent format parsing


filter/src/builtins/http/observability/trace_context.rs
```diff
@@ -0,0 +1,755 @@
...
+
+    async fn on_request(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
+        let incoming = ctx
+            .request
+            .headers
+            .get("traceparent")
+            .and_then(|v| v.to_str().ok())
+            .and_then(Traceparent::parse);
+
+        let new_span_id = generate_span_id(ctx);
```
nerdalert, line 193: **The outgoing parent ID is not the ID of an exported Praxis span.**

The root `http_request` span is created in the protocol handler before this filter runs, but this code generates an unrelated ID and places it in the outgoing `traceparent`. Nothing creates or exports a span with `new_span_id`, so the downstream span points to a missing parent while Praxis's exported server span belongs to a separate context.

This also shows why W3C extraction and injection should be owned by the existing OTel span lifecycle rather than implemented as an independent ID-generating filter. The required flow is:

```text
extract incoming W3C context
  -> create the Praxis SERVER span with that parent
  -> run the request pipeline
  -> inject context derived from the actual outbound/Praxis span
```

Please use the OpenTelemetry W3C propagator and the actual span context for extraction and injection. The filter can remain the configuration surface, but this cannot be corrected safely with a local change in this function because inbound extraction must happen before the root span is created.

filter/src/builtins/http/observability/trace_context.rs
```diff
@@ -0,0 +1,755 @@
...
+    );
+    id
+}
+
+/// Generate a 16-hex-char span ID using the context's ID generator.
+fn generate_span_id(ctx: &HttpFilterContext<'_>) -> String {
+    let id = ctx.id_generator.generate(ctx.time_source);
+    // Take the first 16 hex chars (8 bytes) for a span ID.
+    // IdGenerator produces 32 hex chars; truncate to span-ID length.
+    let id: String = id.chars().take(SPAN_ID_HEX_LEN).collect();
```
nerdalert, line 295: **Truncating the request ID this way can produce duplicate span IDs.**

`IdGenerator` formats its value as timestamp + process seed + counter. Taking the first 16 characters discards the counter, so requests generated in the same microsecond can receive the same span ID. The length assertion only proves that the duplicated value has the required width.

Please remove this generator and use the OpenTelemetry SDK's trace/span ID generation through the real span context. A small truncation adjustment would still leave this implementation separate from the exported Praxis span, so the SDK-owned context is the appropriate fix.

filter/src/builtins/http/observability/trace_context.rs
```diff
@@ -0,0 +1,755 @@
...
+        debug!(
+            trace_id = %tp.trace_id,
+            parent_id = %tp.parent_id,
+            trace_flags = %tp.trace_flags,
+            "joining existing trace"
+        );
+        tp.format(new_span_id)
+    } else {
+        let trace_id = generate_trace_id(ctx);
+        let trace_flags = "01"; // sampled
```
nerdalert, line 227: **The sampled flag must come from the actual span context.**

Hardcoding `01` advertises every newly created trace as sampled even when the configured Praxis sampler does not record it. Downstream services may then record traffic that Praxis dropped, which can increase tracing load and produce inconsistent traces.

Once propagation is connected to the actual Praxis span, inject that span's trace flags instead of constructing them here. This should also preserve any configured parent-based or ratio-based sampling behavior.

shaneutt:
Thank you for your contribution! This has been merged into `main` as part of a bulk merge, see: https://github.com/praxis-proxy/praxis/commit/36ff2507da97f40e7086cd2cda5f80e80097aa84

Issue #951, opened 2026-08-11
core: add streaming SubRequestClient response transport


leseb:
## Description

Add an opaque, pull-based streaming response API to `praxis-core::subrequest`
without replacing the existing buffered `SubResponse` contract.

The core client owns the Pingora session, admission permit, transport
deadlines, limits, cancellation, cleanup, and connection reuse decisions for
the full response-body lifetime. Callers continue to own routing, status
interpretation, retries, failover, and provider-specific behavior.

The representative API consists of `StreamingSubResponse`, an opaque
`SubResponseBody`, `StreamLimits`, and `SubRequestClient::send_streaming()`.
`SubResponseBody::next_chunk()` must be demand-driven so downstream writes
provide backpressure without a background producer queue.

Buffered `SubRequestClient::execute()` should collect through the same
transport path, preserving its current cloneable `SubResponse` result and
buffered response ceiling.

## Acceptance criteria

- `send_streaming()` returns sanitized response headers without buffering the
  response body.
- The response body retains the Pingora session and admission permit until
  EOF, cancellation, timeout, error, or drop cleanup completes.
- Streaming distinguishes bounded buffered/in-flight bytes, an optional total
  byte limit, idle timeout, and optional overall timeout.
- The buffered `max_response_bytes` ceiling is not applied to streaming
  responses unless an explicit total limit is supplied.
- Buffered and streaming execution share request construction, header
  sanitization, framing, admission, connection, and typed-error logic.
- A completed response is safely reusable; an incomplete HTTP/1.1 response is
  closed and an HTTP/2 cancellation resets only its stream.
- A valid response header resolves the circuit-breaker probe; late body
  failures are recorded as stream termination outcomes without holding a
  half-open probe for a long-lived SSE response.
- Tracing and metrics cover time to headers, stream duration, bytes,
  cancellation, timeouts, and termination cause.
- Tests cover incremental delivery, backpressure, limits, stalled headers and
  bodies, cancellation, early drop, admission release, and HTTP/1.1 and HTTP/2
  reuse behavior.

## Non-goals

- Parsing SSE or provider-specific payloads.
- Selecting upstreams or interpreting HTTP status codes.
- A buffered/streaming configuration mode.
- Streaming request bodies or HTTP upgrade tunnels.

## Parent epic

- #877

PR #980, opened 2026-08-14
feat(filter): integrate streaming response transport into IRR

### LINKED ISSUE #953: iterative_request_router: forward filter-selected terminal response streams
## Description

Integrate filter-selected terminal response streaming into
`iterative_request_router` (IRR) while preserving the current buffered path
for ordinary and body-dependent rounds.

Each step continues to run its request-header and request-body filters before
dispatch. After those filters construct the effective outbound request, IRR
reads the typed response mode from the step context. Buffered mode uses the
existing `SubRequestClient::execute()` lifecycle. Streaming mode uses
`send_streaming()`, runs step response-header filters, and evaluates only
information available before the body is exposed.

A streaming response may perform status/origin/transport/default failover
before downstream commitment. IRR must cancel the unread response body before
moving to another step. Once a response is selected for downstream exposure,
IRR returns an owned continuation capable of running the step response-body
pipeline for the complete stream lifetime.

## Acceptance criteria

- IRR reads the filter-selected response mode after step request-body filters
  finish constructing the effective outbound request.
- Buffered remains the default and existing buffered IRR behavior and tests
  remain unchanged.
- Streaming steps run response-header filters before transition evaluation or
  downstream commitment.
- Streaming transitions are limited to status, origin, transport error, and
  default predicates available before the body is exposed.
- Header-safe failover cancels the unread response body and advances to the
  selected step without exposing the failed response.
- Filter-result/body-dependent transitions and `BodyMode::StreamBuffer` fail
  static validation for pipelines declaring streaming capability, with runtime
  guards for dynamically registered filters.
- A selected stream owns its step pipeline, request snapshot, extensions,
  metadata, filter state, executed-filter indices, and completion state beyond
  `on_request()`.
- Step response-body filters run once per chunk before parent response-body
  filters, and both completion lifecycles run exactly once.
- No transition or fallback occurs after downstream headers are committed.
- IRR's buffered `max_response_bytes` limit does not silently cap a stream;
  only an explicit streaming total limit does.
- A test-only filter selects streaming so Praxis does not add a production
  YAML streaming-mode filter solely for testing.
- Integration tests cover early SSE delivery, backpressure, header-based
  failover, per-chunk step and parent filtering, downstream disconnects, late
  errors, completion, and HTTP/1.1/HTTP/2 framing.
- The public selection and terminal-round lifecycle contract is documented for
  external filter crates.

## Non-goals

- Exposing intermediate rounds whose response bodies determine control flow.
- Parsing OpenAI or Anthropic request/response payloads in IRR.
- Silently buffering a filter-selected stream for an incompatible pipeline.

## Dependencies

- #951
- #952

## Parent epic

- #877

leseb:
### What does this PR do?

Adds streaming response support to the Iterative Request Router (IRR). When a step's filters select `SubRequestResponseMode::Streaming`, IRR dispatches via `send_streaming()` instead of the buffered `execute()` path, enabling real-time chunk delivery for SSE and long-lived streaming responses. Response-header filters run before transition evaluation so header-based failover still works. The existing buffered path is unchanged.

This PR is stacked on #979 (streaming terminal response lifecycle) and #951 (streaming SubRequestClient transport).

Fixes #953

### Checklist

- [x] Tests added or updated

No. Streaming is opt-in via filter selection; the default buffered path is unchanged.

### DIFF
--- a/filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
+++ b/filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
@@ -14,12 +14,19 @@
 //! inject its own credentials and required representation headers
 //! (e.g. via a `headers` or `credential_injection` filter).
 //!
-//! # Streaming limitations
+//! # Streaming support
 //!
-//! Intermediate and terminal responses are fully buffered in memory
-//! (bounded by `max_response_bytes`). SSE and long-lived streaming
-//! responses are not supported. The iterative request router is
-//! designed for bounded, buffered workflows.
+//! When a step's filters select [`SubRequestResponseMode::Streaming`],
+//! IRR dispatches via `send_streaming()` instead of the buffered
+//! `execute()` path. Response-header filters run before transition
+//! evaluation; only status, origin, transport error, and default
+//! predicates are available. Once a streaming response is committed
+//! downstream, no failover or replacement response can occur.
+//!
+//! Transitions that depend on filter results (filter/key/value
+//! predicates) or `BodyMode::StreamBuffer` response mode are
+//! rejected at config validation time for streaming-capable pipelines,
+//! with a runtime guard for dynamically registered filters.
 //!
 //! # Position requirement
 //!
@@ -31,6 +38,7 @@
 //! See proposal 00786 for the full design rationale.
 
 mod config;
+mod streaming;
 #[cfg(test)]
 #[expect(clippy::allow_attributes, reason = "blanket test suppressions")]
 #[allow(clippy::unwrap_used, clippy::expect_used, reason = "tests")]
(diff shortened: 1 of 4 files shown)


filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs
```diff
@@ -14,12 +14,19 @@
...
+//! IRR dispatches via `send_streaming()` instead of the buffered
+//! `execute()` path. Response-header filters run before transition
+//! evaluation; only status, origin, transport error, and default
+//! predicates are available. Once a streaming response is committed
+//! downstream, no failover or replacement response can occur.
+//!
+//! Transitions that depend on filter results (filter/key/value
+//! predicates) or `BodyMode::StreamBuffer` response mode are
+//! rejected at config validation time for streaming-capable pipelines,
+//! with a runtime guard for dynamically registered filters.
```
shaneutt, line 29: Good docs 👍

shaneutt (approved):
There are several very long functions with bypasses that we should assess, but that's not something entirely new to this PR so I don't want to hold this up for that, we'll follow-up.

PR #982, opened 2026-08-14
fix(core): keep Tokio runtime alive for OTel batch exporter

Ladas:
## Problem

The default `BatchSpanProcessor` in opentelemetry_sdk 0.32 drives export
calls via `futures_executor::block_on`, which cannot poll Tokio I/O
primitives. The tonic gRPC transport requires a Tokio reactor for its
Hyper-based HTTP/2 connections. Without a live Tokio runtime, traces are
silently dropped — the exporter builds successfully but every batch
export fails at the network layer.

## Fix

Create a persistent multi-thread Tokio runtime (1 worker thread) stored
in `TracingGuard` that stays alive for both the initial tonic channel
build and all subsequent batch export calls. The runtime is dropped on
graceful shutdown alongside the tracer provider.

## Testing

Verified end-to-end on KIND with OTel Collector + Tempo:
- Before fix: 0 traces in Tempo, collector receives nothing
- After fix: traces flow correctly, visible in Grafana

Related: praxis-proxy/forge#2 (OTel benchmark demo)

### DIFF
--- a/core/src/logging.rs
+++ b/core/src/logging.rs
@@ -33,6 +33,9 @@ pub struct TracingGuard {
     /// Tracer provider to shut down when the guard is dropped.
     #[cfg(feature = "otel")]
     provider: Option<opentelemetry_sdk::trace::SdkTracerProvider>,
+    /// Tokio runtime kept alive for the tonic gRPC exporter.
+    #[cfg(feature = "otel")]
+    _otel_runtime: Option<::tokio::runtime::Runtime>,
 }
 
 #[cfg(feature = "otel")]
@@ -135,7 +138,7 @@ fn init_with_otel(
 ) -> Result<TracingGuard, ProxyError> {
     use opentelemetry::trace::TracerProvider as _;
 
-    let provider = build_otel_provider(telemetry)?;
+    let (provider, otel_runtime) = build_otel_provider(telemetry)?;
 
     // `OpenTelemetryLayer<S>` requires `S` to match the composed subscriber type.
     // JSON and text fmt produce different types, preventing a shared binding.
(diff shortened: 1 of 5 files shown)


core/src/logging.rs
```diff
@@ -217,7 +229,13 @@ fn build_otel_provider(
...
     }
 
-    let exporter = build_span_exporter(endpoint, &config.otlp_headers)?;
+    let runtime = ::tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(1)
+        .enable_all()
+        .build()
+        .map_err(|e| ProxyError::Config(format!("failed to create OTel runtime: {e}")))?;
+
+    let exporter = runtime.block_on(async { build_span_exporter(endpoint, &config.otlp_headers) })?;
```
nerdalert, line 238: This keeps the Tokio reactor alive for the exporter, which looks like the correct fix. However, the current tests only cover the no-endpoint path, and the existing OTLP integration test allows an unreachable collector. Neither test would catch the exporter failing to perform batch I/O after initialization.

Could we add a focused regression test that starts an in-process OTLP/gRPC receiver, exports a span through the real tonic batch exporter, flushes or shuts down the provider, and verifies that the receiver received the span? That would directly cover the runtime-lifetime failure this PR fixes.

Ladas: yes, I am now trying to go back from live testing, to see how I can cover the gaps in the integration tests

nerdalert (approved):

shaneutt (changes requested):
Needs some conflict and comment resolution and then I think we're pretty good to go.

Issue #797, opened 2026-07-10
Non-blocking log writer with file rotation


shaneutt:
## Summary

Logging is synchronous today (`tracing_subscriber::fmt` writing to stdout). Under high throughput this can block the async runtime. Add a non-blocking writer via `tracing-appender` with optional file output and rotation.

## Design

New dependency: `tracing-appender` (same tracing ecosystem, maintained by tokio-rs).

Changes to `core/src/logging.rs`:
- Replace direct `fmt().init()` with a layered setup using `tracing_appender::non_blocking` writer
- Optional file output with `RollingFileAppender` (daily or size-based rotation)
- Guard handle returned to caller to ensure flush on shutdown

Config:
```yaml
runtime:
  logging:
    output: stdout  # or file
    file_path: /var/log/praxis/proxy.log
    rotation: daily  # or size:100mb
    non_blocking: true
    buffer_size: 8192
```

When `output: stdout` (default), behavior is identical to today but non-blocking. When `output: file`, logs write to a rotating file instead.

## Acceptance criteria

- [ ] Non-blocking writer wrapping stdout (default)
- [ ] File output with daily and size-based rotation
- [ ] Graceful flush on shutdown (guard pattern)
- [ ] Config parsing and validation
- [ ] No log loss under normal operation (verify buffer sizing)
- [ ] Integration test: file rotation produces expected files

## Related issues

- #315 Wire tracing subscriber: fmt + OTLP layers

PR #989, opened 2026-08-17
feat(logging): non-blocking process log writer with file rotation (#797)

henschwartz:
## Summary
- Add `runtime.logging` for process log destination (`stdout`/`stderr`/`file`), rotation, retention, and non-blocking buffering
- Wire non-blocking writers via `tracing-appender` with lossy overflow; extend `TracingGuard` to flush on shutdown (OTLP teardown first, then `WorkerGuard`)
- Validate logging on startup/reload; warn when `runtime.logging` changes require restart
- Add example config, operating docs, and integration/schema coverage for rotation and shutdown flush
Implements [#797](https://github.com/praxis-proxy/praxis/issues/797) per the merged How? in [#947](https://github.com/praxis-proxy/praxis/pull/947).
## Test plan
- [x] `make fmt`
- [x] `make lint`
- [x] `make doc`
- [x] `make test-unit`
- [x] `cargo test -p praxis-proxy-core logging`
- [x] `cargo test -p praxis-tests-integration process_log process_logging`
- [x] `cargo test -p praxis-tests-schema process_logging`
- [x] `make audit`
- [x] `make coverage-check` (96.68% lines)
- [x] Live smoke: file logging + HTTP 200 + graceful SIGTERM flush


core/src/config/logging.rs
```diff
@@ -0,0 +1,306 @@
...
+#[serde(default, deny_unknown_fields)]
+pub struct LoggingConfig {
+    /// Log destination (`stdout`, `stderr`, or `file`).
+    pub output: LogOutput,
+    /// Active log file path when `output` is `file`.
+    pub file_path: Option<String>,
+    /// Optional rotation policy (`daily` or `size:<N><kb|mb|gb>`).
+    pub rotation: Option<LogRotation>,
+    /// Maximum rotated files to retain.
+    #[serde(default = "default_max_log_files")]
```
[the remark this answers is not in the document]
henschwartz: Fixed — docs now say total files retained including the active file, matching tracing-appender max_log_files (e.g. 7 ⇒ active + .1….6).

core/src/config/logging.rs
```diff
@@ -0,0 +1,306 @@
...
+        };
+        let err = cfg.validate().unwrap_err();
+        assert!(err.contains("rotation is only valid when output is file"), "{err}");
+    }
+
+    #[test]
+    fn max_files_zero_rejected() {
+        let cfg = LoggingConfig {
+            max_files: 0,
+            ..LoggingConfig::default()
```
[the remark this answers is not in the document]
henschwartz: Added unit tests for buffer_size: 0, empty file_path, file_path with stdout, and rotation: size:0.

core/src/logging/size_rotation.rs
```diff
@@ -0,0 +1,194 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Praxis Contributors
+
+//! Size-based log file rotation.
+
+#![allow(clippy::missing_docs_in_private_items, reason = "internal rotation writer")]
+
```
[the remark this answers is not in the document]
henschwartz: Removed module-level allows; added /// on size_rotation and writer items instead.

tests/integration/tests/suite/process_logging.rs
```diff
@@ -0,0 +1,214 @@
...
+            .arg(child.id().to_string())
+            .status()
+            .expect("send SIGTERM");
+    }
+    #[cfg(not(unix))]
+    {
+        child.kill().expect("kill child");
+    }
+    let status = child.wait().expect("wait child");
+    assert!(status.success(), "graceful shutdown should exit zero");
```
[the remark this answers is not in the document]
henschwartz: Moved status.success() into the Unix SIGTERM path only; non-Unix uses kill+wait without asserting graceful exit (suite + example tests).

core/src/logging/size_rotation.rs
```diff
@@ -0,0 +1,206 @@
...
+            path,
+            max_bytes,
+            max_files,
+            state: Mutex::new(ActiveFile { file: Some(file), size }),
+        })
+    }
+
+    /// Roll the active file when it exceeds `max_bytes`.
+    fn roll_locked(state: &mut ActiveFile, path: &Path, max_files: u32) -> io::Result<()> {
+        if let Some(mut file) = state.file.take() {
```
jordigilh, line 53: `state.file` is cleared here before `prune_and_shift`/`rename` run below. If either fails (disk full, permission error, concurrent directory deletion), this returns `Err` but the handle is never restored — every subsequent `write` then permanently fails for the rest of the process. Checked `tracing-appender` 0.2.5's actual worker source: it swallows that error silently (no metric, no stderr), so this becomes a completely invisible, permanent logging outage from an ordinary disk hiccup — which is exactly what #797's acceptance criteria call out ("no log loss under normal operation"). Might be worth attempting to reopen the original path on a failed roll so it degrades to "no rotation" rather than "no logging".

jordigilh (commented):
One inline comment below on the load-bearing bug. One more note here since it compares against files this PR doesn't touch: `runtime.logging.file_path` doesn't get the permission-hardening or symlink checks this repo already applies elsewhere (`warn_insecure_key_permissions` for TLS keys in `startup_checks.rs:151-179`, `warn_if_symlink` for `upstream_ca_file` in `validate/rules.rs:256-286`) — worth a follow-up given logs can carry sensitive metadata.

core/src/logging/size_rotation.rs
henschwartz: agreed, roll_locked now reopens the active path (or undoes the rename) when prune/rename fails, so we keep appending instead of leaving file unset. Added continues_logging_when_roll_fails unit test.

core/src/config/logging.rs
```diff
@@ -0,0 +1,351 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Praxis Contributors
+
+//! `runtime.logging` configuration.
+
+use std::path::Path;
+
+use serde::{Deserialize, Serialize};
+use tracing_appender::non_blocking::DEFAULT_BUFFERED_LINES_LIMIT;
+
```
shaneutt, line 10: This isn't using the standard comment separators consistently with the rest of the codebase.

core/src/logging/size_rotation.rs
```diff
@@ -0,0 +1,287 @@
...
+
+use std::{
+    fs::{self, File, OpenOptions},
+    io::{self, Write},
+    path::{Path, PathBuf},
+    sync::Mutex,
+};
+
+use crate::errors::ProxyError;
+
```
shaneutt, line 14: Missing standard comment separators

tests/integration/tests/suite/process_logging.rs
```diff
@@ -0,0 +1,215 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2024 Praxis Contributors
```
shaneutt: These copyright notices come from old Praxis, needs to be updated to:


```suggestion
// Copyright (c) 2026 Praxis Contributors
```

tests/integration/tests/suite/process_logging.rs
```diff
@@ -0,0 +1,215 @@
...
+use std::{
+    fs, io,
+    net::TcpStream,
+    process::{Command, Stdio},
+    thread,
+    time::Duration,
+};
+
+use praxis_test_utils::{free_port, praxis_bin};
+
```
shaneutt, line 15: Comment separators

shaneutt (changes requested):
Did an initial pass and pointed out some minor stuff.

This is a huge PR however, so I'll need to find some significant time to sit down and work through it, more passes to come.

core/src/config/logging.rs
henschwartz: added

tests/integration/tests/suite/process_logging.rs
henschwartz: changed

tests/integration/tests/suite/process_logging.rs
henschwartz: added

core/src/logging/size_rotation.rs
henschwartz: added

core/src/logging.rs
```diff
@@ -21,29 +24,30 @@ use crate::{config::Config, errors::ProxyError};
...
 /// RAII guard that flushes and shuts down the `OTel` tracer provider on drop.
 ///
-/// Store this in `main()` to ensure pending spans are exported on graceful
-/// shutdown. Without the `otel` feature, this is a zero-size no-op.
-///
 /// ```no_run
 /// let config = praxis_core::config::Config::load(None, "listeners: []").unwrap();
 /// let _guard = praxis_core::logging::init_tracing(&config).unwrap();
 /// ```
 #[must_use = "dropping the guard immediately shuts down the tracer provider"]
```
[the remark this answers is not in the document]
henschwartz: Fixed — TracingGuard docs and #[must_use] now describe log-writer flush; OTLP shutdown is called out only with the otel feature. init_tracing return docs updated to match.

core/src/logging/size_rotation.rs
```diff
@@ -0,0 +1,307 @@
...
+/// Path for rotated file `path.N`.
+fn rotated_path(path: &Path, index: u32) -> PathBuf {
+    let file_name = path.file_name().map_or_else(
+        || format!("rotated.{index}"),
+        |name| format!("{}.{}", name.to_string_lossy(), index),
+    );
+    path.with_file_name(file_name)
+}
+
+/// Drop the oldest archive and shift `.N` → `.N+1` before rolling.
```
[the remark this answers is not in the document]
henschwartz: Fixed — validate() now rejects max_files < 2 when size rotation is enabled; daily rotation may still use max_files: 1. Added unit tests for both paths.

core/src/logging/size_rotation.rs
```diff
@@ -0,0 +1,307 @@
...
+        let previous_size = state.size;
+        if let Some(mut file) = state.file.take()
+            && let Err(e) = file.flush()
+        {
+            state.file = Some(file);
+            return Err(e);
+        }
+
+        if let Err(roll_err) = Self::perform_roll(path, max_files) {
+            Self::restore_active_after_failed_roll(state, path, previous_size, roll_err)
```
aslakknutsen: No indication of roll error and the continues writing to the old file that is happening. An operator has no indication that "things are not working as intended".

henschwartz: Fixed, when a size roll fails but we successfully reopen the active file, we now emit tracing::warn! with the path and error ("log size rotation failed; continuing to append to the active file"). If reopen also fails, we emit tracing::error! and return the I/O error. continues_logging_when_roll_fails covers the degrade path (append-only, no new archive, warning captured); I also ran the scenario manually and confirmed on-disk state and the warning text.

core/src/logging/size_rotation.rs
```diff
@@ -70,14 +71,42 @@ impl SizeRotatingWriter {
...
+    fn restore_or_warn_after_failed_roll(
+        state: &mut ActiveFile,
+        path: &Path,
+        fallback_size: u64,
+        roll_err: io::Error,
+    ) -> io::Result<()> {
+        let roll_error = roll_err.to_string();
+        match Self::restore_active_after_failed_roll(state, path, fallback_size, roll_err) {
+            Ok(()) => {
+                tracing::warn!(
```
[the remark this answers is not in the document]
henschwartz: Fixed: rotation degrade/unrecoverable diagnostics now use eprintln! instead of tracing::warn!/error!, because the blocking file path runs inside the fmt layer's on_event dispatch where nested tracing events are dropped. Added failed_roll_diagnostic_reaches_stderr_under_subscriber_dispatch to prove the diagnostic fires under subscriber reentrancy while a tracing::warn! would not.

PR #1036, opened 2026-08-24
feat(tcp): add praxis_tcp_connection_duration_seconds histogram metric

### LINKED ISSUE #1027: praxis_tcp_connection_duration_seconds histogram
Add a `praxis_tcp_connection_duration_seconds` histogram measuring the lifetime of TCP connections from accept to close, labeled by `listener`. Expose it on the admin `/metrics` scrape endpoint to analyze connection duration distributions.

tsisodia10:
Record TCP connection lifetime (accept to close) as a Prometheus histogram labeled by listener name. The metric is emitted on all disconnect paths and exposed via the admin /metrics endpoint when enabled.

### What does this PR do?

Adds a `praxis_tcp_connection_duration_seconds` Prometheus histogram that records the lifetime of each TCP connection (from accept to close), labeled by listener name. This gives operators visibility into how long TCP connections live across different listeners.

**Changes:**
- New `protocol/src/tcp/metrics.rs` module with `record_tcp_connection_duration()` 
- Records duration at all disconnect exit paths in `PingoraTcpProxy::process_new()`
- Uses the existing `listener_label_for()` infrastructure for accurate per-listener labeling
- Example config: `examples/configs/observability/tcp-connection-metrics.yaml`
- Unit tests (no-panic without recorder) and integration tests (metric emission + correct labels)

Fixes #1027

### Checklist

- [x] Tests added or updated

No. The metric is purely additive and only emitted when the admin interface is enabled.

### DIFF
--- a/docs/operating/observability.md
+++ b/docs/operating/observability.md
@@ -122,9 +122,10 @@ return **400** JSON errors.
 
 ## Metrics Reference
 
-Praxis records two categories of Prometheus metrics:
-HTTP request metrics (always on when admin is
-enabled) and per-filter duration histograms (opt-in).
+Praxis records Prometheus metrics in three
+categories: HTTP request metrics, TCP connection
+metrics (both always on when admin is enabled), and
+per-filter duration histograms (opt-in).
 
 Recorder upkeep runs every five seconds whenever the
 admin endpoint is enabled. It is independent of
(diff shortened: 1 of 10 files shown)


protocol/src/tcp/proxy.rs
```diff
@@ -360,6 +360,10 @@ impl ServerApp for PingoraTcpProxy {
                 0,
             )
             .await;
+            super::metrics::record_tcp_connection_duration(
```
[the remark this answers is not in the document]
tsisodia10: Fixed, metric is now recorded on all three exit paths (connect-failure, peeked-write-error, and normal completion).

protocol/src/tcp/proxy.rs
aslakknutsen, line 322: record_tcp_connection_duration?

aslakknutsen, line 224: record_tcp_connection_duration?

aslakknutsen (changes requested):
sni_timeout and filter_rejection have measurable connection times and will be in connection_close logs, but not recorded as part of histogram. 

Intentional? 

Would it perhaps make sense to add a label for the connection close type?

tests/integration/tests/suite/tcp_connection_metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use praxis_core::config::Config;
+use praxis_test_utils::{free_port, http_get, start_full_proxy, start_tcp_tagged_backend, wait_for_tcp};
+
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+#[test]
+fn tcp_connection_duration_histogram_emitted_after_connection_close() {
```
aslakknutsen, line 20: Not a strict requirement, but how easy/hard would it be to trigger the other "reasons" in a test to ensure they are locked in and working?

tsisodia10: Added a `connect_failure` reason test that verifies the label by pointing the proxy at an unreachable upstream.

For `sni_timeout`: triggering it requires a client that connects but never sends data, then waiting out the full SNI_PEEK_TIMEOUT (5s hardcoded). That's doable but adds 5+ seconds to the suite per invocation. I'd lean toward making the peek timeout configurable first (useful for operators anyway), then covering it in a test with a low value. Happy to add it as an #[ignore] test in the meantime if you'd prefer coverage now over speed.

`filter_rejection` and `peeked_write_error` are harder to trigger deterministically without purpose-built test fixtures, open to suggestions if there's existing infra I'm missing.

aslakknutsen (approved):

tests/integration/tests/suite/tcp_connection_metrics.rs
aslakknutsen: connection_failure is ok enough for this. :)

docs/operating/observability.md
```diff
@@ -157,6 +158,45 @@ Wall-clock duration of completed HTTP requests in
...
+
+The `reason` label captures why the connection
+closed:
+
+| Reason | Meaning |
+| --------------------- | ---------------------------------------- |
+| `completed` | Normal forwarding finished |
+| `sni_timeout` | SNI peek timed out before routing |
+| `filter_rejection` | Connect filters rejected the connection |
+| `connect_failure` | Upstream connection failed |
```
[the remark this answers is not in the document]
aslakknutsen: it comes in https://github.com/praxis-proxy/praxis/pull/1037

PR #1060, opened 2026-09-01
feat(filter): add a bounded, incremental SSE codec

### LINKED ISSUE #986: filter: add a bounded incremental SSE codec
## Description

Add a provider-neutral, bounded incremental Server-Sent Events codec to
`praxis-filter`.

Pingora supplies response body chunks at arbitrary byte boundaries. The codec
must turn those chunks into complete SSE records for inspecting upstream
streams and encode records produced locally. It must remain independent of
provider JSON schemas and of Pingora session mechanics.

The implementation should be a small reusable state machine, publicly exported
for external filter crates. It should retain only the incomplete line or record
between calls and expose explicit limits rather than buffering a complete
event stream.

## Acceptance Criteria

- A public incremental decoder accepts successive byte chunks and yields zero
  or more complete provider-neutral SSE records.
- A public encoder serializes provider-neutral SSE records into canonical
  event-stream bytes.
- Records can represent standard `event`, `data`, `id`, and `retry` fields plus
  comments without assuming that `data` contains JSON.
- Multiline `data` fields and both LF and CRLF input are handled correctly.
- UTF-8 code points, field lines, delimiters, and multiple records may be split
  across arbitrary input chunks.
- Configurable maximum partial-line and record sizes prevent unbounded retained
  memory and return typed errors when exceeded.
- End-of-stream behavior is explicit and tested for complete, incomplete, and
  empty records.
- Exhaustive split-point tests feed representative streams at every possible
  byte boundary and produce the same records as an unsplit input.
- Tests cover comments/heartbeats, empty data, multiline data, consecutive
  records, malformed fields, overflow, and recovery/error terminal behavior.
- The codec performs no provider-specific JSON parsing, lifecycle validation,
  accumulation, sequence generation, persistence, or replay.
- Public docs include examples for decoding from `BodyMode::Stream` and
  encoding chunks for `StreamingTerminalResponse`.

## Non-Goals

- HTTP transport, upstream reads, downstream writes, or background producer
  queues.
- OpenAI, Anthropic, A2A, or other provider event types.
- Response event journals or `starting_after` semantics.
- Changes to Pingora H1/H2 framing or the streaming terminal lifecycle.

## Parent Epic

#985

leseb:
### What does this PR do?

Adds a provider-neutral, bounded, incremental Server-Sent Events (SSE) codec to `praxis-filter`, exposed as `praxis_filter::sse`. `SseDecoder` frames records from streamed body chunks under configurable byte/field limits, tolerating cross-chunk splits, CR/CRLF/LF terminators, and a leading UTF-8 BOM, and it poisons itself on a limit violation so a caller cannot silently lose framing. A validated `SseRecordBuilder` plus `encode`/`encode_into` serialize locally generated `SseRecord`s to canonical wire bytes. It is a leaf codec: Pingora keeps ownership of transport streaming, backpressure, and H1/H2 framing, while provider JSON, `[DONE]`, and lifecycle concerns stay in the consumer. There are no consumers yet, so this adds the library only.

Fixes #986

### Checklist

- [x] Tests added or updated
- [x] Documentation updated (if applicable) — full rustdoc plus 4 runnable doctests on the module; no operator/reference docs apply since no filter consumes it yet
- [x] `make lint && make test` passes locally — `make lint` and `make build` pass, and the filter crate's 2114 unit tests + 65 doctests pass. The full-workspace `make test` is green except a pre-existing, unrelated flaky test `tls::watcher::tests::watcher_reloads_on_file_change`, which fails only under full-suite parallel load (fixed 2000 ms filesystem-watch window) and passes 262/262 in isolation. This PR touches only `filter/src/sse/*` and one line of `filter/src/lib.rs`, so it has no causal path to the TLS watcher.

No. It adds a new, additive module (`praxis_filter::sse`) with no changes to existing code and no current consumers.


filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,964 @@
...
+    fn decode_errors_are_copy_and_boxable() {
+        let err = SseDecodeError::LineTooLong { size: 20, limit: 10 };
+        let copy = err;
+        assert_eq!(err, copy, "SseDecodeError is Copy");
+        assert!(err.to_string().contains("20"), "Display includes the size");
+        assert!(err.to_string().contains("10"), "Display includes the limit");
+        let _boxed: Box<dyn std::error::Error + Send + Sync> = err.into();
+    }
+
+    // ----- Test Utilities -----
```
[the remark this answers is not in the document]
leseb: d81b62e9

filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,964 @@
...
+
+    // Module-scoped BOM constant for the tests below.
+    const BOM_BYTES: &[u8] = &[0xEF, 0xBB, 0xBF];
+
+    // Push one chunk, assert no error, return the completed records.
+    fn push_ok(decoder: &mut SseDecoder, chunk: &[u8]) -> Vec<SseRecord> {
```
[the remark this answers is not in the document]
leseb: d81b62e9

crates/filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+
+use super::record::{SseField, SseRecord};
+
+// -----------------------------------------------------------------------------
+// Constants
+// -----------------------------------------------------------------------------
+
+/// UTF-8 byte order mark, stripped once at stream start.
+const BOM: [u8; 3] = [0xEF, 0xBB, 0xBF];
+
```
shaneutt, line 16: Missing some of the standard comment separators throughout:


```suggestion

// -----------------------------------------------------------------------------
// DecoderState
// -----------------------------------------------------------------------------

```

filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+}
+
+/// Bounds on retained memory. Enforced continuously as bytes accumulate.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct SseLimits {
+    /// Max bytes for a single field line held across chunks.
+    pub max_line_bytes: usize,
+    /// Max total retained bytes for one in-progress record: the sum of every
+    /// field's value plus every `Unknown` field name, so large unknown names
+    /// cannot bypass the limit.
```
shaneutt:
```suggestion
    /// Max bytes retained for the committed fields of one in-progress record:
    /// the sum of every field's value, plus the name of every `Unknown` field
    /// so large unknown names cannot bypass the limit. Resets when the record
    /// is dispatched.
    ///
    /// The line currently being accumulated is *not* counted until it commits;
    /// it is bounded separately by `max_line_bytes`. Peak memory the decoder
    /// can hold for a single record is therefore
    /// `max_record_bytes + max_line_bytes`.
```

filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+    pub max_record_bytes: usize,
+    /// Max number of fields in one record (bounds per-field allocations so many
+    /// tiny fields cannot evade `max_record_bytes`).
+    pub max_fields_per_record: usize,
+}
+
+impl Default for SseLimits {
+    fn default() -> Self {
+        Self {
+            max_line_bytes: 10_485_760,   // 10 MiB
```
shaneutt: This seems too high as a default, should we start with 1 MiB instead and callers can raise it via `SseLimits`?

crates/filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+    pub max_fields_per_record: usize,
+}
+
+impl Default for SseLimits {
+    fn default() -> Self {
+        Self {
+            max_line_bytes: 10_485_760,   // 10 MiB
+            max_record_bytes: 10_485_760, // 10 MiB
+            max_fields_per_record: 4096,  // far above any real record
+        }
```
shaneutt, line 70: These values should be in constants (more consistent with the rest of the codebase)

filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+            record_bytes: 0,
+            state: DecoderState::Active,
+            bom_len: 0,
+            bom_resolved: false,
+        }
+    }
+
+    /// Feed one body chunk; returns the records it completed and an optional
+    /// error.
+    pub fn push(&mut self, chunk: &[u8]) -> SseBatch {
```
shaneutt: The signature here is something I would wanna get settled early since it's the expensive thing to change once filters depend on it 🤔

Pingora gives body filters a `Bytes`, and by taking `&[u8]` we're committing to copying every field value twice: once into `line_buf`, then again with `Bytes::copy_from_slice` when the line is classified, plus a fresh `Vec<SseField>` allocation per record because fields gets `mem::taken`.

For an LLM token stream that's roughly three allocations per token. If push took `Bytes` instead, any field that lands entirely inside one chunk (which is nearly all of them in practice) could just be a `slice_ref` into the chunk, and we'd only fall back to copying when a line actually straddles a chunk boundary.

It would be OK in my book to have this be a follow-up for later and keep the internals  as they are for now I just don't want the public signature to rule that out for us to optimize later.

filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+            return SseBatch {
+                records: Vec::new(),
+                error: Some(err),
+            };
+        }
+        let mut records = Vec::new();
+        match self.parse(chunk, &mut records) {
+            Ok(()) => SseBatch { records, error: None },
+            Err(err) => {
+                self.state = DecoderState::Poisoned(err);
```
shaneutt: Something we might optimize: When we poison `line_buf` and fields are left as-is so a decoder that just hit a limit is holding onto up to `max_line_bytes` + `max_record_bytes` until someone drops it. That's fine if the consumer drops it immediately, but if a filter keeps the decoder in a per-connection ctx (which is the natural place for it) that memory sits for the life of the connection.

Since a poisoned decoder can't produce anything further anyway, I think I would just clear both buffers as part of the transition? 🤔

crates/filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+        matches!(self.state, DecoderState::Poisoned(_))
+    }
+
+    /// Signal end of stream, transitioning `Active` to `Finished`.
+    ///
+    /// Processes any buffered partial line as if terminated and, if the
+    /// in-progress block accumulated any field, returns it as a single trailing
+    /// record. Idempotent: a second call returns an empty batch. While poisoned,
+    /// re-reports the limit error.
+    pub fn finish(&mut self) -> SseBatch {
```
shaneutt, line 254: I think this one might be worth changing before anything consumes it:

If the upstream connection drops in the middle of a `data:` line, the `finish` method currently takes whatever's buffered, treats it as a terminated line, and hands it back as a normal record.

From the consumer's side that's indistinguishable from a real one, so a half-written JSON payload would get parsed (or forwarded to the client) as though it were complete?

The event-stream spec actually says the opposite: "an incomplete event at EOF is discarded, not dispatched". That feels right, WTDY? If I'm not misunderstanding, we need a mechanism here so the caller doesn't lose framing without knowing.

I'd rather finish say "there was a partial record at EOF" explicitly, either as an `SseDecodeError::Truncated` alongside the completed records, or as a separate trailing field on `SseBatch`, and let the caller decide whether that's an error or something to salvage.

crates/filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+    /// Reset to `Active`, clearing buffers and any `Finished`/`Poisoned` state.
+    pub fn reset(&mut self) {
+        self.line_buf.clear();
+        self.fields.clear();
+        self.record_bytes = 0;
+        self.prev_cr = false;
+        self.state = DecoderState::Active;
+        self.bom_len = 0;
+        self.bom_resolved = false;
+    }
```
shaneutt, line 331: If you wanted to ensure that this reset doesn't let new fields on the struct slip past in the future, you can use a destructuring instead of the type fields directly:

```Rust
let Self { line_buf, fields, record_bytes, prev_cr, state, bom_len, bom_resolved, limits = _ } = self;
line_buf.clear();
```

Notably _not_ using the range operator (`..`) in the destructuring so any future field added that doesn't get a reset implementation gets a compile-time error.

All this said, I don't think we actually _want_ a reset method here: you would normally implement a reset method for something that's on the hot path with allocations so you can reduce memory thrash, however, that isn't the case here:

`with_limits` is already allocation-free (`Vec::new()` doesn't allocate), so constructing a fresh decoder costs the same as resetting one. The only capacity reset could preserve is `line_buf` (`fields` is `mem::taken` on every dispatched record, so its allocation never survives anyhow) and a `decoder` lives for a whole stream, so construction happens per stream, not per chunk. A couple of `line_buf` reallocs per streamed response isn't measurable given this context.

Semantically this also is a bit perplexing: the main thing reset does is "unpoison" or "unfinish", but once a limit violation has lost framing, resuming the same stream just yields garbage? The only legitimate use is starting a new stream, and a fresh `with_limits(self.limits)` does that without offering an escape hatch. With no consumers yet, I might drop reset entirely? If you want to keep the name, make it *self = Self::with_limits(self.limits), which also sidesteps the field-drift problem above, but IDK if that's really worth it over just having the caller make their own?

crates/filter/src/sse/decoder.rs
```diff
@@ -0,0 +1,966 @@
...
+                return self.feed(chunk, records);
+            }
+        }
+        self.bom_resolved = true;
+        self.bom_len = 0;
+        self.feed(chunk.get(i..).unwrap_or_default(), records)
+    }
+
+    /// Consume a byte slice, appending completed records to `records`.
+    fn feed(&mut self, bytes: &[u8], records: &mut Vec<SseRecord>) -> Result<(), SseDecodeError> {
```
shaneutt, line 369: This is the loop that runs for every byte of every streamed response, and right now it's a `Vec::push` plus a limit check per byte: so scanning ahead for the next "\n/\r" with `iter().position` and then `extend_from_slice`-ing the whole run with a single length check per run would be a lot cheaper and slightly simpler to read.

crates/filter/src/sse/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+//!     });
+//! ```
+
+mod decoder;
+mod encoder;
+mod record;
+
+pub use decoder::{SseBatch, SseDecodeError, SseDecoder, SseLimits};
+pub use encoder::{encode, encode_into};
+pub use record::{SseBuildError, SseField, SseRecord, SseRecordBuilder};
```
shaneutt, line 199: I appreciate the generous rustdocs in the `mod.rs` 🙇

shaneutt (changes requested):
Overall approach looks good, thank you @leseb!

I have comments about some of the specifics to resolve, and then 👍

filter/src/sse/decoder.rs
leseb: sounds good

crates/filter/src/sse/decoder.rs
leseb: https://github.com/praxis-proxy/praxis/pull/1060/changes/2a12cd203b791a6687454efa9ac7889103701ea3

filter/src/sse/decoder.rs
leseb: https://github.com/praxis-proxy/praxis/pull/1060/changes/2a12cd203b791a6687454efa9ac7889103701ea3

filter/src/sse/decoder.rs
leseb: https://github.com/praxis-proxy/praxis/pull/1060/changes/2a12cd203b791a6687454efa9ac7889103701ea3

crates/filter/src/sse/decoder.rs
leseb: https://github.com/praxis-proxy/praxis/pull/1060/changes/2a12cd203b791a6687454efa9ac7889103701ea3

crates/filter/src/sse/decoder.rs
leseb: Sounds good. I trust your judgment on this potential performance improvement.

crates/filter/src/sse/decoder.rs
leseb: You're right WHATWG [spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) says incomplete EOF events are not dispatched and that also makes more sense.

filter/src/sse/decoder.rs
leseb: Ok will release/take the buffers when poisoning.

filter/src/sse/decoder.rs
leseb: Nah let's do it now if it's presented as Bytes, let's update the signature then.

crates/filter/src/sse/decoder.rs
leseb: Thanks for the detailed comment! Will remove the reset() from the API.

filter/src/sse/decoder.rs
leseb: a8bd656da5cfbcecf9b43226be54967f98d2a20e

filter/src/sse/decoder.rs
leseb: 010327682444437200f9a2c239419609e9a78a71

crates/filter/src/sse/decoder.rs
leseb: 6610c744c8f070007979c7864aa9f8a8f9843ace

crates/filter/src/sse/decoder.rs
leseb: f3bf07ac2382dde50b710fecd05be95e08d1e8a6

crates/filter/src/sse/decoder.rs
leseb: bf7dea8e7f4da090581bd11b57112919604afcf6

filter/src/sse/record.rs
```diff
@@ -0,0 +1,475 @@
...
+                name.is_empty() || name.iter().any(|&b| matches!(b, b':' | b'\r' | b'\n')) || is_known_name(name);
+            if bad_name {
+                return Err(SseBuildError::InvalidFieldName);
+            }
+        },
+    }
+    Ok(())
+}
+
+/// Build-time validation errors for `SseRecordBuilder::build`.
```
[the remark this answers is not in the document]
leseb: 9488c542809f358780b37dbacefdfce99319be8c

shaneutt (dismissed):
needs conflict resolution then I think we're about ready to go

shaneutt (approved):

Issue #189, opened 2026-05-13
Epic: Expressive Filter Conditionals


mariusdanciu:
Filter conditions need to be more flexible and expressive for admin use. The current `when` and `unless` YAML properties do not offer flexible enough predicates to describe when a filter must or must not be applied for request and response paths.

## Current Limitations

The existing `when`/`unless` system supports simple field-value matching but lacks:

- **No boolean composition**: conditions cannot be combined with AND, OR, or NOT; each condition is evaluated independently
- **No regex matching**: headers and paths can only be matched with exact or prefix/suffix strings, not regular expressions
- **No numeric comparisons**: cannot express status code ranges (e.g., 4xx, 500-599) or content-length thresholds
- **No body field access**: cannot inspect JSON body fields or use JSON path expressions in conditions
- **No cross-phase conditions**: request-phase conditions cannot reference response status or headers, and vice versa
- **No set membership**: cannot check if a header value is one of a configured list of values
- **No variable interpolation**: cannot reference dynamic values or previously computed filter results in conditions

## Desired Capabilities

1. **Boolean composition (AND, OR, NOT)**: combine multiple predicates with logical operators for complex matching
2. **Regex matching**: match headers, paths, query parameters, and other string fields against regex patterns
3. **Numeric comparisons**: express ranges and thresholds for status codes, content lengths, and other numeric fields
4. **Body field access**: JSON path expressions to inspect request/response body fields in conditions
5. **Request + response cross-phase conditions**: reference request attributes in response-phase conditions and vice versa
6. **Set membership**: check whether a header or field value belongs to a configured set
7. **Variable interpolation**: reference dynamic values, environment variables, or filter result set entries in conditions

### Option A: Expanded Predicate DSL

Extend the current `when`/`unless` YAML syntax with richer predicate operators. This keeps configuration familiar and avoids introducing a new language, but may become unwieldy for complex expressions.

### Option B: CEL Integration

Adopt the Common Expression Language (CEL), an industry-standard expression language used by Envoy, Kubernetes, and Google Cloud IAM. CEL provides a well-defined, sandboxed evaluation environment with strong typing. Adds a dependency but offers proven semantics and broad ecosystem familiarity.

### Option C: Hybrid

Use simple predicates (exact match, prefix, suffix) for common cases and CEL for complex expressions. This provides an easy on-ramp while supporting advanced use cases.

## Prior Art

- Spike #191: POCs and benchmarks for filter chain conditionals
[...]

Issue #187, opened 2026-05-12
Spike: Praxis ExtProc Deployment Mode


shaneutt:
Look into deploying Praxis as an Envoy [ExtProc](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter) as an alternative to the standard server mode, such that Praxis filters can be run with Envoy.

shaneutt:
After some investigation, it turns out to be straightforward and I was able to get a POC up quickly. I have created https://github.com/praxis-proxy/extproc to continue the effort.

Issue #186, opened 2026-05-12
Spike: configurable operator-managed reserved header prefixes


nerdalert:
## Summary

  PR #180 adds built-in reserved internal header prefixes for agentic protocol metadata:

  - `x-praxis-*`
  - `x-mcp-*`
  - `x-a2a-*`

  These are currently hard-coded so Praxis can reject client-supplied spoofing attempts and strip internally generated metadata before forwarding upstream. We
  should define whether operators need an additive configuration mechanism for deployment-specific reserved prefixes.

  ## Context

  The current TODO is in:

  `protocol/src/http/pingora/handler/reserved_headers.rs`

  ## Questions To Resolve

  - Should Praxis support operator-managed reserved header prefixes?
  - Should custom prefixes be global, listener-scoped, filter-chain-scoped, or some combination?
  - Should custom prefixes only add to the built-in set, or should operators be able to override built-ins?
  - Should this apply to both client request rejection and upstream stripping?
  - What validation should config enforce for prefix values?
  - Should an insecure strip-and-continue mode exist, or should spoofed reserved headers always fail fast?

  ## Acceptance Criteria

  - Decide whether configurable reserved prefixes are needed.
  - Define the config shape if needed, for example:

  reserved_header_prefixes:
    - x-mycompany-
    - x-secret-squirrel-

  - Document scope and semantics: additive vs override, global vs listener/filter-chain.

PR #1096, opened 2026-09-04
feat(policy)!: update Praxis to use PPE 0.2.0

araujof:
### Description

This moves the policy filter to the released `ppe` 0.2.0 crate and removes the temporary path dependency. It also adds the Praxis support needed for request and response assertions and route-specific authentication.

### Praxis changes

- Give the engine an `HttpTransport` for JWKS, OAuth token exchange, and CIBA. The new `allow_private_idp` option supports local and in-cluster identity providers.
- Run ordinary HTTP traffic through the `http.request` hooks. The previous CMF path had no chat message, so content scanners could inspect an empty body and report it as clean.
- Apply request assertions to upstream requests and response assertions before sending response headers to the client.
- Pass the HTTP method and path to identity hooks so route-specific `authentication:` rules work.
- Preserve JSON claim values returned by 0.2.0.
- Deny requests without identity metadata as `unidentified_request`.
- Update fixtures and policy files for removed settings and the new `dispatch: policy` default.

### Bugs fixed during the upgrade

1. Route-specific authentication was skipped because identity resolution ran before the request method and path were available. A route using `replace_inherited: true` could therefore get the weaker global authenticator.
2. Asserted headers were not authoritative. Duplicate or differently cased input headers could survive or cause an asserted header to be removed.
3. Header changes queued by earlier filters could override policy assertions.
4. Response assertions were accepted but never run because Praxis did not dispatch `http.response`.

Headers controlled by `assertions:` are now taken directly from the policy result. Other headers still use the existing diff, so delegation and unrelated filters keep working. Entity-level response assertions are rejected at load time because their CMF hook runs after response headers have already been sent.

### praxis-policy 0.2.0 highlights

- Assertions at global, default, bundle, and route scope.
- HTTP route matching by method and path, plus `http.response` and `http.status` hooks.
- Configurable JWT claim mapping and set-valued roles and permissions.
- Cached delegated tokens with a size limit.
- A host-provided HTTP transport with HTTP/2, response-size limits, safe retries, and blocked-address checks.
- Tests that compare behavior across Cedar, CEL, and OPA.
- `docs/apl-grammar.md` as the policy language reference.

The main breaking change is that `dispatch:` now defaults to `policy`. Declaring a plugin that no policy uses is now a load error. Several obsolete settings are also rejected. Quoted strings use one escaping rule, and invalid attribute paths, numbers, and `not` expressions fail validation instead of quietly producing no value.

Parallel branches now deny access if they panic. Invalid numeric comparisons, unreadable handler results, and JWT issuers without audiences also fail safely. Token-exchange denials no longer expose the identity provider's response body. See the engine `CHANGELOG.md` for the full list.

### Testing

- All praxis tests pass.
- The praxis policy engine demo passes all 12 scenarios on Cedar, CEL, and OPA. Upstream logs confirm that asserted headers have the expected values, `x-user-token` is removed, spoofed `x-auth-user-id` values do not reach the backend, and delegated tokens still do.

## PR size

This is over the 500-line PR limit, but the commits can become stacked PRs with smaller additions if needed.

## Policy Engine Epic (#678)

This release covers most of [agent credential isolation #189](https://github.com/praxis-proxy/ai/issues/189) and [external service access #188](https://github.com/praxis-proxy/ai/issues/188). It provides part of the work for tool policies (#185), policy chains (#184), A2A delegation (#187), data sovereignty (#190), inference access (#186), A2A method routing (#447), and inference routing (#448).


shaneutt (changes requested):
Let's get the rebase for this and fix the conflicts, if you need help please lemme know 🖖

crates/filter/src/builtins/http/security/policy/assertions.rs
```diff
@@ -0,0 +1,409 @@
...
+
+use ppe::praxis_policy_core::{
+    assertions::{Direction, StripPattern},
+    config::PolicyConfig,
+    hooks::Extensions,
+};
+
+use super::filter::PolicyFilter;
+use crate::filter::HttpFilterContext;
+
```
shaneutt, line 19: missing the standard comment separators here, and several other places.

crates/filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -402,15 +476,29 @@ impl PolicyFilter {
...
-    /// Build the public raw-credential-free projection from PPE's private
-    /// validated payload.
+    /// Install the bundled transport with the configured destination policy.
+    fn install_http_transport(mgr: &Arc<PolicyEngine>, allow_private: bool) -> bool {
+        if allow_private {
+            tracing::info!(
+                target: "policy.filter",
+                "policy: allowing the engine to reach private and loopback IdP addresses"
+            );
+            return mgr.set_http_transport(Arc::new(ppe::HyperTransport::new().with_allow_private_destinations()));
```
shaneutt, line 486: We need to comment this, and make sure we have some kind of core-level follow-up to drop Hyper in favor of the builtin.

crates/filter/src/builtins/http/security/policy/filter.rs
```diff
@@ -950,6 +1065,88 @@ impl HttpFilter for PolicyFilter {
         Ok(FilterAction::BodyDone)
     }
 
+    #[expect(
+        clippy::too_many_lines,
+        reason = "linear response-phase flow: rebuild identity, dispatch, apply the contract"
+    )]
```
shaneutt, line 1071: </img>

shaneutt (approved):
Question, and then some comments to resolve, then LGTM:

Are we literally covering everything the policy engine can do with our examples?

crates/filter/src/builtins/http/security/policy/filter.rs
araujof: yep, already have the issue drafted... will post it.

araujof: https://github.com/praxis-proxy/praxis/issues/1098

crates/filter/src/builtins/http/security/policy/assertions.rs
araujof: Addressed in [a separate PR. ](https://github.com/praxis-proxy/praxis/pull/1099)

crates/filter/src/builtins/http/security/policy/filter.rs
araujof: Addressed in https://github.com/praxis-proxy/praxis/pull/1099

Issue #1067, opened 2026-09-01
Epic: Operations & Support


shaneutt:
## Summary

Praxis is a general-purpose L4/L7 proxy, but its architecture makes its resource behavior fundamentally different from a "classic" proxy. Body access and buffering, a composable filter pipeline (compression, decompression, body rewriting, payload processing, AI filters), in-memory caching and counters, embedded Wasm runtimes, and hot-reload (which briefly holds two pipelines at once) all mean memory can be large and, more importantly, can spike hard and unpredictably with the *shape* of traffic rather than just request count. A proxy sized purely on CPU or connection count will get surprised. Before v1 we need a firm, measured, and documented grasp of these characteristics: how memory and CPU behave across features and traffic profiles, what signals autoscaling must key on, and exactly what resources an operator should provision at a given scale.

This is broader than raw throughput/latency benchmarking (see #453). It is about characterizing the full operational envelope, proving it with automated tests, exposing the right signals, and shipping documentation an operator can plan against. #453 can serve as the raw performance-benchmark input to this work.

## Goals

- Measure and document Praxis's memory and CPU behavior across representative feature sets and traffic profiles, with particular emphasis on worst-case and spike behavior.
- Define the signals operators should autoscale on (CPU alone is insufficient for a memory-spike-prone proxy) and how to export them.
- Publish resource-sizing guidance at specific, named scales.
- Back all of the above with automated performance, soak, and spike tests that run on a regular cadence and gate regressions.
- Ship operator-facing documentation prior to v1.

### Memory behavior
- Baseline (idle) footprint per worker/process and how it scales with worker count.
- Per-connection and per-request memory cost across HTTP/1.1, HTTP/2, and HTTP/3.
- Body-buffering memory: request/response body access, buffering thresholds, and configured max buffered body sizes.
- Memory amplification from payload-processing and transformation filters (compression, decompression, body rewrite, JSON transforms) and how it compounds when chained.
- Streaming vs buffering paths (SSE codec, streaming compression) and their contrasting memory profiles.
- Cache storage backend memory (in-memory response cache), growth and eviction behavior under pressure.
- Distributed in-memory counters and KV-store growth over time.
- Worst-case and tail spikes: large bodies, many concurrent large requests, slow clients (slowloris-style), and backpressure buildup.
- Memory under connection-pool growth, keepalive, and H2 stream multiplexing.
- Wasm filter memory (fuel and memory limits), per-instance overhead, and pool sizing.
[...]

Issue #154, opened 2026-05-01
Body Transformation Filter


shaneutt:
Transform request and response bodies using template or expression-based rules.

Scope:
- JSON field manipulation: set, remove, rename fields
- Template-based body generation from request attributes
- Expression-based field computation
- Request and response body transformation
- MIME type filtering (apply only to JSON bodies, etc.)

aviweit:
Hi @shaneutt, (cc @eranra) we opened an issue (https://github.com/praxis-proxy/praxis/issues/766) that describes a lack of filter functionality that we would like to have and believe this issue can address it. 

Thanks a lot.

Issue #788, opened 2026-07-10
Expand default compression content-type allowlist


shaneutt:
## Summary

The compression filter's default `content_types` allowlist uses a `text/` prefix pattern that matches all `text/*` subtypes via `starts_with`. This inadvertently includes `text/event-stream` (SSE), where compression is harmful: buffering for compression destroys real-time delivery semantics critical for AI streaming workloads.

There is currently no way to exclude specific content types without replacing the entire allowlist.

Additionally, the defaults miss compressible types outside the `text/` prefix:

- `application/yaml` — config and API responses
- `application/graphql+json` — GraphQL over HTTP
- `application/graphql-response+json` — GraphQL response content type (per GraphQL over HTTP spec)

## Design

- Add an `exclude_content_types` config field that takes precedence over the allowlist (e.g., compress all `text/*` except `text/event-stream`)
- Add the three `application/*` types above to the default allowlist in `compression_config.rs`

## Acceptance criteria

- [ ] `exclude_content_types` config option works and is validated
- [ ] Exclusions take precedence over allowlist matches
- [ ] Unit tests for exclusion logic (exact match, prefix interaction, empty exclusion list)
- [ ] `application/yaml`, `application/graphql+json`, and `application/graphql-response+json` added to defaults
- [ ] Existing compression tests still pass
- [ ] Filter reference docs updated

PR #1110, opened 2026-09-08
feat(filter): add json_body filter and JsonOps API

### LINKED ISSUE #766: Missing Generic JSON Body Override
### Description

# Praxis: Missing Generic JSON Body Override

## Context / Need

On the `llm-egress` listener (port 8081), our [skillberry-agent-praxis POC](https://github.com/skillberry-ai/skillberry-agent-praxis-poc) which is based on Praxis - must enforce routing policy by overriding client-supplied values before forwarding to the upstream LLM provider:

| What | Source | Status |
|------|--------|--------|
| Overwrite `model` in body | `SPAPRAXIS_MODEL` env var | **Blocked** — no filter available |
| Overwrite `temperature` in body | `SPAPRAXIS_TEMPERATURE` env var | **Blocked** — no filter available |
| Overwrite `Authorization` header | `SPAPRAXIS_API_KEY` env var | Done — `credential_injection` filter (currently using `OPENAI_API_KEY`, env var name will change) |

## The Gap

No existing Praxis filter can declaratively set/overwrite arbitrary top-level fields
(`model`, `temperature`, etc.) in a `POST /v1/chat/completions` JSON body.

We believe that the below opened issue is relevant:
**[praxis-proxy/praxis#154 — Body Transformation Filter](https://github.com/praxis-proxy/praxis/issues/154)**

- Milestone: **v0.5.0**
- Author: shaneutt
- Planned scope: JSON field set/remove/rename, template-based body generation, expression-based field computation, MIME type filtering
- Status: Open, not yet in progress

## Current Workaround

Using a local workaround in our worker service until #154 lands upstream.

### Motivation

Our Praxis-based skillberry-agent-praxis [deployment](https://github.com/skillberry-ai/skillberry-agent-praxis-poc) must enforce routing and inference policies independently of client-provided request parameters.
While Praxis already provides a mechanism to override authentication credentials through the credential_injection filter, there is currently no equivalent way to enforce policy-controlled values in the JSON request body. 

### Praxis filters that exist today that do not meet our needs

| Filter | Body access | Limitation |
|--------|-------------|------------|
| `json_body_field` | Read-only | Extracts field to header, does NOT write |
| `model_to_header` | Read-only | Delegates to `json_body_field` |
| `openai_responses_model_rewrite` | Read-write | `/v1/responses` path only |
| `prompt_enrich` | Read-write | `messages[]` array only |
| `ext_proc` | Read-write | Requires external gRPC service |

### Alternatives Considered

_No response_

### Area

None

### Parent Epic (optional)

_No response_

aslakknutsen:
## Summary

Adds JSON Pointer body mutation in two forms:

1. **`json_body` filter** — YAML-configured HTTP filter for operators.
   Request: add, remove, replace, extract. Response: remove and extract
   (add/replace rejected; `Content-Length` already committed).

2. **`JsonOps` API** — programmatic op builder in `praxis_filter::json_ops`
   for other filters and Rust code. Same tokenizer engine as the filter;
   no DOM, no serde round-trip for the document.

### Filter (`json_body`)

- RFC 6901 pointers in one pass over a `StreamBuffer` body
- Value sources: static JSON, `filter_metadata`, structured metadata,
  `env_var` (read at construction)
- Extract destinations: metadata, structured metadata, request header
  (`extra_request_headers`)
- `on_invalid`: `continue` | `reject` | `error`
- Example: `examples/configs/payload-processing/json-body.yaml`

### API (`JsonOps`)

- `JsonOps::builder()` → `.extract()` / `.add()` / `.replace()` /
  `.remove()` → `.build()`
- `JsonValue` sources and `ExtractDest` mirror filter YAML semantics
- `JsonOps::apply(body, store)` for extract + optional rewrite in one walk
- Exported from `praxis_filter`; `JsonBodyFilter::from_ops()` for tests
  and programmatic filter construction

### New crate: `praxis-simd-scan`

`praxis-simd-scan` is a new published crate (`crates/simd-scan/`) that
provides a single function, `find_json_string_delim`, which finds the first
byte in a slice that is a JSON string delimiter: `"`, `\`, or any control
character (< 0x20).

It has three backends selected at compile time via `cfg(target_feature)`:
- **x86_64 (SSE2)**: 16-byte vectorised loop using `_mm_loadu_si128` / `_mm_cmpeq_epi8` / saturating subtract for control-char detection.
- **aarch64 (NEON)**: 16-byte vectorised loop using `vld1q_u8` / `vceqq_u8` / `vcltq_u8`, with a `vshrn_n_u16` movemask emulation for first-match extraction.
- **Scalar fallback**: `iter().position()` for all other targets.

The crate is consumed by `praxis-proxy-filter`'s `json_ops::skip` module,
where it replaces the inner loop of `skip_string_with_meta` to accelerate
JSON string scanning during the single-pass rewrite walk.

**Unsafe code**: the crate contains ~594 lines including SIMD intrinsic
calls that require `unsafe` blocks. Each block has a structured safety
comment citing the Rust Reference and std pointer contracts. The workspace
`lints` table `forbid`s `unsafe_code`, so this crate duplicates the
workspace lint rules locally with `unsafe_code = "deny"` instead, using
targeted `#[expect(unsafe_code)]` on the specific functions that need it.

Tests verify every backend agrees with the scalar definition (via
`assert_both`) across boundary offsets, all 256 byte values, and
misaligned base pointers.

### Other changes

- Duplicate-key semantics, array `/-` on objects, response growth guards
- Generated filter docs, features.md, Criterion + dhat benchmarks

## Benchmark results

Workload: OpenAI-style chat completion JSON at **10 KiB**, **256 KiB**,
**512 KiB**, and **10 MiB**. Ops: extract `/model`, replace `/model`, add
`/tenant` + `/original_model`, remove `/secret`.

| Layout | Meaning |
|--------|---------|
| **prefix** | `secret` near document start |
| **spread** | `secret` after `messages` (full scan) |

### Speed (`cargo bench -p praxis-tests-benches --bench json_body`)

**Tokenizer (shipped `JsonOps` path)** — ~**1.5–1.8 GiB/s**.

**DOM reference** (not shipped) — ~**115–162 MiB/s**.

| Case | Tokenizer | DOM | vs DOM |
|------|-----------|-----|--------|
| prefix / 10 KiB | 1.49 GiB/s | 141 MiB/s | **~10.8×** |
| prefix / 256 KiB | 1.77 GiB/s | 158 MiB/s | **~11.5×** |
| prefix / 512 KiB | 1.77 GiB/s | 161 MiB/s | **~11.3×** |
| prefix / 10 MiB | 1.66 GiB/s | 117 MiB/s | **~14.5×** |
| spread / 10 KiB | 1.53 GiB/s | 145 MiB/s | **~10.8×** |
| spread / 256 KiB | 1.73 GiB/s | 159 MiB/s | **~11.1×** |
| spread / 512 KiB | 1.77 GiB/s | 161 MiB/s | **~11.3×** |
| spread / 10 MiB | 1.69 GiB/s | 115 MiB/s | **~15.0×** |

Throughput is Criterion's point estimate (middle of the `thrpt` range).
Speedup = tokenizer ÷ DOM (GiB/s converted to MiB/s at 1024).

### Heap (`json_body_heap --features dhat-heap`, 20 iters)

| Size | Tokenizer peak | DOM peak | Ratio |
|------|---------------:|---------:|------:|
| 10 KiB | 11,030 B | 93,821 B | ~8.5× |
| 256 KiB | 263 KB | 2.4 MB | ~9.5× |
| 512 KiB | 525 KB | 4.7 MB | ~9.5× |
| 10 MiB | 10.0 MB | 89.0 MB | ~8.9× |

Tokenizer peak ≈ output size + overhead; DOM builds full `Value` trees.

### Removed DOM reference benchmark

The earlier commits in this branch included a `json_body_dom` Criterion
benchmark group (`dom_apply_request`) alongside the tokenizer benchmark.
It was removed in `ffca9d54` ("split internals and drop dev benchmarks").

Context on what it was and why it existed:

- `json_body_field` was originally DOM-based (`serde_json::Value` parse →
  tree walk → serialize). It was later rewritten to stream-extract without
  building a full JSON DOM (see `cbf3ca20`). `json_body_field` only supports
  extract (header promotion from top-level fields); it has no add/remove/replace.

- The DOM benchmark simulated what expanding the *old* `json_body_field`
  approach to support add/replace/remove would cost: parse the full document
  into `serde_json::Value`, apply pointer operations on the tree, then
  `serde_json::to_vec` the result. This is also representative of the cost
  paid by *every other filter* in the codebase that does any JSON
  extraction or modification through a DOM round-trip (e.g., `json_rpc`
  envelope parsing before its streaming rewrite).

- The benchmark served its purpose during development to validate the
  single-pass tokenizer approach. Once the tokenizer was stable and the
  numbers were captured, the DOM path was dropped to avoid maintaining dead
  code.

## Test plan

- [ ] `make test`
- [ ] `make test-integration` (`payload_examples::json_body_*`)
- [ ] `cargo test -p praxis-proxy-filter --lib json_body`
- [ ] `cargo test -p praxis-proxy-filter --lib json_ops`
- [ ] `cargo xtask lint-filter-docs`

## Additional notes

This filter can generally replace the json_body_field filter.
This API can generally replace other filters json manipulation needs. 

Fixes #766 
Related to #154 
Related to #1067


crates/filter/src/builtins/http/payload_processing/json_body/mod.rs
```diff
@@ -0,0 +1,353 @@
+// SPDX-License-Identifier: MIT
```
leseb: Apache now

crates/filter/src/builtins/http/payload_processing/json_body/config.rs
```diff
@@ -0,0 +1,297 @@
+// SPDX-License-Identifier: MIT
```
leseb: Apache now

crates/filter/src/json_ops/builder.rs
```diff
@@ -0,0 +1,337 @@
+// SPDX-License-Identifier: MIT
```
leseb: Aapche now

leseb (changes requested):
[P2] skip.rs:92 accepts invalid escapes such as \q and raw control characters. Thus on_invalid: reject can still forward malformed JSON.
[P2] rewrite.rs:283 skips descendant extracts when their parent is removed/replaced. extract("/a/b") + remove("/a") succeeds but leaves metadata empty; root replacement similarly loses root extracts.
[P2] config.rs:250 accepts invalid header names, which are later silently discarded during request pre-read.

crates/filter/src/builtins/http/payload_processing/json_body/mod.rs
aslakknutsen: fixed https://github.com/praxis-proxy/praxis/pull/1110/commits/326ee008bd51dfe6d41f3d15e13e36360eafcb39

crates/filter/src/builtins/http/payload_processing/json_body/config.rs
aslakknutsen: fixed https://github.com/praxis-proxy/praxis/pull/1110/commits/326ee008bd51dfe6d41f3d15e13e36360eafcb39

crates/filter/src/json_ops/builder.rs
aslakknutsen: fixed https://github.com/praxis-proxy/praxis/pull/1110/commits/326ee008bd51dfe6d41f3d15e13e36360eafcb39

leseb (dismissed):
fix lint - otherwise lgtm!

shaneutt (changes requested):
> @alexsnaps @shaneutt thoughts on including the MIT bits and pulling unsafe into this project (isolated in a single crate)?
> 
> [aslakknutsen@5e755e8](https://github.com/aslakknutsen/praxis/commit/5e755e87ccf7a76d0b4169eeb8b3ff2991af0acc)

What other mechanisms do we have to solve the problems that this solves?

crates/filter/src/json_ops/ops.rs
```diff
@@ -0,0 +1,153 @@
+// SPDX-License-Identifier: Apache-2.0
```
shaneutt, line 1: This file and some others don't have the comment separator sections as per the conventions.

crates/filter/Cargo.toml
```diff
@@ -44,6 +45,7 @@ opentelemetry = { workspace = true, optional = true }
 percent-encoding = { workspace = true }
 pingora-core = { workspace = true }
 praxis-core = { workspace = true }
+praxis-simd-scan = { workspace = true }
```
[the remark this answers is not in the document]
aslakknutsen: Dearest Respected [first-name] [last-name],

I hope this unsolicited communication reaches you in magnificent health and peak quarterly synergy!

My name is Aslak from Global Agile Software Growth Ltd, and while doing a complimentary, automated audit of your profile, our proprietary machine-learning crawlers discovered that you are currently reviewing critical engineering deliverables through an un-optimized, low-ROI, sub-fractional code commitment viewpoint!!

By solely scrutinizing individual commits, our diagnostics show that you are missing 100% of the Holistic Value Synergy contained within the Pull Request ecosystem as a totality!

Kindly direct your attention to this revolutionary use-case evidence:
👉 [CLICK SECURE LINK] [https://github.com/praxis-proxy/praxis/pull/1110/changes#diff-96a5423dce743dcb868ae49cf936c73fd8f6d61e1c9b797237d23789057ea266R7](https://github.com/praxis-proxy/praxis/pull/1110/changes#diff-96a5423dce743dcb868ae49cf936c73fd8f6d61e1c9b797237d23789057ea266R7) 👈

Warmest Personal Regards,

Aslak Knutsen (Not A Bot)

Senior Lead Strategic Code Outreach Specialist

twghu:
Some points regarding memory pressure :

# Pressure

Every limit in the path is **per request**:

- `max_body_bytes` (filter, default 10 MiB) — 413 on overflow
  (`stream_buffer.rs:152`)
- `body_limits.max_request_bytes` / `max_response_bytes` (global, default 10 MiB)
  — `clamp_body_mode` takes `min(filter, ceiling)` (`pipeline/mod.rs:604`), so a
  filter cannot exceed the global ceiling
- `ABSOLUTE_MAX_BODY_BYTES` = 64 MiB hard cap even with `allow_unbounded_body`

None of these is an aggregate. There is no counter of total buffered bytes in
flight, no admission control keyed on memory pressure, and no gauge exporting it.
The only relevant metrics are `praxis_http_active_requests` (a count, not bytes)
and `praxis_http_request_body_bytes` (a per-request histogram, recorded after the
fact). So 200 concurrent 10 MiB requests through this filter is ~4 GiB and
nothing in Praxis will shed load — the proxy will OOM before it returns 503.

That is precisely the gap issue #1067 names: *"Which signals to autoscale on:
memory pressure, in-flight requests, buffered bytes, queue depth... (not just
CPU)"* and *"Behavior at memory exhaustion: fail-closed vs crash."* This PR adds
the measurement apparatus for it (the dhat bench) but not the signal or the
control.

**Not a multiplier (verified):** Pingora's retry buffering
(`enable_retry_buffering()`, `stream_buffer.rs:122`) uses a fixed-size buffer
that truncates rather than growing, and the handler guards retries when
`retry_buffer_truncated()` (`with_body.rs:128`). Large bodies simply become
non-retryable rather than doubling memory.

# Suggestions

## Recommendations

1. **Document the 2N peak** in the filter docs. It is the number an operator
   needs to size against and nothing currently states it: peak heap is about
   2 x `max_body_bytes` per concurrent request per mutating direction;
   extract-only is 1x.
2. **Lower `max_body_bytes` in the example** rather than pinning it at the 10 MiB
   default. `examples/configs/payload-processing/json-body.yaml` explicitly sets
   `max_body_bytes: 10485760` for a chat-completions workload — the worst default
   to teach, since real chat bodies are kilobytes.
3. **Export a buffered-bytes gauge** (in-flight sum, incremented in
   `BodyBuffer::push`, decremented on drop). This is the missing #1067
   autoscaling signal and the prerequisite for any later admission control.
4. **Consider releasing `retained_pre_read_body`** once the upstream request is
   non-retryable. This is framework behaviour rather than this PR's, but
   `json_body` is the filter that makes it expensive.

twghu:
Some comments from scan :

  Findings

  1. The Summary's benchmark tables can't be reproduced from this PR. The body
  presents a DOM-vs-tokenizer comparison (~10.8×–15× speed, ~8.5–9.5× heap)
  under cargo bench -p praxis-tests-benches --bench json_body. There is no DOM
  path in the shipped benches — json_body.rs benches only tokenizer_apply in two
  layouts, and json_body_heap.rs:33 hardcodes "tokenizer" as the path label, a
  vestige of a two-path harness dropped in 048db56a. Either restore the
  reference implementation behind the bench or mark the tables as historical
  measurements.

  2. crates/simd-scan isn't in the Summary at all. It's a new published crate
  (publish = true), 594 lines of unsafe SIMD, plus a 180-line hand-copy of the
  workspace lint table because the workspace forbids unsafe_code. That
  duplicated block will drift silently from Cargo.toml — nothing checks the two
  stay in sync. A new published crate with unsafe code deserves its own section
  in the PR body, and arguably its own PR.

  3. PR size vs the project's own gate. CONTRIBUTING.md caps PRs at 750 added
  lines of production code. Excluding tests.rs files this is ~3,630. That's a
  rule the repo enforces in CI, so it needs an explicit override label or a
  split (simd-scan → json_ops → json_body filter is a natural three-way).

  4. The example doesn't prove all its configured variants. CLAUDE.md requires
  "every example must prove its feature works with all configured variants."
  examples/configs/payload-processing/json-body.yaml configures a header extract
  (X-Stream), a request structured_metadata extract (/user), and two
  response_extract destinations. The integration test
  (payload_examples.rs:165-203) asserts none of them — it only checks the echoed
  body. lint-example-tests passes because it only checks a test exists. The echo
  backend would need to reflect request headers for X-Stream, and the
  structured-metadata extracts need a downstream observer.

  5. env_var credential handling diverges from in-repo precedent.
  credential_injection deliberately wraps secrets in secrecy::SecretString "to
  prevent accidental logging" (credential_injection/config.rs:67) and zeroizes
  on drop. The new env_var source stores the value as a plain Bytes in
  ValueSource::Static, and JsonOps/CompiledOpSet/CompiledOp/JsonBodyOps all
  derive Debug — so {:?} on a compiled op set prints the secret. Nothing logs it
  today, so this is latent, not an active leak; but the generated doc's own
  example teaches pointer: /api_key, env_var: TENANT_API_KEY, which is exactly
  the case that should be SecretString.

  6. Two pieces of dead configuration.
  - memchr = "2.7.4" added to [workspace.dependencies] (Cargo.toml:82) is used
    by nothing — the only hits repo-wide are a test name (skip.rs:304) and a doc
    comment. Leftover from before simd-scan. Neither lint-deps nor machete
    catches it (it's workspace-level, no crate depends on it).
  - The bench-internals feature (crates/filter/Cargo.toml:18) gates pub(crate)
    mod test_utils (lib.rs:436). Since the module is pub(crate), no external
    bench can reach it, and nothing enables the feature. It only exists to make
    --all-features builds compile, hence the cfg_attr(…, allow(dead_code)) on
    the next line.

  7. Body bytes are not preserved. Any object level containing an op is
  re-emitted token-by-token (rewrite_object writes key_span : value-span joined
  by ,), dropping inter-token whitespace — pretty_printed_object_rewrites pins
  this ({\n "a": 1,…} → {"a":1}). Levels with no ops are copied verbatim as
  spans, so output is a mix. Consequence worth documenting: a request that
  traverses json_body can no longer satisfy an upstream body-signature check
  (webhook HMAC, SigV4), even when every op is a no-op.

  8. response_remove redacts but does not shrink. fit_response
  (json_body/mod.rs:343) pads with trailing spaces to the committed
  Content-Length. Correct and legal, but it means response removal buys
  redaction, not bandwidth — and it pads unconditionally, including on chunked
  responses where no padding is needed. The docs say "padded with trailing
  spaces"; they don't say the transferred size is unchanged.

  9. No MIME gating. on_request_body/on_response_body run regardless of
  Content-Type. compression has a content_types option for exactly this
  (compression_config.rs:106), so there's precedent for the knob. Today an
  operator must reach for filter conditions. Also: with on_invalid: reject, a
  bodyless request on a matched route hits handle_invalid(…, "empty body") →
  400.

twghu (commented):
I've added some notes about memory pressure and a few points that were detected by a more detailed scan for consideration.

aslakknutsen:
@twghu 

Recommendations

> 1. Document the 2N peak in the filter docs

Added docs https://github.com/praxis-proxy/praxis/pull/1110/changes/964356fbd48845c484df4ab63132187f53490c2a

> 2. Lower max_body_bytes in the example rather than pinning it at the 10 MiB.

Default max_body_bytes is 10 MiB, but changed the example to 1 MiB. https://github.com/praxis-proxy/praxis/pull/1110/changes/c86672f4aa4e7a1a2ea6410c7f997f5cb86a2703

> 3. Export a buffered-bytes gauge 

Outside of scope for this PR but could be considered for a future PR.

> 4. Consider releasing retained_pre_read_body once the upstream request i

Outside of scope for this PR but could be considered for a future PR.


Findings

> 1. The Summary's benchmark tables can't be reproduced from this PR. 

Updated PR description to explain the missing DOM benchmark

> 2. crates/simd-scan isn't in the Summary at all. 

Updated PR description to explain the new praxis-simd-scan crate.

> 3. PR size vs the project's own gate. 

Even just certain parts of the Tokenizer blows that limit. The blanace between small PRs and working software can be tricky.

> 4. The example doesn't prove all its configured variants.

Updated the config structure to include most of the variants.
Resposne extract is still a bit tricky to test, but the example should be enough to prove the concept. As well as the request extract to header requires a custom config file (in the test case) based on how the start_echo backend functions work. A future version could/should probably split this up so both HEADER and BODY are returned in the same call as separate entities. https://github.com/praxis-proxy/praxis/pull/1110/changes/f4cf4a40dcb0506e47d7ff9fb9bc6437400906dc


> 6. Two pieces of dead configuration.

* memchr removed in https://github.com/praxis-proxy/praxis/pull/1110/changes/a47ac0da52a081a4f2b2971b00d9c2ccc17f7cba
* remvoed old bench-internals feature gate https://github.com/praxis-proxy/praxis/pull/1110/changes/b4a38571eec699bafec062d8cfdbdd9890006439

> 7. Body bytes are not preserved.

Extract only ops works without going through rewrite_object and changing the output, also the Filter v indicate BodyRead mode if only extract ops are defined so a Signature should hold in that scenario.

A possible "bug" here is that Add/Replace ops that actually doesn't execute, e.g. added path already exist, or replace that doesn't exist, will change the output. But at that point the user attempted to change the content and it would be expected that the signature would/could break and some form of resign filter would be required.

Added clearifications in docs https://github.com/praxis-proxy/praxis/pull/1110/changes/237add65869ac61e979b1950f8c97dd0ad40fd17

> 8. response_remove redacts but does not shrink.

We could expand the scope of the response support to rely on the Committed Headers and by that allow the response to be shrinked. As for now, relying on original.len should work even tho not optimal in e.g. the case of Chunked Encoding. Clearified the docs for now https://github.com/praxis-proxy/praxis/pull/1110/changes/ae52dcbf9e3c453c193e8dfa3919080c3cabf371

> 9. No MIME gating. on_request_body/on_response_body run

* Added support for MIME gating via content_type config field for on_request_body/on_response_body https://github.com/praxis-proxy/praxis/pull/1110/changes/380bfa8839c8d615d7cb29ba611b2bcfa0971481

* Added support for empty bodies https://github.com/praxis-proxy/praxis/pull/1110/changes/e45f2577af68d897a41cb8a4d7f04bf4431c4d64

crates/filter/src/builtins/http/payload_processing/json_body/mod.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// prefix comparison, following the same convention as the compression
+    /// filter's `content_types` option.
+    fn matches_content_type(&self, content_type: &str) -> bool {
+        if self.content_types.is_empty() {
+            return true;
+        }
+        self.content_types.iter().any(|pattern| {
+            content_type
+                .get(..pattern.len())
+                .is_some_and(|prefix| prefix.eq_ignore_ascii_case(pattern))
```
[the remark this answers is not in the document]
aslakknutsen: Already documented as "starts with" and "prefixes".

The pattern follows the Compression filter and we should probably rather add an exclude_content_type option of required. https://github.com/praxis-proxy/praxis/issues/788

crates/filter/src/builtins/http/payload_processing/json_body/tests.rs
```diff
@@ -0,0 +1,1284 @@
...
+        r#"
+        request_remove:
+          - /a
+        "#,
+    );
+    assert!(
+        !filter.needs_request_context(),
+        "empty content_types should not need request context"
+    );
+}
```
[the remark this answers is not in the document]
aslakknutsen: fixed https://github.com/praxis-proxy/praxis/pull/1110/changes/22783c999cd163a2002831074730b78f037cb328

crates/simd-scan/src/x86_64.rs
```diff
@@ -0,0 +1,154 @@
...
+
+//! SSE2-accelerated scanner for JSON string delimiters (`x86_64`).
+//!
+//! This module is compiled only when `cfg(target_feature = "sse2")` holds for
+//! the whole crate (see the gate in `lib.rs`). SSE2 is in the default feature
+//! set of every `x86_64` target, so the gate is normally satisfied; a build
+//! that disables SSE2 falls back to the scalar scanner instead of relying on
+//! an ISA assumption.
+
+use core::arch::x86_64::{
```
shaneutt, line 12: Please create a issue as a reminder for us to do a check-up on this situation before `v1.0.0`, evaluate it again and make sure we remain satisfied with what we've got.

Cargo.toml
```diff
@@ -73,6 +75,7 @@ praxis-core = { version = "0.5.5", path = "crates/core", package = "praxis-proxy
 praxis-filter = { version = "0.5.5", path = "crates/filter", package = "praxis-proxy-filter" }
 praxis-protocol = { version = "0.5.5", path = "crates/protocol", package = "praxis-proxy-protocol" }
 praxis-tls = { version = "0.5.5", path = "crates/tls", package = "praxis-proxy-tls" }
+praxis-simd-scan = { version = "0.5.5", path = "crates/simd-scan" }
```
shaneutt: Everything in `crates/` is intended to be a published crate. If there's no strong reason this can't be a module closer to its callers let's please do that and make sure it's not exposed as a public library in any way.

shaneutt (dismissed):
Thanks for all the hard work @aslakknutsen 👍 

I'm not _thrilled_ about adding unsafe, but it's done in a limited way and we have a build flag so that's good.

Only a couple comments. Once resolved, 🚢

Cargo.toml
aslakknutsen: We can make it a module in praxis-proxy-filter if we update the unsafe_code directive to level deny so the unsafe parts can use expect(unsafe_code). That opens up the whole Filter crate to unsafe code and a possible larger blast radius then the current code. 

The reason for the crate separation is the top level unsafe_code forbid level and an attempt to isolate it. 

We could maybe make an internal non published crate by moving it under filter/

alexsnaps:
> 2\. crates/simd-scan isn't in the Summary at all. It's a new published crate
>         (publish = true), 594 lines of unsafe SIMD, plus a 180-line hand-copy of the
>         workspace lint table because the workspace forbids unsafe_code. That
>         duplicated block will drift silently from Cargo.toml — nothing checks the two
>         stay in sync. A new published crate with unsafe code deserves its own section
>         in the PR body, and arguably its own PR.

I somewhat concur here. Which is why I did a separate review on that code. It resulted [in a PR](https://github.com/aslakknutsen/praxis/pull/5) that @aslakknutsen merged on his branch. The code was clean as far as I could tell. Ran Miri on it and that confirmed if only my reading of the no need for alignment (tests added) & the haystack length. Ran Google's [Unsafe Rust Review Skill](https://github.com/google/rust-skills) on Fable 5.1.

wrt the lint rules and unsafe, I think it'd be fine to make an exception for this tho. I mean, it's probably fair to keep `unsafe` out by default, but... isn't there always an exception to any rule?! 😜

Cargo.toml
alexsnaps: I'd 👍 that

crates/simd-scan/src/x86_64.rs
alexsnaps: An issue for what?

crates/filter/src/json_ops/ops.rs
aslakknutsen: https://github.com/praxis-proxy/praxis/commit/f4037ccb02cf4c4ec76b2b7578d8ec158fd1ebd0

crates/filter/src/builtins/http/payload_processing/json_body/mod.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// Returns `true` when `content_type` matches the configured allowlist.
+    ///
+    /// An empty allowlist matches everything. Matching uses case-insensitive
+    /// prefix comparison, following the same convention as the compression
+    /// filter's `content_types` option.
+    fn matches_content_type(&self, content_type: &str) -> bool {
+        if self.content_types.is_empty() {
+            return true;
+        }
+        self.content_types.iter().any(|pattern| {
```
alexsnaps, line 245: Do we need to check for `Content-Encoding: gzip` ?

aslakknutsen: Wouldn't that be another filter in front? We have some "Compression" story.

alexsnaps: We discussed that offline, but tl;dr:

today, that filter would just follow the `on_invalid` path for gzip'ed JSON. 
Probably a user error. There is a wider discussion around these things tho: how does praxis want to "fail/operate" in those cases? Take the user by the hand (and e.g. automagically decompress?); give a warning at bootstrap time ("Hey! I see you're hoping for this to work fine... it might not"?); `on_invalid` err out "nicely" (this might have been valid, but was compressed and you didn't explicitly ask for decompression);... just do nothing?

cc/ @shaneutt @leseb

crates/simd-scan/src/x86_64.rs
shaneutt: Literally let's revisit this and decide if we still want and need this. Before `v1.0.0` where things lock in harder.

Cargo.toml
aslakknutsen: moved it to filter/simd-scan as an internal crate: https://github.com/praxis-proxy/praxis/pull/1110/commits/3b30bd30296fb7fa5ec2ca210732a6577ffa2c50

aslakknutsen: No that didn't work at all.  If we want to keep unsafe=forbid we need a crate, if we need a crate we need to publish it seems...

We can move it to a module under filter/json_ops, but then we need to make the unsafe forbid a deny instead and rewrite the project rules/conventions ever so slightly.

alexsnaps (approved):
Modulo the snafu with the publish crate, I'd approve this. 
This had my 👍 from the `simd-scan` "crate" perspective already, but it does meet the standard in place as far as I'm concerned. 

I'd like us to make a call on the `Content-Encoding: gzip` stuff, but that's somewhat out of scope for this I think (maybe create an issue to follow up on?)

Good stuff @aslakknutsen 🙌

Cargo.toml
aslakknutsen: inlined under filter /json_ops and lifted unsafe forbid; https://github.com/praxis-proxy/praxis/pull/1110/commits/65901ff2aab4f3aca8caa1119fd8123dbfc57f07

PR #1119, opened 2026-09-09
fix(protocol): reject private DNS-resolved HTTP upstreams

### LINKED ISSUE #1111: HTTP upstream/sub-request DNS resolution has no runtime private-IP (SSRF/rebinding) check
> Broken out from #1097. Independently verified against commit `a9bb26e7` (adversarial re-read of the source); still present on current `main`.

### 1. HTTP upstream/sub-request DNS resolution has no runtime private-IP (SSRF/rebinding) check

**Severity:** 🔴 High

**Location:** `protocol/src/http/pingora/handler/upstream_peer.rs:195-219 (build_peer, resolve_address call at 196)`, `filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs:1156-1184 (build_peer, resolve_address call at 1162)`, `core/src/connectivity/peer.rs:121-152 (resolve_address) and 155-169 (resolve_uncached) — no post-resolution filtering`, `core/src/connectivity/network.rs:87-90 (is_private_ip IPv6 branch omits the unspecified address ::)`

**Summary.** The documented runtime SSRF/DNS-rebinding control (`is_private_ip` + `allow_private_upstreams`) is wired only into the TCP proxy (`protocol/src/tcp/proxy.rs:797-823`, `connect_upstream`). The primary HTTP upstream path (`upstream_peer.rs::build_peer`) and the iterative sub-request executor (`iterative_request_router::build_peer`) both call `peer_utils::resolve_address`, take the returned `SocketAddr`, and connect with zero private/reserved-IP filtering. `resolve_address` caches the resolved address for 60 s (`DNS_TTL_SECS`) with no filtering. A hostname that passes config-time validation by resolving publicly can rebind after the TTL to loopback, RFC 1918, link-local, CGNAT, IPv6 ULA, or `169.254.169.254`, and the proxy will connect. Secondarily, `is_private_ip`'s IPv6 branch (network.rs:87-90) checks only loopback, `fe80::/10`, and `fc00::/7`, so the unspecified address `::` returns false/allowed even though config-time `is_ssrf_sensitive` rejects it via `is_unspecified()`.

**Impact.** Server-side request forgery and DNS rebinding against the HTTP data plane: after a configured public hostname rebinds, requests reach cloud metadata (169.254.169.254 -> credential theft), loopback, and internal-only services. The `allow_private_upstreams` flag defaults to false and is documented as preventing exactly this ("DNS-resolved upstream addresses ... are rejected to prevent DNS rebinding and SSRF attacks"), and startup_checks.rs warns it disables "runtime SSRF protection ... for upstream connections" — but that protection is silently absent on the HTTP path, giving operators a false sense of security.

**Trigger / reachability.** HTTP (not TCP) listener with an upstream/cluster whose address is a DNS hostname the attacker controls or can poison, with insecure_options.allow_private_upstreams at its secure default (false). Also reachable via iterative_request_router sub-requests.

**Reproduction.**

1. Configure an HTTP cluster upstream using a controlled DNS name that resolves to a public address so validation passes.
2. Start praxis with default insecure_options (allow_private_upstreams=false).
3. Change the DNS record to 127.0.0.1, 169.254.169.254, or :: .
4. Wait >60s (DNS_TTL_SECS=60) for the positive cache entry to expire.
5. Send another request; build_peer -> resolve_address returns and connects to the rebound private/metadata address with no rejection.

**Suggested fix.** Mirror the TCP path's post-resolution check on the HTTP path, gated on `allow_private_upstreams`. Add a checked helper in core/src/connectivity/peer.rs, e.g. `resolve_address_checked(address: &str, allow_private: bool)` that calls `resolve_address`, and if `!allow_private && is_private_ip(&addr.ip())` returns a new `AddressResolutionError::PrivateAddress { address, ip }` (reuse `praxis_core::connectivity::is_private_ip`). Have both HTTP `build_peer` functions call the checked variant: `upstream_peer.rs:196` and `iterative_request_router/mod.rs:1162`. Thread the `allow_private_upstreams` bool to both sites — it is already loaded into config (`InsecureOptions`); in the protocol handler it can be carried on `PingoraRequestCtx`/pinned pipeline (same source the TCP proxy reads at `tcp/mod.rs:92`), and passed into the sub-request executor's runtime. Applying the check at build_peer (rather than inside resolve_address) validates both cache hits and fresh resolutions. Separately, fix the secondary bug in network.rs:87-90 by adding `v6.is_unspecified()` to the IPv6 branch so `::` is treated as private, matching the config-time `is_ssrf_sensitive` behavior (the IPv4 unspecified 0.0.0.0 is already caught via `octets()[0]==0`). Add tests for the new error path and for `is_private_ip(::) == true`.

_Fix touches:_ `core/src/connectivity/peer.rs (add resolve_address_checked + AddressResolutionError::PrivateAddress variant)`, `protocol/src/http/pingora/handler/upstream_peer.rs (build_peer: call checked resolver; thread allow_private_upstreams via PingoraRequestCtx)`, `filter/src/builtins/http/traffic_management/iterative_request_router/mod.rs (build_peer: call checked resolver; thread flag through sub-request runtime)`, `core/src/connectivity/network.rs (is_private_ip IPv6 branch: add v6.is_unspecified())`

**Regression test.** Integration test on an HTTP listener: point an upstream at a hostname resolving to a private IP (e.g. via a test resolver hook or the localhost hosts entry) with allow_private_upstreams=false; assert build_peer/resolve_address_checked returns AddressResolutionError::PrivateAddress and the request fails (502), while the same config with allow_private_upstreams=true succeeds. Plus a unit test asserting is_private_ip(&"::".parse().unwrap()) == true.

**Notes.** The HTTP path's 60 s positive DNS cache means rebinding detection is delayed up to one TTL versus the TCP path (which deliberately uses uncached per-connection resolution); applying the check at build_peer validates the cached address on every request, so the only exposure window is the TTL after a rebind — acceptable and consistent with the existing cache design. Exploitation requires the operator to use a DNS hostname (not a literal IP) for the upstream and an attacker able to control/poison that record; literal-IP upstreams are validated at config time and unaffected. The `::` sub-issue is lower severity on its own but should be fixed in the same change for consistency with config-time validation.

shaneutt:
fixes #1111


leseb (approved):
Just nit, new tests contain inline comments, prohibited by docs/developing/conventions.md. Affected locations include network.rs:506, peer.rs:457,468, filtered_subrequest/tests.rs:487, upstream_peer.rs:440,462, and hostname_upstream.rs:19. Move explanations into assertion messages or tracing calls.

Otherwise, looks good!

PR #1124, opened 2026-09-09
fix: low-severity correctness and hardening cleanup batch

### LINKED ISSUE #1117: Low-severity cleanup batch: 47 confirmed correctness/robustness/hardening defects (L1-L47)
> Broken out from #1097. All 47 low-severity findings independently verified against commit `a9bb26e7`.
> **Note:** **L6 is already fixed on current `main`** by commit `c19338f6` ("fix(tcp): report the real close reason on the duration histogram"); the other 46 remain present. L15/L39 are doc-vs-code mismatches; L17/L22/L46 are behaviors the source explicitly documents as intentional trade-offs.

## Low severity (47)

Correctness, robustness, and hardening defects confirmed real but low-impact (edge-case inputs, non-security misbehavior, or operator-misconfiguration paths). Each is real and worth batching into cleanup; none is individually urgent.

### server

**L1. Config discovery and config loading use separate filesystem snapshots**  
`server/src/main.rs:76-82, server/src/server.rs:64-80, server/src/server.rs:254-260`  
The implicit praxis. yaml source is detected separately from loading it.

**L2. Every config reload cancels and respawns all health-check probes, resetting probe timers regardless of what changed**  
`server/src/reload.rs:180-200, server/src/reload.rs:152-158`  
`reload_pipelines` unconditionally calls `respawn_health_checks` on every successful reload, which cancels the old cancellation token and spawns a brand-new probe generation even when the cluster/health-check configuration is byte-identical (e. g.

**L3. Referenced-document watches remain pinned to the startup pipeline generation**  
`server/src/server.rs:260-269, server/src/watcher.rs:74-86, server/src/watcher.rs:196-209, server/src/watcher.rs:244-257`  
A successful main-config reload can change which external policy or document files the active pipelines consume, but the watcher never recomputes its path filter, watched directories, or composite-hash inputs.  Edits to newly referenced files therefore cannot trigger or pass the reload gate.

### protocol

**L4. HTTP health probes accept non-HTTP status lines when the second token matches**  
`protocol/src/http/pingora/health/probe.rs:98-100, protocol/src/http/pingora/health/probe.rs:133-149`  
A TCP service can be marked healthy without returning an HTTP response.  For example, `SMTP 200 ready` or `garbage 200 anything` parses as status 200.

**L5. Released StreamBuffer request bodies skip the global request-body ceiling (asymmetric with response path)**  
`protocol/src/http/pingora/handler/request_body_filter.rs, protocol/src/http/pingora/handler/response_body_filter.rs`  
In `request_body_filter::execute`, the match arm reached once a StreamBuffer request body has been Released (`request_body_released == true`) is an empty no-op: `BodyMode::StreamBuffer { ..  } => {}`.

**L6. TCP duration metrics label every completed forwarding path as `completed` regardless of the actual close reason**  
`protocol/src/tcp/proxy.rs:429-464, protocol/src/tcp/proxy.rs:692-716, protocol/src/tcp/metrics.rs:33-50`  
Sessions terminated by an I/O error, shutdown, idle timeout, or maximum-duration timeout are all recorded in `praxis_tcp_connection_duration_seconds` with `reason="completed"`.  This makes the histogram contradict the structured connection-close log and prevents operators from distinguishing normal closes from forced or failed sessions in Prometheus.

**L7. The HTTP/2 health probe rejects valid handshakes split across reads or kept open after SETTINGS**  
`protocol/src/http/pingora/health/probe.rs:219-238, protocol/src/http/pingora/health/probe.rs:241-247, protocol/src/http/pingora/health/probe.rs:174-203`  
TCP does not preserve write or HTTP/2 frame boundaries, so a valid nine-byte SETTINGS header may arrive in multiple reads.  The first short read is currently treated as a protocol failure.

### filter / engine + load-balancing

**L8. Body-phase filters returning TerminalResponse/StreamingTerminalResponse are silently discarded**  
`filter/src/pipeline/http_utils.rs:137-140`  
FilterAction exposes TerminalResponse and StreamingTerminalResponse to every filter, and in the header phase they short-circuit the pipeline with a synthesized response.  In the body phase, dispatch_body_result folds both variants into BodyFilterOutcome::Continue, so a body filter (on_request_body/on_response_body) that builds a terminal response has it thrown away with no warning logged, while Reject/BodyDone/Release are all honored.

**L9. ConsistentHash ring has no size cap, unlike RingHash — reload-time memory blowup**  
`filter/src/load_balancing/consistent_hash.rs:33-38, core/src/config/validate/cluster/load_balancer.rs:16-23`  
The consistent-hash ring is a Vec<usize> of length Σ(endpoint weight), rebuilt from scratch on every config reload.  Config validation caps MAX_ENDPOINTS = 10_000 and MAX_ENDPOINT_WEIGHT = 1_000 but applies the MAX_RING_ENTRIES = 1_048_576 ceiling ONLY to the RingHash strategy.

**L10. Content-Type parameter matching uses raw byte-equality instead of media-type normalization**  
`filter/src/condition/response.rs:110-134`  
When the expected Content-Type carries parameters, matching requires params(actual) == params(expected), an exact byte compare of everything after the first ';'.  The media type itself is compared case-insensitively, but the parameter segment is not normalized, so 'application/json; charset=utf-8' fails to match expected 'application/json; charset=UTF-8' (value case), 'application/json;charset=utf-8' (whitespace), or reordered parameters.

**L11. External configuration used only inside branch chains is never watched**  
`filter/src/pipeline/mod.rs:421-441, filter/src/pipeline/tests.rs:2756-2811`  
Dynamic reload relies on referenced_files to discover and hash external documents.  The method only visits top-level filters, even though resolved branch filters are active members of the same pipeline.

**L12. FilterResultSet enforces key/value length caps but not an entry-count cap**  
`filter/src/results.rs:136-147`  
FilterResultSet::set enforces MAX_KEY_LEN (64) and MAX_VALUE_LEN (256) but never bounds the number of distinct keys, unlike the sibling metadata maps in context. rs which are capped (MAX_METADATA_ENTRIES = 128).

**L13. Header conditions only match the first value of a multi-valued header**  
`filter/src/condition/request.rs:21-23, filter/src/condition/request.rs:131-134`  
HeaderMap::get returns only the first value when a header occurs multiple times.  Request (and response, condition/response.

**L14. P2C pick_two can choose the same endpoint twice for endpoints with weight >= 2**  
`filter/src/load_balancing/p2c.rs:156-171`  
pick_two guarantees two distinct weight slots (a ! = b), but two distinct slots map to the same endpoint position when that endpoint's weight >= 2.

**L15. Priority failover capacity ignores endpoint weights**  
`filter/src/load_balancing/priority.rs:23-33, filter/src/load_balancing/priority.rs:58-64, filter/src/load_balancing/priority.rs:111-140, filter/src/load_balancing/priority.rs:310-318`  
The implementation promises weight-based healthy capacity but computes a simple endpoint-count ratio.  For a primary tier containing weight-9 and weight-1 endpoints, losing only the weight-1 endpoint leaves 90% of capacity and should remain above the default approximately 71% threshold.

**L16. Status-code constructors panic on invalid input and disagree on the valid range**  
`filter/src/actions.rs:240-244, filter/src/actions.rs:71-75`  
Rejection::status accepts 100.. =599 (and 1xx is explicitly tested), while StreamingTerminalResponse::new and TerminalResponse::new require 200..

**L17. Subset/ZoneAware release() decrements in-flight counters on both inner strategies**  
`filter/src/load_balancing/subset.rs:117-122, filter/src/load_balancing/zone_aware.rs:101-106`  
Subset and ZoneAware hold two inner Strategy instances, each with its own in-flight counters when the inner strategy is least_connections or p2c.  select() increments only the strategy that served, but release() decrements BOTH.

### filter / security

**L18. CSRF Referer origin extraction leaks URL fragment when the URL has no path**  
`filter/src/builtins/http/security/csrf/origin.rs, filter/src/builtins/http/security/csrf/tests.rs`  
extract_origin_from_url splits the host portion only on '/', so a Referer like 'https://example. com#section' (fragment, no path) yields the malformed origin 'https://example.

**L19. Duplicate credential-injection cluster rules silently replace earlier secrets**  
`filter/src/builtins/http/security/credential_injection/config.rs:44-68, filter/src/builtins/http/security/credential_injection/filter.rs:137-142`  
When two rules use the same cluster name, HashMap::insert silently discards the first resolved credential and keeps the last.  The proxy starts successfully but may send a different secret or even a differently named credential header than the operator intended.

**L20. forwarded_headers: trusted-proxy append only reads the first X-Forwarded-For / Forwarded header line, dropping additional lines**  
`filter/src/builtins/http/security/forwarded_headers.rs, filter/src/builtins/http/security/forwarded_headers.rs, filter/src/builtins/http/security/forwarded_headers.rs, filter/src/builtins/http/security/forwarded_headers.rs`  
HeaderMap::get returns only the first value for a header name.  When a trusted upstream sends X-Forwarded-For (or Forwarded) as multiple separate header lines — which is valid HTTP and semantically equivalent to a single comma-joined line — the filter reads only the first line, appends the client IP, and pushes the result as an overwrite.

**L21. Host policy factories execute while the global registry read lock is held**  
`filter/src/builtins/http/security/policy/host_plugins.rs:68-78, filter/src/builtins/http/security/policy/host_plugins.rs:61-65`  
The temporary RwLockReadGuard produced by read() remains alive through collect(), so each arbitrary host-supplied make closure runs under the registry read lock.  A factory that registers another factory, directly or through initialization code, waits forever for the write lock held back by its own read lock.

**L22. Post-response policy denials can emit truncated, invalid JSON-RPC envelopes**  
`filter/src/builtins/http/security/policy/filter.rs:1055-1082, filter/src/builtins/http/security/policy/filter.rs:1134-1175, filter/src/builtins/http/security/policy/mod.rs:106-110, filter/src/builtins/http/security/policy/tests.rs:1875-1888`  
A normal short upstream result can be smaller than the generated denial envelope.  In that case fit_to_original_length slices the envelope at an arbitrary byte offset, so the client receives malformed JSON rather than the documented structured policy denial.

### filter / traffic management

**L23. Expired lookup can delete a concurrently refreshed session binding** *(independently found by both reviewers)*  
`filter/src/builtins/http/traffic_management/sticky_sessions/store.rs:106-116, filter/src/builtins/http/traffic_management/sticky_sessions/store.rs:144-149`  
When `get` observes an expired entry, it drops the DashMap guard and then removes the key in a separate operation.  During that gap, another request can acquire the key, refresh `last_accessed`, and update its endpoint through `put`; the original `get` then removes that fresh binding.

**L24. Sticky-session entry typos are silently accepted** *(independently found by both reviewers)*  
`filter/src/builtins/http/traffic_management/sticky_sessions/config.rs:20-40, filter/src/builtins/http/traffic_management/sticky_sessions/config.rs:181-206, filter/src/builtins/http/traffic_management/sticky_sessions/config.rs:251-256`  
The top-level configuration denies unknown fields, but each flattened cluster entry and its tagged persistence variants do not.  Unknown keys inside a cluster entry are therefore ignored rather than rejected.

**L25. TLS Host-derived SNI is discarded when selecting an alternate retry endpoint**  
`filter/src/builtins/http/traffic_management/load_balancer/entry.rs:78-95, filter/src/builtins/http/traffic_management/load_balancer/entry.rs:125-139, filter/src/builtins/http/traffic_management/load_balancer/reselector.rs:75-83`  
Initial selection clones the TLS config and fills missing SNI from the request Host header.  The retry reselector instead captures the original cluster TLS object before that per-request fallback and later builds alternate upstreams directly from it.

**L26. Zero half-open timeout is accepted and destabilizes circuit recovery**  
`filter/src/builtins/http/traffic_management/circuit_breaker/config.rs:73-78, filter/src/builtins/http/traffic_management/circuit_breaker/mod.rs:188-212`  
The factory rejects zero failure thresholds and recovery windows but passes `half_open_timeout_secs = 0` directly to the circuit state machine.  A zero duration makes a newly issued half-open probe immediately stale to another request, allowing concurrent traffic to reset the circuit and invalidate the probe before its result can close the breaker.

### filter / transformation + payload + observability

**L27. Access-log sample_rate is rounded to coarse reciprocal rates**  
`filter/src/builtins/http/observability/access_log.rs:234-237, filter/src/builtins/http/observability/access_log.rs:277-282, filter/src/builtins/http/observability/access_log.rs:300-306, filter/src/builtins/http/observability/access_log.rs:1069-1087`  
Most accepted rates cannot be represented by one-in-N sampling.  For example, sample_rate:0.

**L28. Configurable promoted-header names are not validated at configuration time** *(independently found by both reviewers)*  
`filter/src/builtins/http/payload_processing/json_body_field/config.rs:62-70, filter/src/builtins/http/payload_processing/json_body_field/mod.rs:196-204, filter/src/builtins/http/observability/request_id.rs:85-90, filter/src/builtins/http/observability/request_id.rs:154-176`  
Both factories accept configurations such as header:"bad header" or header_name:"bad header" even though these values cannot identify an HTTP header.  The invalid mutation will subsequently fail or be omitted at the protocol boundary, making an apparently valid deployment silently lose its promoted metadata or request IDs.

**L29. json_body_field promotes metadata from a complete prefix before end-of-stream**  
`filter/src/builtins/http/payload_processing/json_body_field/mod.rs:183-204, filter/src/builtins/http/payload_processing/json_body_field/tests.rs:194-235`  
A client can make the first delivered chunk a complete JSON document such as {"model":"premium"} and then send additional bytes.  The first invocation has end_of_stream=false but still promotes the field and returns BodyDone, permanently skipping this filter for the rest of the body.

**L30. Multiple request_add entries for the same header discard earlier additions**  
`filter/src/builtins/http/transformation/header/mod.rs:194-209, filter/src/context.rs:306-313, filter/src/builtins/http/transformation/header/tests.rs:483-538`  
For an existing header H:original and two configured request_add entries H:a and H:b, the filter queues two set operations, original,a and original,b.  Because set operations overwrite, the final value loses a.

### core / config parsing + validation

**L31. Config validation recurses into iterative_request_router step filters without a depth bound**  
`core/src/config/validate/inline_clusters.rs:73-78, core/src/config/validate/filter_chain.rs:76-82`  
Branch-chain nesting is explicitly depth-capped (MAX_BRANCH_DEPTH) in validate_branch_chains, but the recursion through `iterative_request_router` step filters in validate_entry (inline_clusters) and validate_entry_conditions (filter_chain) has no depth parameter or ceiling.  A config whose step filters nest another iterative_request_router, whose step filters nest another, etc.

**L32. Endpoint unknown-field rejection message omits valid fields (metadata, priority, zone)**  
`core/src/config/cluster/endpoint.rs:115-124`  
WeightedEndpointRaw accepts `address`, `weight`, `metadata`, `priority`, and `zone`, but the error emitted for an unknown/typo'd key tells the operator that only 'address' and 'weight' are expected.  An operator who typos `metadataa:` or `zonee:` will be told the real fields metadata/priority/zone are not valid, which is misleading and could push them to delete a legitimately intended field.

**L33. Endpoint validation accepts strings that are not valid host:port addresses**  
`core/src/config/validate/cluster/endpoints.rs:55-87`  
After SocketAddr parsing fails, validation merely splits on the final colon, parses the suffix as u16, and checks that the remaining host text is nonempty.  It does not validate hostname syntax or balanced IPv6 brackets.

**L34. Health-check hostname validation is vulnerable to DNS rebinding SSRF**  
`core/src/config/validate/cluster/health_check.rs:158-177, core/src/config/validate/cluster/health_check.rs:235-244`  
For IP literals, validation checks the actual address, but every ordinary hostname is checked only against a small lexical blacklist such as localhost, *. internal, *.

**L35. Invalid response-condition status codes and header names silently create non-matching predicates**  
`core/src/config/condition/response.rs:64-73, core/src/config/validate/filter_chain.rs:140-170`  
Response conditions deserialize status values as unrestricted u16 values and header names as arbitrary strings.  Validation checks only that lists and maps are nonempty.

**L36. Logging buffer_size accepts values large enough to exhaust process memory**  
`core/src/config/logging.rs:28-40, core/src/config/logging.rs:58-90, core/src/config/logging.rs:93-97`  
buffer_size is an unconstrained u32.  Validation rejects only zero, and effective_buffer_size_lines converts every other value directly to usize.

**L37. Terminal filter placement is validated per chain instead of per listener's flattened pipeline**  
`core/src/config/validate/filter_chain.rs:27-33, core/src/config/validate/filter_chain.rs:181-198, core/src/config/mod.rs:132-174`  
TERMINAL_FILTERS is documented as requiring the last position in both its chain and the flattened listener pipeline, but validate_terminal_filters examines each chain independently and ignores listener chain ordering.  A listener can reference one chain ending in iterative_request_router followed by another nonempty chain; Config::from_yaml returns a validated Config, but later pipeline construction rejects it.

**L38. The YAML alias pre-scan misses valid aliases whose names begin with a hyphen**  
`core/src/config/parse.rs:150-171, core/src/config/parse.rs:179-209`  
The safety contract says every YAML alias node is rejected before deserialization, but line_contains_alias recognizes an alias only when the byte after '*' is alphanumeric or underscore.  The YAML parser also permits '-' in anchor names, including as the first character, so an alias such as '*-node' passes the pre-scan and reaches serde_yaml.

### core / runtime + health + kv + retry

**L39. CircuitBreakerRegistry::evict_idle doc comment contradicts the actual (intentional) eviction predicate**  
`core/src/circuit.rs:475-478, core/src/circuit.rs:337-344, core/src/circuit.rs:364-371, core/src/circuit.rs:331-336`  
The `evict_idle` method-level doc states breakers are evicted when they "have been `Closed` with zero failures", but the underlying `is_idle` implementation and its own detailed doc explicitly reject that criterion (it caused unbounded registry growth) and instead evict on idleness + in_flight==0 while exempting only Open-in-recovery-window breakers.  A reader relying on the `evict_idle` doc would misunderstand which breakers survive eviction (e.

**L40. In-memory KV entry limit is bypassable and non-atomic**  
`core/src/kv/memory.rs:84-93, core/src/kv/memory.rs:141-154`  
The resource ceiling is not an invariant.  `from_pairs` can create a store of arbitrary size, and concurrent writers can all observe a length below `MAX_ENTRIES` before inserting distinct keys, pushing the store beyond the cap.

**L41. Malformed non-ASCII Connection values bypass nominated-header stripping**  
`core/src/subrequest/internals.rs:329-341, core/src/subrequest/internals.rs:343-360`  
HTTP header values may contain valid obs-text bytes that are not UTF-8.  If a `Connection` value contains such a byte after an otherwise valid token, `HeaderValue::to_str` fails for the entire value and all nominations in it disappear.

**L42. Retry tokens remain above the dynamic cap after traffic falls**  
`core/src/retry.rs:15-17, core/src/retry.rs:71-97, core/src/retry.rs:115-131, core/src/retry.rs:194-200`  
A high active-request count can fill the bucket to a large cap.  If traffic then drops, the lower cap is not applied until a later call actually adds refill tokens.

### tls / config + SNI + cert loading

**L43. Cached upstream client identities are accepted without validating certificate structure or key matching**  
`tls/src/cached.rs:243-247, tls/src/cached.rs:301-312, tls/src/cached.rs:356-367, tls/src/cached.rs:370-382, tls/src/setup/loader.rs:29-47`  
The cached cluster path only base64-decodes certificate PEM blocks and independently parses a private key.  It neither validates that the certificate bytes are a valid X.

**L44. Dropping the watcher shutdown sender leaves a permanently ready select branch that spins**  
`tls/src/watcher.rs:103-106, tls/src/watcher.rs:165-193`  
When every watch::Sender is dropped, Receiver::changed() returns Err immediately and continues returning Err on every call.  The select branch handles only Ok(true), then loops and polls the already-closed receiver again.

**L45. Private key DER from loader path is not wrapped in Zeroizing before/after parse**  
`tls/src/setup/loader.rs:95-104`  
`load_cert_and_key` reads the key PEM into a `Zeroizing<Vec<u8>>` (good), but the parsed `PrivateKeyDer<'static>` it returns owns a fresh, non-zeroizing copy of the private key bytes.  This value is threaded through `load_certified_key` and handed to the crypto provider; when the intermediate `PrivateKeyDer` is dropped, the secret DER is not guaranteed to be scrubbed.

**L46. Reloadable mTLS keeps advertising the initial CA after a successful CA rotation**  
`tls/src/reload.rs:194-197, tls/src/reload.rs:208-214, tls/src/reload.rs:227-230, tls/src/reload.rs:261-263`  
A CA reload swaps the verifier but never updates the certificate-authority subjects sent in CertificateRequest.  After rotating from CA A to disjoint CA B, the server verifies against B while continuing to tell clients that A is acceptable.

**L47. The SNI parser accepts duplicate or malformed extensions after the first SNI extension**  
`tls/src/sni.rs:439-458, tls/src/sni.rs:462-468, tls/src/sni.rs:472-506`  
parse_extensions returns as soon as it sees the first SNI extension, so it never validates subsequent extension framing or rejects a second SNI extension.  parse_sni_extension also accepts extra bytes beyond the declared ServerNameList and silently ignores a one- or two-byte truncated list tail.

shaneutt:
fixes #1117


leseb (changes requested):
[P2] #1124 can hash a different document version than the pipeline loaded.
A newly referenced document is consumed during pipeline construction and then re-read for the hash at [watcher.rs](https://github.com/praxis-proxy/praxis/blob/1e6312bf5dad9cc0b765579a2e63d694474b0182/crates/server/src/watcher.rs#L475-L478). An intervening edit can leave stale live content recorded as current and initially unwatched.

[P2] #1124 does not retry failed watcher re-registration.
[rewatch](https://github.com/praxis-proxy/praxis/blob/1e6312bf5dad9cc0b765579a2e63d694474b0182/crates/server/src/watcher.rs#L517-L520) only logs failure. State and hashes are already advanced, so subsequent reloads need not attempt registration again.

[P2] #1124 misses referenced-document symlink rotations.
The referenced-path matching at [watcher.rs](https://github.com/praxis-proxy/praxis/blob/1e6312bf5dad9cc0b765579a2e63d694474b0182/crates/server/src/watcher.rs#L577-L580) records lexical/current canonical paths but does not apply the main config’s symlink-aware behavior. Kubernetes-style ..data rotations can remain invisible.

[P2] #1124 still loses repeated additions when a request header is absent.
[request_add](https://github.com/praxis-proxy/praxis/blob/1e6312bf5dad9cc0b765579a2e63d694474b0182/crates/filter/src/builtins/http/transformation/header/mod.rs#L212-L214) queues separate additions, but the protocol applies them with replacement semantics. Only the last value survives, contradicting the new stacking documentation.

[P2] #1124 silently increases very small sampling rates.
[sample_rate](https://github.com/praxis-proxy/praxis/blob/1e6312bf5dad9cc0b765579a2e63d694474b0182/crates/filter/src/builtins/http/observability/access_log.rs#L291) accepts values below 1e-6 but clamps them to one sample per million. For example, 1e-9 produces 1000× the configured rate.

leseb:
Also, it goes without saying that post-v1 we will stop doing bulk fixes like that.

leseb (changes requested):
- Later credential/header replacements are appended instead of replacing earlier values.
- Shared documents can still have conflicting loaded versions hidden by digest deduplication.
- Repeated request_add values still collapse inside filtered subrequests.
- Rates such as 0.00000051 still round up to 0.000001.

PR #1127, opened 2026-09-09
feat(policy): back policy http calls with the proxy connector

### LINKED ISSUE #1098: feat(policy): use Praxis's Pingora client for policy HTTP requests
### Description

The policy engine uses an injected `HttpTransport` for JWT, OAuth, and CIBA requests. Praxis currently injects the engine's Hyper transport, which leaves us with two HTTP stacks: Pingora for proxy traffic and Hyper for identity-provider requests.

This means separate connection pools, TLS configuration, and egress rules. It also keeps `hyper-util`, `hyper-rustls`, `rustls`, and `webpki-roots` in the dependency graph. The comment in `crates/ppe/Cargo.toml` says Praxis supplies a Pingora transport, but that has not been implemented yet.

Implement `praxis_policy_core::http::HttpTransport` using Praxis's Pingora client. Install it with `PolicyEngine::set_http_transport` in `PolicyFilter::new`, before `initialize()` fetches JWKS.

### Important details

- Map errors conservatively. Callers use `may_have_reached_peer` to decide   whether retries are safe. Only failures that prove no request was sent should   be `Connect`; ambiguous failures should be `Io`. Retrying a token exchange or   CIBA dispatch can create a second credential or prompt a user twice.
- Enforce each request's response-size limit. Current limits include 256 KiB for   JWKS and 64 KiB for token responses.
- Apply the deadline to the full exchange, including the response body.
- Create the pool lazily. `PolicyFilter::new` runs `initialize()` on a temporary   runtime, so a pool created there may be tied to a reactor that is later   dropped.

### Egress policy

Choose one source of egress rules and document it. Either Praxis owns the rules, or the transport checks the shared, I/O-free `praxis_policy_core::http_addr` table before connecting. Reusing the shared table avoids maintaining another range list; make sure it covers `100.64/10` and IPv4-mapped IPv6 addresses such as `::ffff:169.254.169.254`.

If Praxis owns egress policy, replace the policy filter's `allow_private_idp` setting with the equivalent Praxis setting instead of keeping both.

### Acceptance criteria

- `PolicyFilter` installs the Pingora transport before initialization.
- The `http-hyper` feature is disabled, and `hyper-util`, `hyper-rustls`,   `rustls`, and `webpki-roots` are gone from `cargo tree`.
- Tests cover retry classification for connect failures, timeouts, and ambiguous   I/O failures.
- Tests cover response-size limits and deadlines that include body reads.
- Transport construction opens no socket; the pool is created on first use.
- A metadata-range destination is rejected unless private IdP access is enabled   through the chosen egress policy.
- The praxis-demos policy example passes all 12 Cedar, CEL, and OPA scenarios   against a local Keycloak instance.
 `praxis_policy_core::http_testing::FakeTransport` can cover cases that are hard to reproduce with a mock server, including connect failures, timeouts, and key rotation between fetches.

## Out of scope

Changing the `HttpTransport` trait or inspecting response bodies beyond what the existing callers require.

## References

- `crates/ppe-core/src/http.rs`, `http_retry.rs`, and `http_addr.rs`
- `crates/ppe/src/http_hyper.rs`
- `filter/src/builtins/http/security/policy/filter.rs::install_http_transport`

araujof:
**Description**

The policy engine's outbound calls (JWKS, RFC 8693 token exchange, CIBA backchannel) went through PPE's bundled hyper client, so a `policy-engine` build ran two HTTP stacks against the same IdP: two pools, two trust stores, and an egress path Praxis's own config never saw. They now go through `SubRequestClient` over the process's single `SubRequestConnector`.

The connector is registered in `resolve_pipelines` and read through a process-global holder, because the engine fetches JWKS while its filter is being constructed, before any pipeline exists to hand it a client. Server startup, `--validate`/`--dump`, and reload all go through that one seam.

Policy calls now inherit `runtime.subrequest_pool_size`, `runtime.subrequest_max_connections`, and `runtime.subrequest_circuit_breaker`. `body_limits.max_response_bytes` deliberately does not apply — the transport keeps its own 1 MiB ceiling, so a tight proxy-wide response limit can't fail a JWKS fetch. The breaker is keyed per address and SNI, so cross-talk needs a data-plane failure against the very same address, and an open circuit is refused before anything is sent. Sharing `subrequest_max_connections` cuts both ways, so size it for the identity path too.

Closes: #1098 

**Changes**

- Policy calls are HTTP/1.1 only. `PeerOptions::new()` sets `alpn: H1` and nothing in the sub-request path offers h2; hyper offered it by default.
- The egress check runs on the resolved address at dial time rather than on the URL, using PPE's shared address table so the two paths can't drift. Only the address actually dialled is judged, so there's no rebinding window and no second lookup.
- An https URL naming an IP literal is refused: `derive_sni` yields an empty SNI, which leaves hostname verification with nothing to check. hyper would have attempted it.
- A request with no connect bound gets PPE's 2s default. Without it the overall deadline fires first and an unreachable IdP reports a timeout, which a delegating caller has to treat as a possibly-minted token instead of a clean retry.
- Name resolution is charged against the request deadline. `SubRequestClient`'s clock only starts once a peer exists, so an unbounded lookup would be covered by nothing; `HyperTransport` wrapped it because its resolver sat inside the connector.
- Admission exhaustion reports a connect failure rather than a refusal. PPE never retries a refusal at any idempotency level, so a JWKS fetch would have silently lost the retries it asks for. An open circuit stays a refusal.
- Policy peers carry a dedicated group key, so their connections are never interchanged with data-plane ones. A peer's reuse hash ignores `options.ca`, so a connection established under a cluster's private CA could otherwise serve a policy call that never trusted it.
- A port outside the dialable range is refused instead of quietly falling back to the scheme default, and a rejected URL's userinfo is kept out of the error text.

**Notes** 

- `allow_private_idp` is unchanged and there is no new config surface, so no new example config.
- Dropping `http-hyper` removes `hyper-rustls`, `webpki-roots`, `http-body-util` and `tower-service` from the graph entirely, and `hyper`/`hyper-util` unless `otel` is on (they come back via `opentelemetry-otlp → tonic → hyper-timeout`). `rustls` stays regardless — `praxis-proxy-tls`, Pingora and the valkey session store all pull it — so the original acceptance criterion asking for it to disappear isn't achievable.
- Known gap: a private-CA IdP still has no config surface. Verification uses the platform trust store (honouring `SSL_CERT_FILE`/`SSL_CERT_DIR`), and only cluster TLS config can attach a custom CA, which a policy URL doesn't have. `HyperTransport` had the same limitation (webpki roots only). Worth a follow-up for an optional CA / client-cert reference on the policy filter.
- The diff is over the 500-line gate, and most of it is the transport's in-crate unit tests. The size check only excludes paths starting with `tests/`, while `coverage-check` excludes `tests/` anywhere and treats the same file as tests.


leseb (changes requested):
P1 — DNS failover regression: transport.rs:74 collapses all DNS answers into one address cached for 60 seconds. A denied or unreachable preferred address prevents using another healthy IdP address; retries hit the same cached address.

P2 — connector leaks across in-process runtimes: shared_connector.rs:43 permanently retains the first connector. A second proxy runtime uses the first runtime’s pool, admission limit, and breaker rather than its own. Normal reload is unaffected, but public embedders and integration processes are.

P2 — valid HTTPS IP endpoints are rejected: transport.rs:228 rejects IP literals even with allow_private_idp. Rustls can verify an IP SAN without emitting DNS SNI, and the previous Hyper transport supported this configuration.

leseb (approved):

PR #1154, opened 2026-09-11
feat(server): add server composition API (#1085)

### LINKED ISSUE #1085: feat(server): expose embeddable server composition API
## Summary

Expose a supported server-composition API for downstream Praxis
distributions without exposing or duplicating the server lifecycle.

## Scope

- Accept a downstream filter registry, pipeline-scoped extensions,
  pipeline activators, configuration-source defaults, application
  metadata, CLI integration, and branding.
- Keep extension factories synchronous and side-effect-free.
- Ensure explicit operator configuration and CLI values override
  distribution defaults.
- Keep parsed operator configuration immutable from the composition
  boundary.
- Preserve `run_server` and existing custom-registry entry points as
  compatible wrappers.
- Keep plugin registration and lifecycle outside this API.

## Acceptance Criteria

- [ ] Downstream distributions can describe all supported composition
      inputs through one stable entry point.
- [ ] The API does not expose watcher, listener, reload, or publication
      internals.
- [ ] Factory contracts cannot require asynchronous initialization or
      activation I/O.
- [ ] Operator configuration and CLI precedence are covered by tests.
- [ ] Existing entry points retain their behavior and source
      compatibility.
- [ ] API documentation includes a minimal downstream example and
      ownership rules.

## Related

- Parent epic: #1076
- Proposal: https://github.com/praxis-proxy/enhancements/blob/main/proposals/00009_embeddable-server-composition.md
- External activation dependency: https://github.com/orgs/praxis-proxy/discussions/973

leseb:
## Summary

Introduces `ServerComposition`, a **data-only** description of how an embedding server customizes the downstream request path, plus `run_server_with_composition` as the single server-lifecycle owner. `run_server` and `run_server_with_registry` become thin wrappers over it, so existing entry points are unchanged.

Closes #1085.

## What a composition carries

Three hooks, each handed narrow, immutable, server-owned context:

- **Registry factory** — builds the `FilterRegistry` from a `RegistryContext` exposing only the server-owned `SubRequestClient`. The server, not the embedder, owns the sub-request client's lifecycle.
- **Pipeline-extension factories** — mint a *fresh* `Box<dyn PipelineExtension>` per candidate listener pipeline via `ExtensionContext` (config + listener + sub-request client). One extension instance never leaks across pipelines.
- **Read-only validators** — inspect the config, listener, flattened filter entries, and the built pipeline through `ValidatorContext` without mutating any of them.

## Lifecycle & reload

Extensions and validators run at startup **and on every hot reload** — the composition is carried into the reload watcher — so downstream customization survives config reloads rather than being silently dropped on the first swap.

## Validator correctness

Validators receive an **immutable snapshot of the flattened filter entries captured before `build_with_chains` runs**. `build_with_chains` drains `conditions`, `response_conditions`, and `branch_chains` out of the entries as it moves them into the built pipeline; taking the snapshot first means conditional filters and branch chains stay visible to downstream validation instead of appearing empty. A regression test asserts request conditions, response conditions, and branch chains are all still observable via the validator context.

## Testing

- `cargo test -p praxis-proxy` — all unit, binary, and doctests green
- `make lint` — clippy (workspace all-features + `-p praxis-proxy --no-default-features`), nightly `fmt --check`, machete, and xtask lints all pass
- `make doc` — clean with `-D warnings`

### DIFF
--- a/crates/server/src/composition.rs
+++ b/crates/server/src/composition.rs
@@ -1,0 +1,611 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 Praxis Contributors
+
+//! Data-only server composition for embedding Praxis.
+//!
+//! [`ServerComposition`] describes how a downstream binary customizes an
+//! embedded Praxis server without reaching into its lifecycle. It is passed
+//! to [`run_server_with_composition`] and carries three things:
+//!
+//! 1. **Registry construction** — how the filter [`FilterRegistry`] is built. The factory runs once at startup and
+//!    receives a [`RegistryContext`] exposing the server-owned [`SubRequestClient`], the one runtime handle a
+//!    downstream registry commonly needs while wiring its filters.
+//! 2. **Pipeline-extension factories** — each produces a fresh [`PipelineExtension`] per listener pipeline. Factories
+//!    receive an immutable [`ExtensionContext`] and must be synchronous and side-effect-free; they run again on every
+//!    hot reload.
+//! 3. **Read-only pipeline validators** — each inspects a built pipeline via a [`ValidatorContext`] and may reject it,
+//!    but cannot mutate it.
+//!
+//! The composition never exposes mutable pipelines, watchers, listeners, or
+//! publication handles: it is a description consumed by the server, not a hook
+//! into a running one.
+//!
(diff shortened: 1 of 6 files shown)


shaneutt (approved):
I'll be interested to see where this goes 👍

PR #1165, opened 2026-09-13
feat(filter): opt-in body-model authorization in the policy filter

### LINKED ISSUE #1164: policy filter should authorize on the request body model
**What would you like to be added**:

An opt-in that lets the `policy` filter authorize on the request body `model` field, so a deny rule can gate which models a caller may invoke.

The pure-L7 global policy authorizes at the header phase, over identity and inbound headers. The OpenAI-style top-level `model` lives in the request body and never reaches the policy bag, so no rule can gate the model a request is served.

Behind an `authorize_on_model` flag (default off), the filter buffers the body, defers the global authorization to end-of-stream, parses the top-level `model`, and populates the typed `Extensions.llm.model_id` slot the engine already exposes as the `llm.model_id` bag attribute. A deny rule keyed on `llm.model_id` then enforces the caller's allowed set.

**Why is this needed**:

Governance has to key on the model a request is actually served with, which is the body value. A client-supplied header is spoofable and can disagree with the body, so the proxy must set the attribute itself from the parsed body. This lets an operator write a plain membership deny in the existing policy language, reusing the policy engine, the filter's own body parser, and the tested `llm.model_id` bag-key extraction, with no change to the engine crate.

The value is server-set and lives in an internal typed attribute, not a client-settable header, so it cannot be spoofed and needs no reserved-prefix strip. A missing or malformed `model` leaves `llm.model_id` unset, so a membership deny fails closed.

Scope is the `model` field. The same mechanism extends to other body-derived attributes later.

hexfusion:
### What does this PR do?

Adds an opt-in `authorize_on_model` flag to the `policy` filter, so a deny rule can gate which models a caller may invoke. When set, the filter buffers the request body, defers the pure-L7 global authorization to end-of-stream, parses the top-level `model`, and populates the typed `Extensions.llm.model_id` slot the engine already exposes as the `llm.model_id` bag attribute. A deny rule keyed on `llm.model_id` enforces the caller's allowed set.

`llm.model_id` is an internal typed attribute the proxy sets from the parsed body, not a client-settable header, so a caller cannot spoof it and no reserved-prefix strip is needed. `StreamBuffer` holds the request until the decision, so a deny stops it before the backend is contacted. A missing or malformed `model` leaves `llm.model_id` unset, so a membership rule fails closed. Off by default, and the existing authorization path is unchanged.

The `llm.model_id` slot and its bag-key extraction already ship and are tested in the engine crate. This PR adds the producer that constructs that type from the filter, with no change to the engine crate.

### Why not existing filters?

A pure-L7 global policy authorizes at `on_request`, before the request body exists, so it cannot gate on a body-derived value. The model to authorize on lives in the request body (the top-level `model` field). Composing `json_body_field` (which promotes the body `model` into a request header) with a policy deny on that header does not work for a global policy: `json_body_field` sets the header at the body phase, which runs after the pure-L7 policy has already authorized and forwarded the request. There is no existing way to make a `global` HTTP policy gate on the body model. `authorize_on_model` is that missing piece: it defers the global authorization to end-of-stream so the parsed model is available, and sources it from the typed `llm.model_id` extension slot the proxy sets server-side rather than a spoofable header.

Fixes #1164

No. The flag defaults off and the existing authorization path is unchanged.


leseb (changes requested):
end-to-end coverage is missing

shaneutt (changes requested):
Hold for @araujof's review

araujof (commented):
Thanks for this PR, @hexfusion 

I [commented](https://github.com/praxis-proxy/praxis/issues/1164#issuecomment-5666667735) on the issue so that we can have a bit of a discussion there. 

I'd like to propose a slightly different direction that reuses what PPE already supports (LLM-based routes and policies). The issue would then be recast to complete the wiring that is currently missing in the Praxis policy filter side. It shouldn't be a lot of changes. I outlined them in my comment, and offered a few examples of how the existing PPE config surface could be used to express model authorization policies. 

cc: @leseb @shaneutt

hexfusion:
resolved via https://github.com/praxis-proxy/praxis/pull/1170

PR #1166, opened 2026-09-14
feat(connectivity): URL-aware target preparation for subrequests

### LINKED ISSUE #1092: feat(subrequest): add URL-aware target preparation
Parent: #1072

Parallel to #1073. Required by praxis-proxy/ai#879 before Praxis AI callouts migrate to filtered subrequests.

## Summary

`SubRequestClient` deliberately accepts a prepared `HttpPeer` plus an origin-form `SubRequest`. URL-based callers must therefore independently parse an absolute URL, resolve DNS, validate the resolved addresses, construct the peer and TLS SNI, rewrite the path and query, and bind `Host` to the URL authority.

That is a security-sensitive preparation boundary. Duplicating it across callers makes it possible for the resolved socket address, TLS SNI, request target, and HTTP authority to diverge.

Add a reusable URL-aware target-preparation API in `praxis-core`. Given an absolute HTTP(S) URL, an operation deadline, and a caller-supplied resolved-address validation hook, it prepares the transport inputs while treating the parsed URL authority as authoritative.

Keep the generic `SubRequestClient::execute(peer, request, ...)` contract unchanged. Its ability to preserve an explicit `Host` is required for callers that intentionally connect to an IP while using HTTP virtual hosting.

This is a core enhancement, not the immediate fix for praxis-proxy/ai#676. The existing AI behavior is fixed independently by praxis-proxy/ai#921; callers can migrate to the core API after it is available.

## Scope

- Parse an absolute `http` or `https` target and derive the host, effective port, TLS mode, SNI, original authority, and origin-form path and query.
- Reject malformed targets, unsupported schemes, fragments, and any userinfo, including empty userinfo forms.
- Resolve hostnames within the caller's absolute operation deadline.
- Normalize and deduplicate the complete resolved address set.
- Invoke a caller-supplied validation hook on the complete resolved address set before any address is dialed, so application-specific SSRF policy remains outside core.
- Produce a prepared target that supports connection fallback across validated addresses without re-resolving DNS.
- When materializing the outbound request, overwrite any caller-supplied `Host` with the parsed URL authority.
- Make the prepared target usable by both buffered and streaming subrequest paths and by the filtered-subrequest executor introduced in #1073.
- Expose typed errors that distinguish invalid targets, resolution failures, policy rejection, and deadline expiry.

## Acceptance Criteria

- The prepared request sends the URL authority as `Host` on the wire, including a non-default explicit port, even when the input request contains a conflicting `Host`.
- HTTPS peers use the URL hostname for SNI while connecting to a validated resolved address.
- IPv4 literals, bracketed IPv6 literals, default ports, explicit ports, empty paths, and path/query combinations are handled correctly.
- All resolved addresses are presented to the validation hook, and rejection prevents every connection attempt.
- DNS resolution and target preparation consume the caller's existing absolute deadline rather than starting a new timeout budget.
- A prepared target can try subsequent validated addresses after a connection failure without resolving the hostname again.
- Buffered and streaming callers share the same target-preparation behavior.
- Existing tests continue to prove that low-level `SubRequestClient::execute` preserves an explicit virtual-host `Host`.
- Boundary tests use controlled DNS results and a test validation hook to cover mixed public/private answers, rebinding resistance, authority binding, fallback, and deadline expiry.
- Public API documentation explains the authority invariant and which policy decisions remain caller-owned.

## Out of Scope

- Defining whether private, loopback, link-local, or otherwise non-public addresses are allowed.
- Configuring or executing outbound filter chains; that remains #1073 and #1074.
- MCP operation-wide pagination and handshake deadlines; that remains praxis-proxy/ai#836.
- Credential selection, projection, or disclosure policy.
- Changing the low-level `SubRequestClient::execute(peer, request, ...)` virtual-hosting behavior.

leseb:
### What does this PR do?

Implements issue #1092. Adds a pin-before-dial URL target preparation pipeline for sub-requests: it parses and shape-validates a URL target, resolves the host through the shared bounded DNS cache, and freezes the result into an immutable `PreparedTarget`/`PreparedSubrequest` so buffered and streaming sub-requests dial the exact address set that was validated — closing the resolve/validate/dial TOCTOU window where a re-resolution could point the dial at a different (e.g. rebound) address than the one that passed validation.

Highlights:

- **Shape validation** (`UrlTargetError`/`InvalidTarget`): accepts query-only URLs, rejects bracketed IPv4 literals, rejects unsupported schemes and userinfo, and redacts the query string from the target `Debug` output.
- **DNS layer rework**: the process-wide bounded cache now stores the complete address set keyed by the canonical (lowercased) host, behind a detached-owner single-flight so concurrent misses coalesce into one `getaddrinfo` and the resolution still completes even if every caller cancels. An injectable `BlockingLookup` seam backs it (`getaddrinfo` in production, a controllable double in tests).
- **Pin-before-dial**: `prepare_url_target` resolves and pins the peer set before dialing, and `PreparedTarget` is frozen so the same validated peers back both buffered and streaming execution.
- **Preserved public API**: `peer::resolve_addresses` and `peer::is_ip_literal` (consumed by the policy transport connector) are retained, reimplemented on top of the reworked host-keyed cache.
- **End-to-end proof**: an IP-literal HTTPS handshake is driven through the prepared peer to verify IP-SAN validation, with the bind-before-peers invariant guarded at compile time.

Fixes #1092

### Checklist

- [x] Tests added or updated
- [x] `make lint && make test && make test-integration` passes locally

No.


shaneutt (approved):
I see a few things that could use some touch-ups but given the blocker status here I'm inclined to fix-it-forward. 👍

Issue #1136, opened 2026-09-10
filter: publish selected cluster application metadata in HttpFilterContext


leseb:
## Parent

Sub-issue of #1134.

## Purpose

Expose the application metadata belonging to the cluster actually selected for the current exchange. Filters need a typed, read-only contract after routing; headers and writable filter metadata are not suitable because they can leak upstream or be overwritten by filters.

## Dependencies

- #1135

## Design

Introduce framework-owned selected-upstream metadata containing the optional application protocol and provider. Store shared values on the resolved cluster entry and publish them only after successful load-balancer selection.

Expose read-only accessors through `HttpFilterContext`:

- `selected_application_protocol()`
- `selected_application_provider()`

The values should be shared rather than allocated per request.

## Acceptance criteria

- [ ] Resolved cluster entries retain the configured application protocol and provider.
- [ ] Every successful selection path publishes the selected metadata, including ordinary, pinned-endpoint, and panic-mode selection.
- [ ] Failed selection leaves selected metadata absent.
- [ ] Request, response, response-body, and logging filters can read the same immutable values for the remainder of the exchange.
- [ ] Metadata is carried in a typed framework-owned extension, not an upstream header, the transport-only `Upstream` value, or writable string metadata.
- [ ] Transport retry within the selected cluster does not republish or change the metadata.
- [ ] Hot-reloaded pipelines retain the metadata generation pinned to the exchange.
- [ ] Unit and lifecycle tests cover all publication paths and absent values.

## Out of scope

This issue does not add mutable post-selection body access or inference-specific protocol interpretation.

PR #1167, opened 2026-09-14
test(credential): credential-store parity acceptance tests (seeding, rotation, deletion)

### LINKED ISSUE #1160: test: credential-store parity acceptance tests (startup seeding, rotation under traffic, deletion fail-closed)
## Context

The IPP data plane hit two high-severity credential-store bug classes that we must not regress in the Praxis chain:

- **Startup seeding** — secrets created before the gateway started were not seeded into the label-filtered secret cache, so an immediate restart meant an empty credential store (every request 500'd until each secret was touched).
- **Stale credential under rotation** — the reconciler read from a cache that was not the same watch that received the change event, so rotation went unnoticed and traffic kept using the old key (intermittent 401s / stale-credential windows).

The general rule both violate: **the event source and the read source must be the same cache** (single source of truth).

Recommended runtime shape for the migration: the controller renders credential material into mounted secret files, and the proxy consumes them via file-watch (praxis core praxis-proxy/ai#900 landed this; residual gap in praxis-proxy/ai#1058 — hot reload ignores documents referenced only from branch-chain filters). Per-namespace RBAC for direct Secret reads is the fallback only (praxis-extproc#47).

## Requested: acceptance tests (reproduce the bug classes)

1. **Seeding** — create N secrets + routing entries, (re)start the gateway, send traffic immediately → all N credentials usable from the first request, zero secret edits.
2. **Rotation under traffic** — steady traffic to a model, rotate the backing secret mid-flight → every request from the rotation point onward uses the new value; no stale-401 / not-found window; reads must observe exactly what the change-detection watch sees.
3. **Deletion** — deleting a secret fails that route closed (deterministic error) — no stale fallback, no filter-chain crash.

## Open contract questions (decide alongside the tests)

- **Secret selector:** keep the IPP-era label contract (`inference.llm-d.ai/ipp-managed: "true"`, already set by existing producers/docs) as a compat alias, or a praxis-native selector only?
- If rotation latency must be < 60s end-to-end, state the target here — the tests should assert against it.

noyitz:
## Summary

Adds the credential-store parity acceptance tests requested in #1160 — three proxy-level tests that reproduce the two IPP bug classes and pin the rule they both violate: **the event source and the read source must be the same store**, i.e. a request may only ever observe a credential set that some version of the watched source actually published.

Closes #1160

## Scope decision

The tests exercise the credential store the core proxy consumes natively: the `credential_injection` filter rebuilt by config hot reload. The config file is the change-detection watch; the swapped pipeline is the read source. That is the core-side seam where the parity guarantee is observable end-to-end through the proxy binary (`start_reloadable_proxy` + `write_config`, same harness as the `hot_reload` suite). The projected-file consumption path referenced in the issue (`praxis-proxy/ai#900` / `#1136`) lives in the AI filters repo and is covered there by the `credential_inject` filter's own rotation/fail-closed tests.

## Tests added (`tests/integration/tests/suite/credential_store_parity.rs`)

1. **Seeding** — the credential set exists in the config at startup with zero subsequent edits; the first request to each of three routes must already carry its own credential (reproduces the "500 until each secret is touched" class).
2. **Rotation under traffic** — steady traffic while one route's credential rotates: every served request carries exactly the pre- or post-rotation value (never missing, never mixed), the switch is monotonic (once the new value appears, the old never reappears), and untouched routes never change. Acceptance form of "reads observe exactly what the watch sees."
3. **Deletion** — a valid config removing one credential takes effect without the removed value ever being served again afterwards, sibling routes keep their credentials, and the pipeline stays healthy (the route fails closed at the provider boundary with no credential forwarded — no stale fallback at the proxy).

Rotation latency is asserted against a 15 s poll budget (reload debounces at 500 ms), which the issue can adopt as the enforced bound — comfortably inside the < 60 s end-to-end target mentioned there.

## Not answered here

The issue's open contract questions (keep the IPP `inference.llm-d.ai/ipp-managed` label as a compat alias vs. a praxis-native selector) are control-plane producer decisions and intentionally left open — these tests pin the proxy-boundary guarantee regardless of which selector ships.

Pure test change; no runtime code touched. `make lint` clean; the three tests pass locally in ~1 s.


tests/integration/tests/suite/credential_store_parity.rs
```diff
@@ -0,0 +1,314 @@
...
+    loop {
+        let (status, body) = request(addr, path);
+        if pred(status, &body) || Instant::now() >= deadline {
+            return (status, body);
+        }
+        std::thread::sleep(Duration::from_millis(50));
+    }
+}
+
+fn request(addr: &str, path: &str) -> (u16, String) {
```
shaneutt, line 56: By convention testing utilities are placed at the bottom of the file (unless they are pub).

shaneutt (approved):
One small comment from me, and a few from praxis-bot to resolve and then 👍

noyitz:
Taking this out of draft: the only failing check is the supply-chain audit, which @shaneutt confirmed is a known main-side issue that will be fixed shortly, so per that guidance we are not blocking on it. All other checks are green and the credential-store parity acceptance tests are covered.

PR #1173, opened 2026-09-15
fix(rate-limit): serialize token bucket state

### LINKED ISSUE #1016: Token-bucket no-op CAS lets one refilled token be consumed by multiple concurrent callers
> Broken out from praxis-proxy/praxis#1097. Independently verified against commit `a9bb26e7` (adversarial re-read of the source); still present on current `main`.

### 6. Token-bucket no-op CAS lets one refilled token be consumed by multiple concurrent callers

**Severity:** 🟠 Medium

**Location:** `filter/src/builtins/http/traffic_management/token_bucket.rs:61-90 (TokenBucket::try_acquire)`, `filter/src/builtins/http/traffic_management/token_bucket.rs:63-64 (independent loads of tokens and last_refill)`, `filter/src/builtins/http/traffic_management/token_bucket.rs:81-88 (compare_exchange_weak on tokens only, then fetch_max on last_refill)`

**Summary.** TokenBucket keeps `tokens` and `last_refill` in two independent AtomicU64s. try_acquire loads both (lines 63-64), computes a time-based refill, then consumes one token via compare_exchange_weak on `tokens` alone (lines 81-84) and bumps `last_refill` with fetch_max only after success (line 86). When the newly computed token value is bit-identical to the value that was loaded (`new_bits == old_tokens_bits`), the CAS is a no-op and succeeds for every concurrent caller, providing no mutual exclusion. This lets a single accrued token be handed out to multiple racing requests.

**Impact.** Enforcement bypass in a filter registered as `security`. Under concurrency the rate limiter admits more requests per period than configured. The over-issue is bounded by the number of concurrent racers per accrual event (not unlimited), so it is a per-period concurrency-bounded multiplier rather than a full bypass; hence medium, not high.

**Trigger / reachability.** Requires (a) concurrent callers hitting the same bucket, which happens in both modes — Global calls acquire_from_bucket with no lock (limiter.rs:155), and PerIp holds only a DashMap read guard (limiter.rs:171-172), never serializing the RMW — and (b) an accrual event where the post-refill token count returns to the pre-read value bit-for-bit. The reliably reproducible case is a depleted bucket with `old_tokens == burst - 1.0` where the refill caps at `burst`: e.g. burst=1 depleted (old=0.0, refill caps to 1.0, new=0.0, so CAS 0.0->0.0). burst=1 is a valid config (mod.rs:265-269 require only burst>=1 and burst>=rate).

**Reproduction.**

1. `let b = TokenBucket::new(1.0);` 2. Consume the initial token: `b.try_acquire(1.0, 1.0, 0)` returns Some. 3. Spawn N threads that block on a barrier, then all call `b.try_acquire(1.0, 1.0, 1_000_000_000)` (1s elapsed => exactly 1 token accrued, capped at burst=1). 4. Count successes: more than one thread can return Some(0.0) because each reads old_tokens_bits=bits(0.0), computes new_tokens=0.0, and the CAS 0.0->0.0 succeeds for all of them.

**Suggested fix.** Eliminate the split-atomic read-modify-write so the refill+consume+timestamp update is a single atomic transition. Minimal correct option: replace the two AtomicU64 fields with one lock-guarded struct, e.g. `Mutex<BucketState { tokens: f64, last_refill: u64 }>` (parking_lot to match the codebase), and perform load, refill (`tokens = (tokens + elapsed_secs*rate).min(burst)`), the `< 1.0` check, `tokens -= 1.0`, and the `last_refill = last_refill.max(now_nanos)` update inside one critical section, returning the new token count. Contention is per-bucket (per-IP or single global), so cost is minimal. If lock-free is required, switch to a single-atomic GCRA formulation: store one AtomicU64 theoretical-arrival-time (TAT) and CAS it forward by the per-request increment; a single monotonically-advancing atomic has no no-op-CAS window. Do NOT try to patch the existing two-atomic design by making the CAS reject no-ops — the token value legitimately returns to prior values, so only single-atomic or locked state closes the race. Update the "Precision"/"lock-free" doc comment (lines 4, 55-60) accordingly.

_Fix touches:_ `filter/src/builtins/http/traffic_management/token_bucket.rs (TokenBucket struct fields, new, try_acquire, current_tokens, last_refill_nanos, Debug impl)`

**Regression test.** Add a unit test in token_bucket.rs: create `TokenBucket::new(1.0)`, consume the initial token at now=0, then spawn 8 threads that synchronize on a `std::sync::Barrier` and each call `try_acquire(1.0, 1.0, 1_000_000_000)`; assert the total number of `Some` results across all threads is exactly 1 (one token accrued over 1s at rate=1, capped at burst=1). This fails on the current code (multiple successes) and passes after the fix. The existing `concurrent_acquire_total_tokens_bounded` test does not catch it because rate=0 means new_tokens always differs from old_tokens, so no no-op CAS ever occurs.

**Notes.** Severity is medium, not high: over-issue is bounded by concurrent racers per accrual event, and the clearest exploit needs burst=1 (or generally old_tokens == burst-1 with a capping refill) plus precise timing. Rate limiting is inherently approximate, but this is a genuine per-period bounded bypass of a security-registered filter, not mere reporting imprecision. The code's own "burst cap bounds the impact" comment (line 60) does not cover this case, since the duplicated consumption happens within a single period.

jordigilh:
Closes praxis-proxy/ai#1016

## Summary

Fix the token-bucket race that allowed concurrent callers to consume the same refilled token. Refill and consumption now execute as one atomic logical state transition.

## Root Cause

The previous implementation stored tokens and the refill timestamp in separate atomics. A caller could successfully CAS the token count while another caller observed the same refill state before the timestamp update completed.

## Design Decision

The implementation now uses `parking_lot::Mutex<TokenBucketState>` to protect the token count and refill timestamp together. This preserves Praxis semantics for arbitrary fractional rates, burst limits, `current_tokens()`, and `last_refill_nanos()`.

Prior art and alternatives evaluated:

- `ratelimit_meter` uses mutex-protected mutable bucket state, closely matching the compound state required here. It was not adopted as a dependency because this fix is local and Praxis already has established bucket semantics and introspection APIs.
- `governor` uses a single atomic GCRA/TAT state. It can be faster in some rejection cases, but changes the algorithm and does not preserve the existing Praxis state semantics.
- Envoy uses an atomic virtual-time design. It also differs from the existing fractional-token and state-inspection contract.
- Split atomics retain the compound-state race. A virtual-time atomic candidate was slower than the mutex on the successful path.

## Validation

- 15 token-bucket unit tests passed.
- 16 rate-limit integration tests passed, including concurrent, per-IP, global, header, reload, WebSocket, and example coverage.
- Targeted clippy and nightly rustfmt checks passed.
- Native release end-to-end tests passed on ARM64 and AMD64 with Fortio and Vegeta.
- ARM64 32-connection open-path runs measured roughly 50.6-62.0k QPS with 1.5-2.3ms p99 latency.
- AMD64 32-connection open-path runs measured roughly 129.9-135.1k QPS with approximately 0.99ms p99 latency.
- Global and per-IP rejection runs returned the expected initial burst successes followed by `429` responses.
- Release microbenchmarks and `perf stat` captures were collected on both architectures.

These are targeted correctness and performance validations, not a formal production SLA benchmark.


jordigilh:
@nerdalert can you take a look? Thanks!

Cargo.toml
```diff
@@ -65,6 +65,7 @@ percent-encoding = "2.3.2"
 rand = "0.10.2"
 plotters = { version = "0.3.7", default-features = false, features = ["svg_backend", "line_series"] }
 ppe = { version = "0.2.0", package = "praxis-policy", features = ["builtins"] }
+parking_lot = "0.12.5"
 proptest = "1.11.0"
 quote = "1.0.47"
 praxis = { path = "crates/server", package = "praxis-proxy" }
```
shaneutt, line 71: I would like to see an accompanying benchmark for this area, however given the urgency I don't think we need to stop _this_ PR to do it. Please just create a follow-up issue for us to just do a little re-checking over this area that that'll suffice.

shaneutt (approved):
Thanks @jordigilh 👍 

One comment to be resolved and then LGTM

Cargo.toml
jordigilh: I was hesitant to give the figures I got from my 2 hosts, they were not normalized (2 different architectures and hosts) and I agreed with the LLM that it would be mostly noise and false expectations had I added them here as valid figures. 
In any case, prior to posting the PR, I had it run performance metrics on each option in case there was another alternative worth considering. I left the exploratory evaluation as a comment in case we want to revisit this in the future.

I'll evaluate running a performance suite in amd64 host and post back a report (probably a google docs).

PR #1176, opened 2026-09-16
test(bench): preserve token bucket benchmark artifacts

jordigilh:
## Summary

- Add the Criterion token-bucket benchmark target for PR #1173 state-management comparisons.
- Preserve corrected pinned runs, perf counters, host preflight, turbostat telemetry, and the locked split-field retry under `docs/benchmarks/token-bucket-pr1173/`.
- Add a responsive HTML report with graphical overviews, exact data tables, methodology, correctness caveats, and implementation tradeoffs.
- Keep the runnable suite focused on applicable implementations: mutex, locked split fields, and virtual time/GCRA.

## Findings

- The historical split-atomic implementation is excluded from the runnable suite and primary report because token count and refill timestamp are not one linearizable state transition.
- Historical split-atomic logs remain archived only as provenance for rejecting that option.
- The locked split-field retry is correct, but its high-contention advantage disappears after serialization is restored.
- Virtual time/GCRA performs well under contention but changes state and introspection semantics.
- The mutex remains the recommended API-preserving implementation.

## Validation

- Remote `cargo check --benches` passed.
- Remote `cargo test --benches` passed, including the reduced 41-case applicable benchmark suite.
- Three independent locked split-field contention rounds completed at 1, 2, 4, 8, and 16 workers for success and rejection workloads.
- `cargo fmt --manifest-path tests/benches/Cargo.toml -- --check` passed.
- `git diff --check` passed.

This branch is based on `upstream/main` after PR #1173 was merged. No production token-bucket code is changed by this PR.


shaneutt:
Thanks @jordigilh, I don't wanna commit all those logs in main, I was actually thinking we would just create a follow up Github issue to track back to this area later as a reminder, SG? LMKWYT

PR #1179, opened 2026-09-16
fix(subrequest): recap absolute stream deadlines onto the live body

mkoushni:
## Summary
- Add `HttpFilterContext::cap_stream_deadline` and copy the absolute cutoff onto the live `SubResponseBody` after body filters.
- Relative read-timeout recaps restart when downstream backpressure delays the next upstream poll; an absolute deadline prevents `timeout_secs` overruns.

## Test plan
- [x] `cargo test -p praxis-proxy-core cap_stream_deadline`
- [x] `cargo test -p praxis-proxy-filter cap_stream_deadline`


crates/core/src/subrequest/tests.rs
```diff
@@ -1133,6 +1133,64 @@ fn cap_read_timeout_keeps_a_tighter_existing_snapshot() {
...
+    let started = Instant::now();
+    let err = body.next_chunk().await.unwrap_err();
+    let elapsed = started.elapsed();
+    backend.abort();
+    assert!(
+        matches!(err, SubRequestError::DeadlineExceeded),
+        "an expired absolute deadline must fail before starting a fresh relative timer, got: {err}"
+    );
+    assert!(
+        elapsed < Duration::from_millis(50),
```
shaneutt, line 1188: This 50ms upper bound is ~20x tighter than the sibling `cap_read_timeout_shortens_the_next_chunk_wait` 🤔

The other one asserts `elapsed < Duration::from_secs(1)` for the same "fails fast, doesn't start a fresh timer" property.

On a loaded runner, and especially under `cargo-llvm-cov` (see #1190, which just had to skip a timing test for exactly this reason), the `.await` can be scheduled more than 50ms out even though the deadline check itself returns immediately, so this will flake.

`< 1s` still proves the point here (the `idle_timeout` is 30s), with margin some margin for thrashing CI runners.

```suggestion
        elapsed < Duration::from_secs(1),
```

crates/filter/src/filtered_subrequest/streaming.rs
```diff
@@ -332,6 +335,24 @@ fn apply_leftover_read_timeout(body: &mut Option<Box<SubResponseBody>>, leftover
     }
 }
 
+fn leftover_stream_deadline(ctx: &mut crate::filter::HttpFilterContext<'_>) -> Option<std::time::Instant> {
+    ctx.take_stream_deadline_cap()
+}
+
+fn std_instant_to_tokio(deadline: std::time::Instant) -> tokio::time::Instant {
```
shaneutt, line 342: `std_instant_to_tokio`, `leftover_stream_deadline`, `apply_leftover_stream_deadline` has no test. `streaming.rs` has no `mod tests`, and nothing drives `run_step_body_filters`. Please add a direct test for the conversion and, ideally, one that drives a body-filter pass which sets a deadline via `HttpFilterContext::cap_stream_deadline` and asserts the live `SubResponseBody` then fails with `DeadlineExceeded` (the two endpoints are unit-tested, but the wire between them is not).

Also it's minor, but the conversion assumes the std and tokio clocks advance together. That holds in real runtime but not under `tokio::time::pause()`. A one-line comment would save a future paused-time test from silently-wrong deadlines.

crates/filter/src/context.rs
```diff
@@ -542,6 +548,32 @@ impl HttpFilterContext<'_> {
         self.extensions.remove::<StreamReadTimeoutCap>().map(|cap| cap.0)
     }
 
+    /// Tighten the live streaming body's absolute deadline.
+    ///
+    /// Unlike [`cap_stream_read_timeout`](Self::cap_stream_read_timeout), this
+    /// publishes a wall-clock cutoff that the transport checks before each
```
shaneutt, line 554: `Instant` (both `std::time::Instant` and `tokio::time::Instant`) is a *monotonic* clock, not wall-clock it is deliberately immune to system-clock/NTP changes, which is exactly what you want for a deadline.

```suggestion
    /// publishes an absolute (monotonic) cutoff that the transport checks before each
```

crates/core/src/subrequest/body.rs
```diff
@@ -84,6 +84,20 @@ impl SubResponseBody {
         self.read_timeout = Some(self.read_timeout.map_or(timeout, |existing| existing.min(timeout)));
     }
 
+    /// Tighten the absolute stream deadline.
+    ///
+    /// Response-body filters can publish a wall-clock deadline through
```
shaneutt, line 89: Same "wall-clock" wording as `crates/filter/src/context.rs`; `Instant` is monotonic, not wall-clock.

```suggestion
    /// Response-body filters can publish an absolute (monotonic) deadline through
```

shaneutt (approved):
Several comments from praxis-bot, and some from me. Some lint failures. Once those are resolved, 👍

PR #1191, opened 2026-09-17
feat(protocol): run selected-upstream body filters in the Pingora HTTP path

leseb:
## What

Wires the bounded **selected-upstream request-body phase** — the
filter-lifecycle primitive from #1137 (merged as #1171) — into the live
Pingora HTTP path. After an upstream cluster is selected, the proxy now runs
the phase over the buffered request body on the same live `HttpFilterContext`
before forwarding upstream. This is the `crates/protocol` counterpart to the
filter-layer phase and advances the #1134 epic. The IRR/subrequest
integration (#1138) lands separately.

## Details

- **Execution:** `run_pipeline` runs `execute_http_selected_upstream_request_body`
  on the same live `HttpFilterContext` immediately after `execute_http_request()`
  completes upstream selection — no new context, no re-buffering. It is gated on
  a successful request action, a selected upstream, and
  `needs_selected_upstream_request_body`.
- **Isolation:** the adapted body is stored separately (`adapted_request_body` /
  `retained_adapted_request_body` / `adapted_request_body_len`); the canonical
  `pre_read_body` is never overwritten, and the adapted representation drains
  exclusively (preserving the #1138 boundary).
- **Limits:** a 413 is enforced against the resolved effective request-body
  limit; `Content-Length` is stamped and `Transfer-Encoding` stripped on the
  adapted body to avoid CL/TE ambiguity.
- **Retry:** reseeds only the adapted representation, and the retry replay guard
  accounts for the adapted length.

## Tests

- End-to-end integration coverage of the phase — adaptation, empty-body
  injection, reject, retry replay, and size-limit rejection
  (`tests/integration/tests/suite/selected_upstream_body.rs`).
- Lifecycle docs updated (`docs/architecture/pingora-handler-lifecycle.md`).

`make lint` and `make build` are clean; the workspace unit tests pass.

Closes #1139

Issue #1139, opened 2026-09-10
protocol: run selected-upstream body filters in the Pingora HTTP path


leseb:
## Parent

Sub-issue of #1134.

## Purpose

Integrate the selected-upstream request-body phase into ordinary Pingora HTTP forwarding after the request pipeline selects a cluster and before any upstream connection or transport attempt.

## Dependencies

- #1136
- #1137

## Required lifecycle

```text
pre-read bounded body
  -> existing pre-route body filters
  -> request filters and upstream selection
  -> selected-upstream body filters
  -> final size and framing enforcement
  -> upstream transport
```

## Acceptance criteria

- [ ] The phase runs after the complete request-filter pipeline and before DNS resolution, connection establishment, or transport.
- [ ] Successful adaptation replaces only the exchange-local body forwarded by Pingora.
- [ ] Rejection returns the requested local response without attempting an upstream connection.
- [ ] Final framing removes stale `Transfer-Encoding`, sets authoritative `Content-Length`, and correctly represents an empty body.
- [ ] The final adapted size participates in existing body-size and retry-body guards.
- [ ] Transport retries replay identical adapted bytes and never rerun request filtering or adaptation.
- [ ] Retry accounting remains based on the larger of the original downstream body and final adapted body.
- [ ] Pipelines without selected-upstream participants retain their existing streaming and allocation behavior.
- [ ] Integration tests cover byte preservation, mutation exactly once, rejection before transport, oversized transformed output, empty output, framing repair, and retry replay.

PR #1197, opened 2026-09-18
feat(filter): run selected-upstream body phase in filtered subrequests

### LINKED ISSUE #1138: filter: run selected-upstream body filters in filtered subrequests and IRR
## Parent

Sub-issue of #1134.

## Purpose

Give every `FilteredSubrequestExecutor` exchange, including each `iterative_request_router` iteration, the same selected-upstream transformation semantics as ordinary Pingora forwarding.

## Dependencies

- #1136
- #1137

## Required lifecycle

```text
canonical nested body
  -> existing nested body filters
  -> nested request filters and upstream selection
  -> selected-upstream body filters
  -> final size and framing enforcement
  -> SubRequest transport
```

Both transport paths must call the same pipeline operation rather than maintaining separate transformation implementations.

## Acceptance criteria

- [ ] Every filtered subrequest runs the shared selected-upstream body operation after selection and before peer construction or transport.
- [ ] A new filtered exchange begins with no selected-upstream metadata inherited from its parent.
- [ ] Successful selection publishes metadata for that exchange only.
- [ ] Selected metadata is not leaked back into the parent exchange when the nested exchange finishes.
- [ ] Each IRR iteration starts from its own canonical or continuation body.
- [ ] Adapted bytes from iteration N never become the implicit source for iteration N+1.
- [ ] Rejection and transformed-size failure occur before any nested network activity.
- [ ] Request framing reflects the final adapted body.
- [ ] Tests prove parity with the normal HTTP path, including native byte preservation, mutation exactly once, iteration isolation, rejection, and oversized output.

leseb:
## What

Runs the bounded selected-upstream request-body phase inside filtered
subrequests, including the internal-redirect-request (IRR) path — the
filter-layer counterpart to the Pingora HTTP-path wiring merged in #1191
(#1139). After a subrequest selects/prepares an upstream, the executor runs
the phase over the buffered request body on the same HttpFilterContext
before dialing, so body/provider filters that adapt a request for the chosen
upstream also run for internally-issued subrequests.

## Details

- Execution: the filtered-subrequest executor runs the selected-upstream
  request-body phase after upstream selection, gated on a successful request
  action, a selected upstream, and needs_selected_upstream_request_body.
- Staged-upstream invariant: the phase receives a mutable ctx and runs after
  the pre-phase re-pin, so the executor re-pins the staged upstream after the
  phase (and clears republished discarded-selection metadata). A body-phase
  ctx.upstream rewrite cannot redirect a callout — whose adapted body may
  itself carry a credential — off the prepared destination.
- Selected-cluster isolation: selected-cluster metadata is scrubbed across
  the subrequest boundary so a subrequest neither leaks nor inherits the
  parent's selection.
- Limits: the retained-state ceiling is enforced after the phase, mirroring
  every other filter boundary — an oversized retention (for the IRR, the
  retained IterationState) is rejected with 413 before dialing rather than
  carried across the dial and caught only post-transport.
- Shared limit: the body-size limit is now a FilterPipeline method so the
  Pingora and filtered-subrequest paths share one definition.

## Tests

- Unit coverage in crates/filter/src/filtered_subrequest/tests.rs, including
  fail-first regression tests for the post-phase re-pin and the pre-dial
  retained-state ceiling.
- End-to-end integration coverage via the IRR path
  (tests/integration/tests/suite/selected_upstream_body_subrequests.rs).

make lint and make build are clean; the workspace unit tests pass.

Closes #1138


crates/filter/src/pipeline/mod.rs
```diff
@@ -257,6 +257,27 @@ impl FilterPipeline {
         self.request_body_ceiling
     }
 
+    /// Effective request-body byte limit for the selected-upstream phase.
+    ///
+    /// The adapted output is checked against the merged `StreamBuffer` limit
+    /// clamped to the listener ceiling — the same bound that governed the
+    /// canonical pre-read — never an individual filter's limit. `min` keeps this
+    /// correct whether or not `request_body_mode` is already clamped at build.
+    #[must_use]
```
alexsnaps, line 279: Unsure why that `must_use` is there, but it also doesn't hurt. Not asking to remove it. 

<details>

The strong/original intent is for values where _dropping the result changes program behavior or hides a consequence_:
- Result / Option you should have checked,
- RAII guards (`MutexGuard`) where discarding immediately releases the lock,
- lazy things (`Future`, iterator adapters, builders) where discarding means the work never happens.

This isn't the case here, and I don't think we have the `must_use_candidate` lint enabled, in which case the argument here would be:

>for a pure function, if you ignore the return value the entire call was dead code, so you probably made a mistake (typo'd, or meant to call an effectful method).

Which, personally, I think is real weak. 

</details>

crates/filter/src/filtered_subrequest/mod.rs
```diff
@@ -977,7 +984,63 @@ impl FilteredSubrequestExecutor {
...
+                }
+                // Enforce the retained-state ceiling before dialing, like every
+                // preceding filter boundary. A phase filter may have grown
+                // cross-sub-request state (for the IRR, the retained
+                // `IterationState`); an oversized retention must be rejected here
+                // rather than carried across a dial and held until the response or
+                // timeout.
+                if self.accounting.exceeds_limit(&filter_ctx.extensions) {
+                    return Ok(RawResponse::Rejected(Rejection::status(413)));
+                }
```
alexsnaps, line 1034: Should we log, or otherwise differentiate these 2? Thinking of debugging something... But we can also cross that bridge when we get there.

alexsnaps (approved):

shaneutt (dismissed):
Once all other comments are resolved, 👍

crates/filter/src/filtered_subrequest/mod.rs
```diff
@@ -977,7 +984,63 @@ impl FilteredSubrequestExecutor {
             // itself — can only ever reach the authority it was prepared for.
             if let Some(upstream) = &pinned_upstream {
                 filter_ctx.upstream = Some(upstream.clone());
+                // #1138 Correction 2: a staged upstream carries no application
+                // metadata. Any value a chain filter published for the discarded
+                // load-balancer selection would misdescribe the pinned
+                // destination, so the body phase must not read it.
+                filter_ctx.extensions.remove::<SelectedClusterApplication>();
             }
```
alexsnaps: [nit] This a dup of [1038:1042]? Should we extract that?

crates/filter/src/filtered_subrequest/mod.rs
leseb: would like to defer this, it's a good observability idea but many other paths return 413. So it's an established pattern.

crates/filter/src/filtered_subrequest/mod.rs
leseb: will extract in a helper now

leseb: https://github.com/praxis-proxy/praxis/pull/1197/commits/73aa4b5e908c43d10a8f1b2651427b03175c1e0b

crates/filter/src/pipeline/mod.rs
```diff
@@ -257,6 +257,27 @@ impl FilterPipeline {
         self.request_body_ceiling
     }
 
+    /// Effective request-body byte limit for the selected-upstream phase.
+    ///
+    /// The adapted output is checked against the merged `StreamBuffer` limit
+    /// clamped to the listener ceiling — the same bound that governed the
+    /// canonical pre-read — never an individual filter's limit. `min` keeps this
```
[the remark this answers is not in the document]
leseb: https://github.com/praxis-proxy/praxis/pull/1197/commits/73aa4b5e908c43d10a8f1b2651427b03175c1e0b

crates/filter/src/filtered_subrequest/mod.rs
alexsnaps: we'll also cross that bridge when we get there, good with me!

PR #1198, opened 2026-09-18
fix(filter): resolve named chains inside iterative_request_router steps

leseb:
### What does this PR do?

An `iterative_request_router` (IRR) step built its pipeline with an empty named-chain map, so a chain-binding filter nested in a step could bind an inline outbound chain but a `ChainRef::Named` reference always failed as an unknown chain, and IRR was the only registry factory that never received the containing `ChainBindingContext`. This threads the containing `ChainBindingContext` into IRR: the registry now hands `HttpWithRegistry` factories the full context, and a new `ChainBindingContext::build_nested_step_pipeline` builds each step as a continuation of the containing build. Steps now resolve top-level named `filter_chain`s, while the shared cycle-detection stack, outbound-nesting depth, and materialization/branch budgets carry across the IRR boundary instead of resetting. Context-free entry points (`from_config`, tests) use a standalone context with empty chains and the strict default posture, so existing inline chains, application-registered filters, and unknown-chain rejection are unchanged.

N/A — no tracked issue (regression observed on Praxis 0.5.6).

### Checklist

- [x] Tests added or updated (5 focused core tests: named-chain resolution in a step, inline-chain still works, unknown-chain rejected, cross-boundary cycle rejected, shared materialization budget)
- [x] Documentation updated (if applicable) — rustdoc comments updated; no user-facing docs affected
- [x] `make lint && make test` pass locally; `make doc` passes. `make test-integration` not run locally (config-time-only change, no runtime/proxy behavior change) — CI covers it.

No. No config or wire-format change; the behavior change is purely additive (a top-level named `filter_chain` referenced from inside an IRR step now resolves instead of erroring). All changed APIs are `pub(crate)`/private.


leseb:
What we are addressing in this PR, suppose Praxis AI registers a web_search_callout filter and the operator defines a reusable outbound chain:

```
filter_chains:
  - name: search-egress
    filters:
      - filter: request_id
      - filter: headers
        request_add:
          - name: Accept
            value: application/json

  - name: main
    filters:
      - filter: iterative_request_router
        initial_step: model
        steps:
          - name: model
            filters:
              - filter: web_search_callout
                outbound_chain: search-egress
            on_result:
              - default: true
                done: true
```

Previously, while building the IRR step, Praxis effectively saw no named chains:

```
outbound_chain: search-egress
                ↓
Error: unknown chain 'search-egress'
```

After this diff, the IRR step inherits the top-level chain table:

```
search-egress
    ├── request_id
    └── headers
```

The configuration now builds successfully, and web_search_callout receives that prebuilt two-filter outbound pipeline.

alexsnaps (approved):
Changes look good to me. nitpick: this isn't really purely additive, as things could now fail earlier, but good change in any case!

Thanks for fixing this 🙏 

Should we eventually:
- add a test for the total-branch budget shared across IRR steps? 
- a test for a branch chain inside an IRR step resolving a top-level named chain (builtin-only; previously error'ed "unknown chain")?
- add a runtime integration test (and, for the branch-chain path, an example config) exercising a named chain resolved inside an IRR step end-to-end?

PR #1201, opened 2026-09-18
feat: add the experimental cloud_events filter.

### LINKED ISSUE #1202: feat: Implement CloudEvents Publisher filter
Implement CloudEvents publisher filter, based on the architectural discussion and the description from the parent epic.

This issue is mostly about the implementation of the code

rikatz:
### What does this PR do?

Adds a cloud_events filter that can be used to ship observability data to remote cloud_events capable receivers

Fixes #1202 

### Checklist

- [X] Signed off all commits (`git commit -s`)
- [X] Tests added or updated
- [X] Documentation updated (if applicable)
- [X] `make lint && make test && make test-integration` passes locally

No


shaneutt:
You have made the PR size check angry.

shaneutt:
You have also made the lint check angry!

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+            response_headers: None,
+            emitted: false,
+        });
+        self.maybe_emit(ctx);
+        Ok(FilterAction::Continue)
+    }
+
+    async fn on_response(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
+        if self.on == CloudEventsPhase::ResponseHeaders {
+            let Some(response) = ctx.response_header.as_ref() else {
```
shaneutt: **bug:** `on_response` never skips informational (1xx) responses, so over an HTTP/1.1 upstream it fires once per interim response *and* once for the final one.

Scenario: an H1 upstream sends `103 Early Hints` (or `100 Continue` for an `Expect: 100-continue` upload) then `200 OK`. `on_response` runs first for the 1xx — it captures `status = 100/103`, calls `insert_filter_state(… emitted: false)`, and emits an event (this ResponseHeaders branch emits immediately; the ResponseComplete branch also emits immediately because `bodyless_response(1xx) == true`). When the final `200` arrives, `on_response` runs again and `insert_filter_state` overwrites the state with a fresh `emitted: false`, so a **second** event is published. Net: duplicate events, the first carrying the wrong (interim) status. Pingora surfaces 1xx to this hook for H1 upstreams (see how `compression.rs` already early-returns on informational responses).

Fix: guard both `on_response` branches, e.g. right after reading `response`:
```rust
if response.status.is_informational() {
    return Ok(FilterAction::Continue);
}
```

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+        BodyAccess::None
+    }
+
+    fn on_response_body(
+        &self,
+        ctx: &mut HttpFilterContext<'_>,
+        _body: &mut Option<Bytes>,
+        end_of_stream: bool,
+    ) -> Result<FilterAction, FilterError> {
+        if end_of_stream && self.on == CloudEventsPhase::ResponseComplete {
```
shaneutt: **bug (gap):** in `response_complete` mode the event is emitted only on the terminal `on_response_body(end_of_stream = true)` (or the bodyless / no-response path in `on_response`). That terminal body hook is *not* delivered when the client disconnects mid-body or the upstream errors mid-stream, so those responses silently produce no event.

Scenario: a long streamed inference response (the motivating `inference.tokens.used` case) where the client disconnects before end-of-stream — tokens were consumed upstream but no CloudEvent is published. `access_log` handles exactly this via an always-run `logging`-phase fallback (see the comment in `observability/mod.rs` about "streamed responses aborted mid-body"); `cloud_events` has no equivalent.

Consider finalizing in the logging phase (or documenting the drop) so aborted responses are still accounted.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+};
+
+/// Maximum receiver response body retained by a best-effort publication.
+const MAX_DELIVERY_RESPONSE_BYTES: usize = 4_096;
+
+/// Maximum serialized CloudEvent size accepted.
+const MAX_EVENT_SIZE_BYTES: usize = 65_536; // 64 KiB
+
+/// Maximum CloudEvents delivery tasks across all publishers.
+const MAX_PENDING_DELIVERIES: usize = 64;
```
shaneutt: **dos/observability:** `DELIVERY_PERMITS` is a single process-global semaphore of 64 shared across *every* `cloud_events` publisher and every config generation, and each permit is held for the whole delivery (up to `delivery_timeout`, default 3s).

Scenario: one publisher whose receiver is slow/unreachable holds permits ~3s each and starves an unrelated publisher, whose events are dropped as `backpressure`; on a high-QPS gateway emitting a per-request event, most events are silently dropped even when receivers are healthy. Fine as a hard safety cap, but easy to hit unknowingly. Worth documenting the cap (and the `praxis_cloud_events_skipped_total{reason="backpressure"}` signal), and/or making it configurable.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+        // TODO (followup): add a dedicated background/event pool if event bursts
+        // contend with request-critical callouts.
+        crate::metrics::record_cloud_events_attempt();
+        tokio::spawn(async move {
+            // Hold the permit until delivery completes so the semaphore bounds active
+            // deliveries, not only task creation.
+            let _permit = permit;
+            let started = Instant::now();
+            let deadline = Instant::now() + timeout;
+            let target = match praxis_core::connectivity::prepare_url_target(&destination, deadline, |_| Ok(())).await {
```
shaneutt: **security (low):** the delivery passes a no-op address-validation hook (`|_| Ok(())`) to `prepare_url_target`, whose docs state that deciding whether private/loopback addresses are allowed is the hook's job. As written the publisher will deliver to loopback, link-local (e.g. the `169.254.169.254` cloud-metadata endpoint), and RFC1918 addresses with no `insecure_options.allow_private_endpoints` gate.

`destination` is static operator config, not client-controlled, so this isn't client-driven SSRF — but it silently bypasses the private-endpoint posture applied elsewhere. Worth confirming it's intended, or wiring the same private-IP policy other outbound paths use.

crates/filter/src/metrics.rs
```diff
@@ -70,6 +82,54 @@ pub(crate) fn record_lb_panic_mode(cluster: SharedString) {
...
+        _ => "other",
+    }
+}
+
+#[cfg(feature = "cloud-events-filter")]
+/// Record one CloudEvents publication attempt.
+pub(crate) fn record_cloud_events_attempt() {
+    counter!(
+        CLOUD_EVENTS_PUBLISH_TOTAL,
+        "outcome" => "attempt",
```
shaneutt: **nit:** `attempt` is recorded on the same `praxis_cloud_events_publish_total` counter as `success`/`failure`. Since every attempt later increments `success` or `failure` too, `sum(praxis_cloud_events_publish_total)` double-counts each delivery (attempt + outcome). Consider a separate `*_attempts_total` counter, or dropping the `attempt` series, so the total is meaningful without always filtering by `outcome`.

crates/filter/src/builtins/http/observability/mod.rs
```diff
@@ -13,3 +13,15 @@ pub use access_log::{
 };
 pub use request_id::RequestIdFilter;
 pub use trace_context::TraceContextFilter;
+
+#[cfg(feature = "cloud-events-filter")]
+mod cloud_events;
+#[cfg(feature = "cloud-events-filter")]
+pub use cloud_events::CloudEventsFilter;
+
```
shaneutt: nit: missing the standard comment separators here

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2024 Praxis Contributors
```
shaneutt:
```suggestion
// Copyright (c) 2026 Praxis Contributors
```

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+use serde::Deserialize;
+use serde_json::{Map, Value};
+use tokio::sync::Semaphore;
+
+use crate::{
+    BodyAccess, FilterAction, FilterError,
+    factory::parse_filter_config,
+    filter::{HttpFilter, HttpFilterContext},
+};
+
```
shaneutt, line 28: missing the standard comment separators here

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+/// Maximum serialized CloudEvent size accepted.
+const MAX_EVENT_SIZE_BYTES: usize = 65_536; // 64 KiB
+
+/// Maximum CloudEvents delivery tasks across all publishers.
+const MAX_PENDING_DELIVERIES: usize = 64;
+
+/// Maximum explicitly configured mapped values in one event.
+const MAX_EVENT_MAPPINGS: usize = 64;
+
+static DELIVERY_PERMITS: LazyLock<Arc<Semaphore>> = LazyLock::new(|| Arc::new(Semaphore::new(MAX_PENDING_DELIVERIES)));
```
shaneutt, line 196: This is a single, process global semaphore of 64 shared across _every_ `cloud_events` publisher and every config generation, and each is held for the whole delivery up to `delivery_timeout` (default 3s) 🤔 

So this has the potential to cause starvation. Because we're going to `experimental` first, if you would like to push this back to a later iteration that's fine, just create a follow-up issue for us to come back to it (probably need an epic for cloud events if we don't already have one?)

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+        let request = praxis_core::subrequest::SubRequest {
+            method: Method::POST,
+            // `PreparedTarget::bind` replaces this with the destination path.
+            uri: Uri::from_static("/"),
+            headers: HeaderMap::from_iter([(CONTENT_TYPE, HeaderValue::from_static("application/cloudevents+json"))]),
+            body: Bytes::from(body),
+        };
+
+        // share the runtime subrequest pool;
+        // TODO (followup): add a dedicated background/event pool if event bursts
```
shaneutt: Make sure this and _any_ other `TODO` or `FIXME` items get a follow-up issue created, and sub-task them into the epic for this feature.

rikatz:
fix  is on its way

Also no rush, I will test this implementation on monday with metering service

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+        ctx.insert_filter_state(CloudEventsState {
+            status: None,
+            response_headers: None,
+            emitted: false,
+        });
+        self.maybe_emit(ctx);
+        Ok(FilterAction::Continue)
+    }
+
+    async fn on_response(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
```
shaneutt, line 649: I think there's a bug in here 🤔

`on_response` never skips information 1xx responses so over an HTTP/1.1 upstream it fires once per interim response and once for the final one. This might end up with duplicated events?

I think we should add something like this:

```rust
if response.status.is_informational() {
    return Ok(FilterAction::Continue);
}
```

but check it out and see what you think. Make sure we get test coverage for this.

crates/filter/src/builtins/http/observability/cloud_events.rs
shaneutt, line 731: I think there's a bug here too perhaps 🤔

I don't think we're covering when the client disconnects mid-body, or the upstream errors mid-stream, so we silently drop the event however in the AI case, tokens were consumed.

We should add handling for client and upstream aborts.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -0,0 +1,1488 @@
...
+        assert!(request_text.starts_with("POST /events HTTP/1.1\r\n"));
+        assert!(
+            request_text
+                .to_ascii_lowercase()
+                .contains("content-type: application/cloudevents+json")
+        );
+        assert!(request_text.contains("\"specversion\":\"1.0\""));
+        server.await.unwrap();
+    }
+}
```
shaneutt, line 1611: Ahhhh I love to see a wall of tests 🧘

crates/filter/src/builtins/http/observability/mod.rs
```diff
@@ -13,3 +13,15 @@ pub use access_log::{
...
+/// Header names rejected at config load time in v1.
```
shaneutt: Missing the standard comment separators here too

crates/filter/src/metrics.rs
shaneutt: `attempt` is recorded on the same `praxis_cloud_events_publish_total` counter as `success`/`failure`.

Every attempt later increments `success` or `failure` too, `sum(praxis_cloud_events_publish_total)` double-counts each delivery (attempt + outcome).

Consider adding something like a separate `*_attempts_total` counter or dropping the `attempt` series so the total is meaningful without always filtering by `outcome`?

rikatz:
linter problem jnow is now my fault! :P

crates/server/Cargo.toml
```diff
@@ -56,6 +56,7 @@ config-reload = ["dep:notify", "praxis-protocol/config-reload"]
 admin-api = ["praxis-protocol/admin-api"]
 dev = ["basic-auth-filter"]
 basic-auth-filter = ["praxis-filter/basic-auth-filter", "experimental"]
+cloud-events-filter = ["praxis-filter/cloud-events-filter", "experimental"]
```
shaneutt, line 59: 👍

examples/configs/observability/cloud-events.yaml
```diff
@@ -0,0 +1,42 @@
+# CloudEvents Publisher
```
shaneutt, line 1: we need to make sure all experimental filters note that they're experimental in their example configs, add it here and if you're feeling froggy maybe we could catch this in CI too 🤔

tests/schema/tests/suite/examples/observability/cloud_events.rs
```diff
@@ -0,0 +1,25 @@
...
+    let backend_port = start_backend("cloud-events");
+    let proxy_port = free_port();
+    let receiver_port = free_port();
+    let config = crate::example_utils::load_example_config(
+        "observability/cloud-events.yaml",
+        proxy_port,
+        HashMap::from([("127.0.0.1:3000", backend_port), ("127.0.0.1:3001", receiver_port)]),
+    );
+    let _proxy = start_proxy(&config);
+}
```
shaneutt, line 225: Perhaps there are some more test cases we could add here?

shaneutt (dismissed):
Great work thank you @rikatz.

I have several comments, and praxis-bot does too. Most of them are minor, though I think I found a couple bugs in there.

Once they're all resolved and tests have been added to cover everything and prove those bugs are resolved, 👍

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: copy/paste ftw! fixed, will push it soon

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: added

crates/filter/src/builtins/http/observability/mod.rs
rikatz: added, pushing soon

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: https://github.com/praxis-proxy/praxis/issues/1203

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: similar to https://github.com/praxis-proxy/praxis/issues/1203 so I have updated the issue to cover a better design for this connection and will use it as part of this followup

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: changing this right now would impact on some other parts of Praxis and Pingora handler (like a deferred_record creation).

I will open an issue for it, as a bug fix, just so I can fix it on a separate PR without blasting changes on every different file

rikatz: https://github.com/praxis-proxy/praxis/issues/1228

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: made codex add a fix, will push it soon

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: :saluting_face:

examples/configs/observability/cloud-events.yaml
rikatz: https://github.com/praxis-proxy/praxis/issues/1229 for the linter, I wouldn't add it here because the reviewer told me that my PR is already too big :P

crates/filter/src/metrics.rs
rikatz: pushing soon

tests/schema/tests/suite/examples/observability/cloud_events.rs
rikatz: added some more on my last commit

examples/configs/observability/cloud-events.yaml
shaneutt: </img>

PR #1233, opened 2026-09-19
feat(filter): Remove drops on filter_metadata

alexsnaps:
I actually don't necessarily want _this_ changeset merged, but I certainly want to start the discussion on what the stance should be.

`HttpFilterContext::set_filter_metadata` was trying to impose a set of rules on both key and values being inserted. Sadly, the field being public, none of those rules are actually enforced and lose their usefulness. In our AI filters we are currently shoving way bigger values [in there](https://github.com/praxis-proxy/ai/pull/1229#pullrequestreview-5241331446), so I guess that ship has sailed?

`filter_metadata` isn't the only issue. There are more fields being public, which are somewhat protected, yet anyone can change them regardless of these guards.


crates/filter/src/context.rs
```diff
@@ -705,21 +705,18 @@ impl HttpFilterContext<'_> {
     pub fn set_metadata(&mut self, key: impl Into<String>, value: impl Into<String>) {
```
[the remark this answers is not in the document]
alexsnaps: Well, they are still "limited", they just aren't dropped anymore. 
As I guess that's my question here... And really, I'd maybe even delete this `fn` entirely. Looks like most code just goes straight for the public field.

shaneutt: Contract drift: the doc comment just above still says "Keys are limited to 64 bytes and values to 256 bytes to bound per-request memory growth." After this change the method only warns and always inserts, so that documented invariant no longer holds. The same promise is repeated in the docs that depend on it: `json_body/mod.rs` ("filter_metadata values over 256 bytes are dropped") and `json_body/config.rs` ("capped at 256 bytes by the context"). If the stance is to drop the caps, those docs have to change with it, otherwise callers keep coding against a guarantee that is gone.

crates/filter/src/context.rs
```diff
@@ -705,21 +705,18 @@ impl HttpFilterContext<'_> {
...
         let key = key.into();
         let value = value.into();
-        if key.is_empty() || key.len() > 64 {
-            tracing::warn!(key_len = key.len(), "metadata key rejected (must be 1-64 bytes)");
-            return;
+        if key.len() > 64 {
```
shaneutt: Two behavior changes are folded into this line. (1) The old predicate was `key.is_empty() || key.len() > 64`; dropping it entirely means an empty-string key is now inserted with no warning at all, unlike every other over-limit case which at least warns. If the new stance is "warn but insert," an empty key should presumably still warn for consistency. (2) A key over 64 bytes is now stored; `get_metadata` and its consumers (access_log field projection, cloud_events) were written assuming bounded keys. Worth deciding explicitly whether empty and oversized keys are now valid rather than letting the first fall through silently.

crates/filter/src/context.rs
```diff
@@ -705,21 +705,18 @@ impl HttpFilterContext<'_> {
...
+            tracing::warn!(key_len = key.len(), "metadata key >64 bytes");
         }
         if value.len() > 256 {
-            tracing::warn!(key = %key, value_len = value.len(), "metadata value rejected (max 256 bytes)");
-            return;
+            tracing::warn!(key = %key, value_len = value.len(), "metadata value >256 bytes)");
```
shaneutt: This is the cap I would push back on removing. The "field is public so the rules aren't enforced" argument covers a *trusted* in-process filter that can already bypass by touching `filter_metadata` directly. But the enforcement value of this cap is for *untrusted* request data copied in through the safe API. `json_body`'s `request_extract` copies request-body JSON into metadata via `set_metadata` (`json_ops/rewrite.rs` `flush_to_store`), and the deleted `extract_oversized_metadata_is_skipped` test guarded exactly that path: a 257-byte body value used to be dropped, now it is stored verbatim. `filter_metadata` is also `mem::take`d into subrequest continuations (`filtered_subrequest/continuation.rs`) and read by access_log/cloud_events, so an attacker-controlled body value now amplifies unbounded into sub-requests and log records. Separately, the warn string has a stray closing paren: `"metadata value >256 bytes)"`.

crates/filter/src/context.rs
```diff
@@ -705,21 +705,18 @@ impl HttpFilterContext<'_> {
...
         }
         if !self.filter_metadata.contains_key(&key) && self.filter_metadata.len() >= MAX_METADATA_ENTRIES {
```
shaneutt: With the `return` gone, `MAX_METADATA_ENTRIES` no longer bounds the map: a filter driven by per-request data (multiple `json_body` extracts, or keys derived from request content) can grow `filter_metadata` without limit, not just past a large single value. For the stance discussion, note the sibling per-request map `FilterResultSet` (`results.rs`) still hard-rejects oversized keys/values (its `set` returns `Err`, it does not warn-and-accept). If the conclusion is that these maps should not be policed, that should apply to both; otherwise the codebase ends up with one map that silently accepts anything and one that rejects, for no visible reason.

crates/filter/src/context.rs
shaneutt: Doc drift (key/value caps). The method doc just above (lines 703-704) still says "Keys are limited to 64 bytes and values to 256 bytes to bound per-request memory growth," but after this change the method only warns and always inserts. The same now-false promise is repeated in every doc that leans on it: `json_body/mod.rs:82` ("filter_metadata values over 256 bytes are dropped") and `json_body/config.rs` at lines 37 ("values are capped at 256 bytes by the context"), 117 ("Values over 256 bytes are dropped by the context"), and 125. If the caps are intentionally dropped, all of these have to change together, otherwise callers keep coding against a guarantee that no longer holds.

crates/filter/src/context.rs
shaneutt: Two behavior changes are folded into this line. (1) The old predicate was `key.is_empty() || key.len() > 64`; dropping it entirely means an empty-string key is now inserted with no warning at all, while every other over-limit case at least warns. If the new stance is "warn but insert," the empty-key case should presumably still warn, for symmetry. (2) Keys over 64 bytes are now stored, and `get_metadata` plus its consumers were written assuming bounded keys. Worth deciding explicitly whether empty and oversized keys are valid rather than letting the empty case fall through silently.

crates/filter/src/context.rs
shaneutt: This is the cap I would push back on removing, and the "the field is public so the rules aren't enforced" premise does not actually hold for the write paths. Grepping the workspace, every production per-entry writer goes through `set_metadata` (grpc_web, grpc_detection, sticky_sessions, load_balancer, policy `llm.model`, `json_ops` rewrite, grpc_timeout); the only direct `filter_metadata.insert` calls are in tests, and the production direct assignments are whole-map phase swaps, not entry insertions. So the cap was load-bearing in practice. In particular `json_body`'s `request_extract` copies untrusted request-body JSON in through `set_metadata` (`json_ops/rewrite.rs`), and the deleted `extract_oversized_metadata_is_skipped` test guarded exactly that: a 257-byte body value used to be dropped and is now stored verbatim. The map is then moved into subrequest continuations (`filtered_subrequest/streaming.rs:135`) and read by access_log (`access_log.rs:606`) and cloud_events, so an attacker-controlled body value amplifies unbounded into sub-requests and logs. Separately, the warn string has a stray closing paren: `"metadata value >256 bytes)"`.

crates/filter/src/context.rs
shaneutt: Removing this `return` drops the entry-count bound and leaves the file internally inconsistent. `MAX_METADATA_ENTRIES` becomes a misnomer, and its own doc (lines 73-78) is now false on both clauses: it claims "Individual keys and values are already size-bounded (64 / 256 bytes)" (no longer true) and warns that "without an entry count cap a filter chain could insert thousands of unique keys per request" (which is now precisely what it permits). The sibling `MAX_STRUCTURED_METADATA_NAMESPACES` doc (line 85) still says it "Mirrors the entry cap on filter_metadata," but that mirror is broken: `set_structured_metadata` (line 909) still hard-rejects both its namespace and key caps, and `FilterResultSet` (`results.rs`) still hard-rejects oversized keys/values. After this PR, one per-request map silently accepts anything while two siblings reject. If the stance is that these caps are not worth enforcing, apply it consistently and fix the docs; otherwise keep the `return`.

crates/filter/src/context.rs
```diff
@@ -705,21 +705,18 @@ impl HttpFilterContext<'_> {
...
             tracing::warn!(
                 key = %key,
                 entries = self.filter_metadata.len(),
-                "metadata entry rejected (max {MAX_METADATA_ENTRIES} entries)"
+                "metadata is above threshold (max {MAX_METADATA_ENTRIES} entries)"
             );
-            return;
         }
         self.filter_metadata.insert(key, value);
```
shaneutt: Test coverage: `set_metadata` is now effectively `insert` plus advisory warnings, but this change removes all six tests that pinned the old contract (the five `set_metadata_rejects_*` / `set_metadata_allows_overwrite_at_limit` tests in this module, plus `extract_oversized_metadata_is_skipped` in `json_body/tests.rs`) and adds none for the new behavior, so the new semantics are entirely untested. If this direction is kept, add tests asserting that oversized values, oversized and empty keys, and entries beyond `MAX_METADATA_ENTRIES` are now inserted (and ideally that the warnings fire), so the intended new contract is actually pinned rather than silently assumed.

shaneutt:
Hardening this is valuable. I'll mark it draft for the moment because we talked about it and there's a bit more to discuss and document, but I would like to target `v0.7.0` for this (~2w).

PR #1235, opened 2026-09-19
chore(gh): Fix template, `id` can't have `?`

alexsnaps:
Sorry, but til...


alexsnaps:
Also, in case it happens again, I also learned today that if you go to the template file (on main only?) on GH, you'd get [the reason explained](https://github.com/praxis-proxy/praxis/blob/main/.github/ISSUE_TEMPLATE/enhancement.yml) why the template isn't working

alexsnaps:
You fixed it! Thanks... and sorry for the initial faulty PR :facepalm:

PR #1237, opened 2026-09-19
fix: low-severity correctness and hardening batch from #1117

### LINKED ISSUE #1117: Low-severity cleanup batch: 47 confirmed correctness/robustness/hardening defects (L1-L47)

szedan-rh:
Address the findings from #1117 that still reproduce on main and have a small, local fix. Each change restores one documented invariant at the boundary where it originates; no new abstractions or config.

- config: bound consistent_hash rings, validate response-condition status codes and header names, reject unbalanced IPv6 brackets, catch aliases with hyphenated anchors, list every endpoint field
- filters: distinct-endpoint P2C sampling, normalized Content-Type parameter matching, result-set entry cap, exact access-log sampling, config-time header-name checks, duplicate credential rules rejected, zero half-open timeout rejected, CSRF origin from path-less URLs, atomic sticky-session expiry, host plugin factories built outside the registry lock, warn on body-phase terminal responses
- protocol: HTTP probes require an HTTP status line, h2 probes tolerate split reads and open connections, released StreamBuffer request bodies honor the global ceiling
- core: retry budget clamps on every refill, Connection tokens parsed from raw bytes
- tls: client cert/key pairs validated at config time, SNI parser rejects duplicate extensions and framing errors
- server: load the config from the path that will be watched
- docs: correct priority capacity, evict_idle, and terminal-filter docs

Findings already fixed, intentional, or needing a design decision are listed in the PR description.
## Findings addressed

| Finding | Change |
| --- | --- |
| L1 | `Config::load_from` loads from the resolved path the watcher uses |
| L4 | HTTP probe requires an `HTTP/` status line |
| L5 | released StreamBuffer bodies honor the global ceiling |
| L7 | h2 probe reads the SETTINGS header exactly and bounds the close |
| ... | one row per L-number in this diff |

## Not in this PR

- L6: already fixed on main by c19338f6.
- L17, L22, L46: documented as intentional in the source.
- L2, L3, ...: need a design decision, tracked in #NNNN.

## Behavior changes

- Result sets are capped at 128 entries (`MAX_ENTRIES`, matching the metadata cap); a filter writing more gets an error on the 129th key.
- ClientHellos with a duplicate SNI extension or bytes after the extension block are treated as having no SNI (RFC 8446 §4.2).
- Endpoint hosts that are not IP literals must satisfy RFC 1035 hostname syntax, the same rule `endpoint_selector` already applies; a name such as `my_backend:80` is now a config error. A single trailing dot (`backend.example.com.:80`) is accepted as the fully-qualified form.

Refs #1117


szedan-rh:
L2 — probes respawn on every reload

**L2 (not fixed here):** `reload_pipelines` rebuilds the health registry and respawns every probe generation on each successful reload, even when no cluster changed. `carry_over_health_state` copies only the `healthy` flag, so `consecutive_failures` resets and an endpoint at N-1 failures can never trip if reloads arrive faster than `unhealthy_threshold × interval`. I left it out because `reload_carries_unhealthy_endpoint_state` (03bf36ec) explicitly asserts a fresh registry on identical config, so this needs a maintainer call on intended reload semantics.

I think the best approach is to make probe lifetime per cluster instead of per generation. Replace the single CancellationToken with a map keyed by cluster name. On reload, diff old and new clusters with the existing serde-based config_value_changed: unchanged clusters keep their registry entry and running task, changed or added clusters get a fresh entry and task, removed clusters are cancelled. Carry the full EndpointHealth (both hysteresis counters) for unchanged clusters, not just the flag. Flip the existing test to assert reuse on identical config and keep reload_resets_health_when_check_config_changes as the "changed" case. Check that the health gauges are neither reset nor double-registered, and add one line to docs/operating/health-checking.md.

L3 — referenced-document watches pinned to startup

**L3 (not fixed here):** the reload watcher builds its path filter, watched directories, and composite-hash inputs once from the startup pipelines. Documents first referenced after a reload are neither watched nor hashed, so edits to them never trigger a reload. `WatcherParams::referenced_files` already documents this as deferred work, and the rewatch attempted in #1124 drew three objections (hashing a re-read instead of the loaded bytes, no retry on failed re-registration, symlink rotations). It needs its own design rather than a batch fix.

I think the best approach is to raise two PRs. First, in the filter crate, have referenced_files() return the path plus the digest of the bytes the pipeline actually consumed at build time, so the watcher records the loaded version and an intervening edit shows up as a change. Second, in the server watcher: have handle_reload return the new referenced set, diff it against the live watch set, unwatch removed directories and watch new ones, and on a watch() failure leave reload_pending set so the existing backoff loop retries registration instead of advancing state. Reuse the main config's symlink-aware PathFilter for referenced directories so Kubernetes ..data rotations are accepted. Tests: a document referenced only after reload triggers a reload, a failed registration is retried, and a symlink rotation is detected.

L13 — header conditions read only the first value

**L13 (not fixed here):** request and response header conditions use `HeaderMap::get`, so a header sent as several field-lines only matches on its first line. The router already uses any-value semantics (`headers_match` in `router/matching.rs`), and the request side now goes through `HeaderSource` plus the pre-read overlay's ambiguity error, so switching semantics touches a trait and three implementations. Worth a decision on the intended matching rule before changing it.

I think the best approach is to define the rule once: a condition matches when any field-line of that header equals the expected value, consistent with the router and RFC 9110 list semantics. Add header_matches(name, expected) to HeaderSource with a default built on header(), override it for Request with get_all().iter().any(), and let EffectiveHeaders keep its ambiguity error for conflicting trusted mutations while falling through to the any-value rule for untouched raw headers. Change the response side to get_all at the same time so both phases agree. Document the rule in docs/filters/README.md and add tests for both phases and the overlay.

L22 — post-response JSON-RPC denial envelopes can be truncated

**L22 (not fixed here):** when a post-phase policy denial replaces a JSON-RPC result that is shorter than the denial envelope, `fit_to_original_length` slices the envelope at an arbitrary byte, so the client gets invalid JSON. That truncation is documented as the last-resort backstop, but `mod.rs` still promises an envelope "fitted to the committed Content-Length". #1170 added exactly the missing mechanism for the LLM path (`llm_error_envelope_bytes_within`); the JSON-RPC path never got it.

I think the best approach is to mirror #1170: add json_rpc_error_envelope_bytes_within(violation, id, max_len) with the same degradation tiers, full envelope, then without data, then a shortened message cut on a char boundary, then code only, then {}. Use it at the three post-phase deny sites and keep fit_to_original_length as the backstop it is documented to be. Add a property-style test that for every budget from zero to the full length the output fits and, when non-empty, parses as JSON, matching the existing a_tiny_budget_still_yields_parseable_json test.

L24 — sticky-session cluster entries accept unknown keys

**L24 (not fixed here):** `ClusterSessionConfig` flattens `PersistenceConfig`, and serde's `deny_unknown_fields` does not work on either side of a `flatten`, so a typo like `cookie_nmae:` inside a cluster entry is silently ignored while the top-level struct rejects typos. Adding the attribute is not a fix; the struct needs the same raw-struct `try_from` pattern `RouteRaw` uses.

I think the best approach is to follow RouteRaw: declare ClusterSessionConfigRaw with every field flat (name, type, cookie_name, cookie_attributes, header_name, ttl_secs, failover, max_entries, eviction) and deny_unknown_fields, then #[serde(try_from = "ClusterSessionConfigRaw")] on the public type. The TryFrom builds PersistenceConfig from type and rejects fields that belong to the other persistence type, so cookie_name under type: header is an error too. Keep the public struct shape unchanged so callers are untouched. Tests: unknown key rejected, cross-type field rejected, and make test-schema proves every example config still parses.

L25 — Host-derived SNI dropped on retry reselection

**L25 (not fixed here):** the first attempt fills a missing cluster SNI from the request `Host`, but the retry path builds the alternate upstream from the reselector's captured cluster TLS config, which never saw the request. A retry to an IP endpoint on a `verify: false` cluster with no `sni` sends no SNI at all. The six-line fix is to copy the previous attempt's SNI forward in `upstream_peer.rs`, but that puts the fallback policy in the protocol crate, and there is no unit seam for `reselect_on_retry`. Flagging for a decision on where the fallback should live.

I think the best approach is to keep the policy in the filter crate where the initial fallback lives. Extract the "fill SNI from Host" logic in entry.rs into one helper and give EndpointReselector::build_upstream an sni_fallback: Option<&str> parameter that applies the same helper. The protocol retry path passes the previous attempt's resolved SNI from ctx.upstream_for_retry. Unit-test the helper and the reselector in the filter crate, and add one integration retry test against a TLS backend with verify: false and no sni, which is the reachable case.

L30 — repeated request_add for one header collapses

**L30 (not fixed here):** two `request_add` entries for the same header each read the immutable client snapshot and queue a `Set`, so the last one wins. The absent-header case collapses too because every header-phase consumer applies `extra_request_headers` with replacement semantics, and that is deliberate (the doc on `apply_pending_header_mutations` explains `request_id` dedup depends on it). Both attempts in #1124 were rejected for incomplete semantics, so this needs an agreed rule first.

I think the best approach is to decide the rule in the filter, not the protocol: repeated request_add entries for one name are grouped in config order and comma-joined into a single value using the existing join convention, for both the existing-header and absent-header cases. This leaves the protocol's replacement semantics and the request_id guarantee untouched and also fixes the filtered sub-request path, which was the second objection. Document the join in docs/filters/http/transformation/headers.md and add tests for existing, absent, and sub-request cases. If the team would rather not define list semantics, the alternative is to reject duplicate names at config time, which is eight lines and zero runtime change.

L34 — health-check hostnames and DNS rebinding

**L34 (not fixed here):** config-time SSRF validation checks IP literals properly but hostnames only against a lexical blocklist, so a name that later resolves to loopback or link-local passes. A validator cannot prevent rebinding by construction, and resolving at validation time would be a TOCTOU and would break offline `--validate`. The runtime guard added in fb9b9827 (`resolve_address_checked`) covers the data plane but the four probe types still call `TcpStream::connect` directly.

I think the best approach is to enforce at connect time. Route every probe connect (tcp, http, h2, grpc) through one helper that resolves the target with resolve_address_checked and connects to the checked IP, not the hostname. Plumb the relevant insecure flag into the probe config, defaulting consistently with allow_private_endpoints since probes target the cluster's own endpoints. Keep the lexical check at config time as early feedback. Test that a hostname resolving to loopback is refused unless allowed, with the helper shared so one test covers all four probe types.

Sibling gap from L41 — other Connection parsers

**Follow-up to L41:** this PR fixes the raw-byte tokenizing only in `subrequest/internals.rs`, which is what the finding named. The same `to_str()`-then-skip pattern exists in `protocol/.../hop_by_hop.rs` (`strip_connection_tokens`, `strip_hop_by_hop_header_map`) and `filter/src/filtered_subrequest/sanitize.rs`, so an obs-text byte in a `Connection` value still disables nominated-header stripping on those paths.

I think the best approach is to one shared connection_tokens(&HeaderValue) -> impl Iterator<Item = &str> in core/src/reserved_headers.rs, which is already the canonical home for hop-by-hop rules, used by all three call sites. That is roughly 25 lines across three crates and removes the duplicated tokenizer rather than adding a fourth copy.

Behavior note on L14

**L14 behavior note:** P2C now always compares two distinct endpoints. With exactly two endpoints under equal load, weight acts only as the tie-breaker, so the heavier endpoint always wins instead of a ~9:1 split. That is what the struct doc already promised, and with three or more endpoints weight still biases which pair is sampled, but it is a visible traffic change for two-endpoint weighted clusters.

I think the best approach is to accept it and state it in the load-balancer docs, since it is the documented P2C semantics. If a graded weight influence on two-endpoint clusters is wanted, the honest option is to document that weighted_round_robin or random is the strategy for weight-proportional splits, rather than making P2C's tie-break probabilistic.

@shaneutt

alexsnaps:
Shouldn't we have broken this up? At least in separate commits, if we really want one single PR for this?

szedan-rh:
Agreed. The batch itself was scoped by #1117, but one commit was the wrong shape for review and bisect. I'll rewrite the branch once, at your request, into one commit per finding with the L-number in each subject, grouped by crate in the order the description lists them. From there every change is a new commit.

alexsnaps:
Thanks a ton for splitting this up @szedan-rh - now, once this reviewed and mergable, please don't squash 🙏

shaneutt:
Ping me when ready, I can do a cherry-pick to keep the history nice.

szedan-rh:
@alexsnaps  Sure thing, @shaneutt - Thank you

leseb (changes requested):
response.rs:139: Content-Type parameter values are always compared case-insensitively. Multipart boundaries are case-sensitive, so boundary=ABC wrongly matches boundary=abc. Only parameter names—and specifically case-insensitive values such as charset—should be folded. [RFC 9110 §8.3.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1)

response.rs:145-151: Splitting parameters on every semicolon breaks quoted values. boundary="foo;one" and "foo;two" both become foo and can wrongly match. Use a quote-aware media-type parser.

probe.rs:139-144: The health probe accepts malformed status lines such as HTTP/nope 200 ready and HTTP/1.1 0200 weird. Validate the HTTP version and exactly three status digits. [RFC 9112 §4](https://www.rfc-editor.org/rfc/rfc9112.html#section-4)

endpoints.rs:83-103: The L33 fix only checks brackets and empty hosts. Invalid addresses such as bad host:80 still pass configuration validation and fail later when connecting. Validate hostname syntax too.

access_log.rs:238: YAML sample_rate: .nan passes validation. Every sampling comparison then fails, silently disabling access logs. Reject non-finite rates.

szedan-rh:
@leseb all five fixed, one commit each on top of the existing history (no rewrite):

- e4a416e4 `access_log`: `sample_rate` has to be finite now. NaN passed the range check because both comparisons are false.
- 09012f6f probe: version must be `HTTP/` DIGIT `.` DIGIT and the status exactly three digits, so `HTTP/nope 200` and `HTTP/1.1 0200` no longer count.
- 5a82e60b Content-Type: only `charset` values fold case. `boundary=ABC` vs `boundary=abc` is a mismatch again.
- 59d7f207 Content-Type: parameters split on `;` outside quoted strings only, with `\"` honored inside quotes.
- b4f3e186 endpoints: non-literal hosts now go through `praxis_tls::dns::validate_dns_hostname`, the same check `endpoint_selector` already applies to its `host:port` values. IP literals are accepted with or without brackets, since the runtime resolver strips them either way. This makes `my_backend:80` a config error; added under "Behavior changes" in the description.

Each has a regression test that failed before the change.

crates/core/src/config/validate/cluster/endpoints.rs
```diff
@@ -75,19 +75,48 @@ fn validate_endpoint_address(addr: &str, cluster_name: &str) -> Result<(), Proxy
...
+    }
+    if bracketed.is_some() {
+        return Err(ProxyError::Config(format!(
+            "cluster '{cluster_name}': endpoint '{addr}' is not an IPv6 address (expected '[addr]:port')"
+        )));
+    }
+    // Same rule `endpoint_selector` applies to its `host:port` values: a
+    // name outside RFC 1035 syntax (`bad host`, `my_backend`) cannot be
+    // resolved, so it can only fail at connect.
+    praxis_tls::dns::validate_dns_hostname(host).map_err(|err| {
```
shaneutt: Low severity, please confirm this is intended. `validate_dns_hostname` splits on `.` and rejects an empty label, so a trailing-dot FQDN such as `backend.example.com.:80` becomes a config error even though it is a valid, resolvable name (the resolver treats the trailing dot as a fully-qualified, search-domain-free lookup). The comment just above states such a host "can never resolve, so it can only fail at connect," which is not true for the trailing-dot case, and the PR description's list of behavior changes does not mention it. If rejecting a trailing dot is deliberate, it would be worth calling out; otherwise consider stripping a single trailing `.` before validating.

crates/filter/src/condition/response.rs
```diff
@@ -127,9 +127,66 @@ fn has_parameters(value: &str) -> bool {
...
+/// Values are compared exactly, except `charset`, which RFC 9110 §8.3.1
+/// defines as case-insensitive (`charset=UTF-8` equals `charset=utf-8`).
+/// A multipart `boundary` is case-sensitive and must not fold.
+///
+/// Compares in place rather than normalizing into owned strings, since this
+/// runs on the response path for every evaluation of such a condition.
+fn params_match(actual: &str, expected: &str) -> bool {
+    params(actual).count() == params(expected).count()
+        && params(expected).all(|(name, value)| {
+            params(actual).any(|(other_name, other_value)| {
```
shaneutt: `params_match` checks equal parameter counts plus "every expected param appears in actual" which tests subset not a true multiset comparison.

If a parameter name is duplicated I expect this can pass on unequal values (e.g. actual charset=utf-8; foo=bar matches expected charset=utf-8; charset=utf-8). The impact of this is probably pretty small, but while we here a per-name multiset compare (or rejecting duplicate names) would seal it up.

shaneutt (commented):
I leave the approval to @leseb and @twghu, but I added a couple of minor comments 🖖

crates/core/src/config/validate/cluster/endpoints.rs
szedan-rh: Not intended. Fixed in feedfcb6: a single trailing dot is stripped before the label check, so `backend.example.com.:80` is accepted and `.:80` still fails as an empty hostname. Regression test `accept_trailing_dot_fqdn_endpoint` failed before the change. The PR description's behavior-change line now says so.

One pre-existing sibling I did not widen into this PR: `is_ssrf_sensitive_hostname` compares the raw host, so `localhost.:80` passes the lexical blocklist. That was already true on main before L33, and the runtime `resolve_address_checked` guard still catches it at connect.

crates/filter/src/condition/response.rs
szedan-rh: Confirmed: with a repeated name the check was a subset test. Fixed in 1b5fdcd1: every parameter must now occur the same number of times on both sides, which with equal totals is multiset equality, so neither a duplicated name nor an extra parameter can pass. Still compared in place; the scan is quadratic in the parameter count, which is tiny for a Content-Type. Tests: `content_type_duplicate_parameter_is_not_a_subset_match` (your `charset=utf-8; charset=utf-8` case) and `content_type_parameters_compare_as_multiset` (`a=1;a=1;b=2` vs `a=1;b=2;b=2`), both failing before the change.

szedan-rh:
@leseb on the follow-up you sent over Slack (`response.rs:185-189`, quotes stripped but quoted-pairs left encoded): fixed in 28d8365a. Values are now compared through a char iterator that removes one pair of quotes and decodes `\x` to `x`, so `boundary="foo\;bar"` equals `boundary="foo;bar"` (RFC 9110 §5.6.4) while `"foo;baz"` still mismatches. Unquoted tokens are unchanged and the comparison stays allocation-free. Test `content_type_quoted_pair_decodes_before_compare` failed before the change.

Shane's two inline comments are addressed in 1b5fdcd1 (multiset parameter compare) and feedfcb6 (trailing-dot FQDN accepted), replies in the threads. Three commits on top of the existing history, no rewrite.

szedan-rh:
@leseb - could you please take another look?

leseb (approved):

shaneutt:
Don't merge: doing history-preservation

shaneutt:
manually cherry-picked

PR #1238, opened 2026-09-20
fix(reload): keep live pipeline when a listener's protocol changes

### LINKED ISSUE #1113: Hot reload swaps a protocol-mismatched pipeline behind the live listener handler, silently dropping all filters
> Broken out from #1097. Independently verified against commit `a9bb26e7` (adversarial re-read of the source); still present on current `main`.

### 3. Hot reload swaps a protocol-mismatched pipeline behind the live listener handler, silently dropping all filters

**Severity:** 🟠 Medium

**Location:** `server/src/reload.rs:137-147 (swap loop, no protocol/address/TLS guard)`, `server/src/reload.rs:149-151 (unconditional listener_meta.store)`, `server/src/pipelines.rs:130-140 (filters_unsupported_by validates against the NEW protocol)`, `server/src/server.rs:235-247 (register_protocols binds handlers once from startup config)`, `server/src/reload_diagnostics.rs:75-88 (detect_protocol_changes_with only warns)`, `filter/src/pipeline/http.rs:70-74 and 157-159 (AnyFilter::Tcp(_) => continue)`

**Summary.** Protocol handlers and sockets are registered exactly once at startup (server.rs:235-247): an HTTP handler is created only if a startup listener is HTTP, a TCP handler only if one is TCP, and each handler holds the ArcSwap slot keyed by listener name. Hot reload (reload.rs:137-147) iterates the new config's listeners and calls live.swap(name, new_arc) for every name whose slot exists, with no comparison of old vs new protocol/address/TLS. resolve_pipelines validates each candidate pipeline against the NEW listener protocol (pipelines.rs:130), so a listener changed in place from HTTP to TCP (with only TCP filters) produces a TCP-validated pipeline that passes and is swapped into the still-running HTTP handler's slot. HTTP execution then skips every TCP filter (http.rs:72-74, 158-159 match AnyFilter::Tcp(_) => continue), so all previously-configured HTTP controls silently disappear. reload.rs:149-151 also unconditionally stores listener_meta_from_config(new_config), making admin GET /api/pipelines report the new protocol/address/TLS state that the still-bound socket does not reflect. detect_protocol_changes_with only emits a "requires restart" warning; it does not gate the swap.

**Impact.** An in-place protocol change on reload takes the live listener out of correct service and silently removes configured security/traffic controls: an HTTP listener keeps accepting HTTP but runs a pipeline whose filters (auth, ip_acl, rate_limit, cors, csrf, router/load_balancer, etc.) are all skipped, so requests are processed with no filtering (and, with no HTTP router/lb, likely no upstream). The reverse (TCP->HTTP in place) is the same class. Admin /api/pipelines misreports protocol/address/TLS, so an operator inspecting live state sees values the socket does not actually honor, defeating incident diagnosis. The "requires restart" warning is emitted but the inconsistent pipeline is installed anyway.

**Trigger / reachability.** Operator edits the running config to change an existing listener's protocol (http<->tcp) in place, keeping the same listener name, with a filter set valid for the new protocol, then triggers a reload (file watcher). Not remotely reachable; requires config-file write access. The change is explicitly warned about ("requires restart"), but the swap still proceeds.

**Reproduction.**

1. Start with a listener named `web`, protocol http (default), whose chain enforces a control (e.g. ip_acl deny + static_response, or basic_auth). Confirm requests are filtered.
2. Edit the config so listener `web` becomes protocol: tcp with a valid TCP upstream and a TCP-only chain; save to trigger reload.
3. resolve_pipelines validates the candidate as TCP (pipelines.rs:130 uses listener.protocol = tcp) -> passes. reload.rs:140-141 sees live.get("web") is Some (the HTTP handler's slot) and swaps the TCP pipeline in.
4. Send an HTTP request to `web`'s original socket: execute_http_request hits AnyFilter::Tcp(_) => continue for every filter, so nothing is enforced (previously-configured HTTP controls are gone).
5. GET /api/pipelines reports protocol=tcp / new address / new TLS state though the socket is still the startup HTTP socket. A TLS-off->TLS-on edit similarly reports tls=true while the socket still accepts plaintext.

**Suggested fix.** Gate the swap on restart-only listener identity in reload.rs. Build an index of old_config.listeners by name, and in the swap loop skip any listener whose protocol differs from the old config's listener of the same name (this is the generation the live handler was built for) — leave the old pipeline live and record it as restart-blocked instead of swapping. Sketch:

```rust
let old_by_name: std::collections::HashMap<&str, &praxis_core::config::Listener> =
    old_config.listeners.iter().map(|l| (l.name.as_str(), l)).collect();
let new_by_name: std::collections::HashMap<&str, &praxis_core::config::Listener> =
    new_config.listeners.iter().map(|l| (l.name.as_str(), l)).collect();
let mut restart_blocked: Vec<&str> = Vec::new();
for name in new_pipelines.listener_names() {
    // Protocol change requires a restart: the handler/socket were bound once
    // at startup for the old protocol. Swapping a pipeline validated for the
    // new protocol behind that handler makes every filter get skipped
    // (AnyFilter::Tcp/Http mismatch in http.rs / tcp execution).
    let protocol_changed = matches!(
        (old_by_name.get(name), new_by_name.get(name)),
        (Some(o), Some(n)) if o.protocol != n.protocol
    );
    if protocol_changed {
        restart_blocked.push(name);
        continue;
    }
    if let Some(new_slot) = new_pipelines.get(name) {
        let new_arc = new_slot.load_full();
        if live.get(name).is_some() {
            live.swap(name, new_arc);
            swapped.push(name);
        } else {
            skipped.push(name);
        }
    }
}
```

Then make the metadata store consistent with what is actually live: start from listener_meta_from_config(new_config) but, for every name in `restart_blocked`, overwrite that entry with the still-live value from listener_meta_from_config(old_config) (or from old_by_name), so /api/pipelines reports the protocol/address/TLS the socket actually honors rather than the unapplied new values. The existing detect_protocol_changes_with warning already tells the operator a restart is needed. (Address/TLS changes do not cause filter-skipping since the pipeline stays HTTP/HTTP, so they need not block the pipeline swap; but their meta should likewise reflect the live socket, which the restart_blocked meta handling can be extended to cover if desired.)

_Fix touches:_ `server/src/reload.rs (reload_pipelines: swap loop at ~137-147 and listener_meta.store at ~149-151)`

**Regression test.** In server/src/reload.rs tests, add a test that sets up live HTTP pipelines for listener `web` via setup_live_pipelines(), captures the old pipeline Arc pointer, then calls reload_pipelines with a new_config where `web` is protocol: tcp with a valid TCP-only chain. Assert (a) the reload returns Ok, (b) live.get("web") still points to the ORIGINAL (HTTP) pipeline Arc — i.e. the protocol-mismatched pipeline was NOT swapped in, and (c) meta.load().get("web").protocol is still http (not tcp), so admin metadata matches the still-live socket.

**Notes.** Severity is medium, not high: the trigger is an operator misconfiguration (in-place protocol change) that is not remotely reachable and is already accompanied by a "requires restart" warning. The escalating factor is that the warning is advisory only — the inconsistent pipeline is installed regardless, silently disabling live controls. Address-only and TLS-toggle changes do not cause filter-skipping (pipeline stays same-protocol) and are lower severity; only the admin-metadata misreport applies to them, which the same meta fix can address. Existing test reload.rs `valid_reload_swaps_pipeline` uses an address change (not a protocol change) so it remains valid under the proposed protocol-only gate.

szedan-rh:
Fixes #1113.

Hot reload installed a pipeline validated for the new listener protocol behind a handler bound at startup for the old one. The handler skips every filter of the other protocol, so an in-place `http -> tcp` (or `tcp -> http`) edit silently removed all configured controls, and `/api/pipelines` reported the unapplied protocol.

**Change**

- `ListenerPipelines` records the protocol each listener's pipelines are resolved for. It is set at construction and a swap never changes it, so for the live pipelines it is the protocol the handler was bound for. This survives a listener being removed from the config and re-added later, which the config-derived metadata store does not.
- `reload_pipelines` checks the new config against that record before building anything. A protocol change on a live listener returns an error: nothing is swapped, no probes are respawned, and the metadata stores are untouched. The watcher keeps the previous config on `Err`, so the error repeats on every reload until the edit is reverted or the process restarts.
- The restart-required warning for protocol changes in `reload_diagnostics` is unreachable now and is removed with its test. Docs updated.

Rejecting the whole reload, rather than skipping one listener, keeps live pipelines, health probes, and admin metadata on a single generation. The cost is that other changes in the same edit also wait until the protocol change is reverted.

**Tests**: `protocol_change_rejects_reload_and_leaves_live_state_untouched` (pipeline pointer, both metadata stores, and the live probe token are unchanged), `protocol_change_is_rejected_again_on_the_next_reload`, `protocol_change_from_tcp_rejects_reload`, and `protocol_change_after_remove_and_readd_rejects_reload`. `ListenerPipelines` has unit tests for the recorded protocol and its survival across a swap.

Address and TLS changes publishing unapplied socket metadata predates this PR and is left as is.


leseb (changes requested):
P1: Its documented remove/re-add gap reintroduces #1113: metadata disappears, the guard is skipped, and a cross-protocol pipeline can again replace the live slot and silently skip every filter. [Guard logic](https://github.com/praxis-proxy/praxis/blob/69d958f9ce3fde7036b6c6d8ee56c9d4639e0629/crates/server/src/reload.rs#L153-L180)

P2: A blocked listener retains its old HealthRegistry, but reload cancels its probes and starts probes only for the new registry. Health state freezes until restart. [Health lifecycle](https://github.com/praxis-proxy/praxis/blob/69d958f9ce3fde7036b6c6d8ee56c9d4639e0629/crates/server/src/reload.rs#L304-L343)

P2: Admin state remains inconsistent: same-protocol address/TLS changes publish unapplied socket metadata, while cluster_meta always advances even when a listener retains its old cluster generation.

szedan-rh:
@leseb - Tanks for the review
All three hold up, so I changed the approach rather than patch the guard.
P1 – the metadata store was the wrong reference. It is rebuilt from the config on every reload, so a removed listener loses its entry while its handler and pipeline slot stay alive, and re-adding it with the other protocol walked straight past the check. The bound protocol is now recorded on ListenerPipelines at startup and a swap never changes it. There is a test for the remove/re-add sequence.

P2 (health) and P2 (admin state) – both come from applying the reload partially. The blocked listener kept a pipeline pinned to a registry whose probes had just been cancelled, and the metadata stores moved on without it. Instead of keeping one listener on an old generation, the reload now fails as a whole when a listener's protocol differs from the one its handler was bound for: no swap, no probe respawn, no metadata store. The watcher keeps the previous config on Err, so the same error is logged on every reload until the edit is reverted.

The trade-off is that other changes in the same edit wait too. I think that is the right call for a change the process cannot honor, and it matches how an invalid logging config is already treated. If you would rather keep partial application, say so and I will rework it.
The restart-required warning for protocol changes in reload_diagnostics is unreachable now, so it is removed with its test. Docs updated to match.
The address/TLS metadata point predates this PR. I left it alone here; happy to open an issue for it.

szedan-rh:
@leseb - Could you please take another look?

szedan-rh:
Went through the bot review point by point.

**Redundant warning after rejection**: not the case. `reject_protocol_changes` returns before `log_restart_required_changes` is reached, and this PR removes the protocol detector from `reload_diagnostics.rs` along with its `protocol_changed_detected` test. There is no second warning path.

**Error message formatting**: `ProtocolKind` is a unit enum with a derived `Debug`, so the message reads `from Http to Tcp`. Same formatting the unsupported-filter error in `pipelines.rs` already uses.

**`ListenerPipelines::new()` bypassing the protocol record**: fair. Every remaining caller is under `#[cfg(test)]` or a doctest; the only production constructor is `with_protocols` in `resolve_pipelines`. I'd rather not widen this PR for it, but I'm happy to tighten the API in a follow-up if you think it's worth it.

**Missing tests**: added both in 2ead2961.
- `protocol_change_on_one_listener_rejects_reload_for_all`: two bound HTTP listeners, one flips to TCP. Neither pipeline is swapped and both metadata stores keep their pointer.
- `new_tcp_listener_alongside_bound_http_listener_is_accepted`: a TCP listener added by reload has no bound handler and no recorded protocol, so the reload applies, `web` gets its rebuilt pipeline, `db` is skipped, and its metadata is published.

leseb (approved):

PR #1241, opened 2026-09-21
feat(filter): match selected upstream in filter conditions

### LINKED ISSUE #1140: filter: add conditions for selected upstream application metadata
## Parent

Sub-issue of #1134.

## Purpose

Let ordinary filter conditions match the application protocol and provider published by upstream selection. This keeps application adaptation and path changes declarative while preserving ownership: routing selects the cluster, conditions gate filters, and existing rewrite filters own URL or path translation.

## Dependencies

- #1136

## Proposed configuration

```yaml
conditions:
  - when:
      selected_upstream:
        application_protocol: openai_chat_completions
        application_provider: vllm
```

Both predicates are optional; fields that are present must match exactly.

## Acceptance criteria

- [ ] Generic request conditions can match selected `application_protocol`, selected `application_provider`, or both.
- [ ] Matching reads the typed selected-upstream metadata rather than headers or writable filter metadata.
- [ ] Missing metadata does not accidentally satisfy a configured predicate.
- [ ] Pipeline validation rejects a selected-upstream condition unless a load balancer is guaranteed to execute earlier on every reachable path.
- [ ] Conditions are evaluated during the normal request phase and are not reevaluated during body transformation.
- [ ] Existing `path_rewrite` and `url_rewrite` filters can use these conditions; provider and protocol filters do not set `ctx.rewritten_path` directly.
- [ ] Unit and pipeline-validation tests cover protocol-only, provider-only, combined, missing, mismatch, and invalid ordering cases.
- [ ] Reference documentation includes the new condition shape and lifecycle constraint.

## Failure behavior

A consuming application adapter remains responsible for failing closed when required metadata is missing or unsupported. A non-match alone must not silently forward an incompatible representation.

leseb:
### What does this PR do?

Lets ordinary filter conditions (`when`/`unless`) match the `application_protocol` and `application_provider` that load-balancer upstream selection publishes, so a filter can be gated on the chosen upstream's metadata. Only the `load_balancer` builtin publishes this metadata; when no upstream has been selected yet, or the selected cluster carries none, the condition fails closed rather than matching on absent data. Condition ordering is validated as a forward reachability dataflow over the pipeline (including branch chains and `SkipTo`/`ReEnter` edges), rejecting configs where a gated filter could run before any load balancer, and `selected_upstream` matcher values are validated at config time against the identifiers clusters actually declare. Ships with unit, integration, and an end-to-end example-config test (`examples/configs/pipeline/selected-upstream-conditions.yaml`).

Fixes #1140

### Checklist

- [x] Tests added or updated
- [ ] `make lint && make test && make test-integration` passes locally

No. `selected_upstream` is a new optional condition field; existing configs and behavior are unaffected.


leseb:
> ## PR Review: feat(filter): match selected upstream in filter conditions
> **Summary**: Adds `selected_upstream` condition matching for gating filters on load-balancer-published application metadata (protocol/provider). Includes config types, validation, and dataflow reachability checks.
> 
> **Overall Assessment**: The core design is sound and the validation approach is thorough. However, there are significant test coverage gaps that must be addressed before merging, particularly around edge cases for matcher validation and reachability analysis.
> 
> Severity	Count
> Critical	0
> Large	7
> Medium	1
> ### Large Severity Gaps
> **1. Missing test: provider-only SelectedUpstreamMatch parsing**
> 
> The code supports both `application_protocol` and `application_provider` independently, but tests only cover `protocol-only` and `both fields` cases. A `provider-only` test is needed to verify serde deserialization works correctly when only the provider field is specified.
> 
> Expected test:
> 
> ```rust
> #[test]
> fn parse_selected_upstream_provider_only() {
>     let m: ConditionMatch = serde_yaml::from_str(
>         r#"
> selected_upstream:
>   application_provider: vllm
> "#,
>     ).unwrap();
>     let su = m.selected_upstream.expect("selected_upstream should parse");
>     assert_eq!(su.application_provider.as_deref(), Some("vllm"));
>     assert!(su.application_protocol.is_none());
> }
> ```
> 
> **2. Missing test: matcher pair validation (protocol and provider must be on same cluster)**
> 
> The validation logic collects `(protocol, provider)` pairs from clusters, and the `DeclaredUpstreams` struct comment explicitly states that when both fields are specified, they must come from the SAME cluster. However, no test verifies rejection when:
> 
> * Cluster A declares `protocol=openai, provider=vllm`
> * Cluster B declares `protocol=anthropic, provider=bedrock`
> * Matcher specifies `protocol=openai, provider=bedrock` (drawing from different clusters)
> 
> This should be rejected because no single cluster has both values, but I see no test covering this critical case.
> 
> **3. Missing test: selected_upstream matcher references inline cluster**
> 
> The validation calls `collect_inline_clusters` to include clusters defined in `load_balancer` filter configs, but no test verifies this works. Need a test where:
> 
> * A matcher references `application_provider: xyz`
> * No top-level cluster defines `xyz`
> * A load_balancer filter's inline `clusters:` list defines a cluster with `application_provider: xyz`
> * Config validation should ACCEPT this (proving inline clusters are checked)
> 
> **4. Missing test: selected_upstream condition in inline branch chain is validated**
> 
> The code in `validate_entry_selected_upstream` recurses into branch chains, but no test verifies `selected_upstream` conditions inside branch chains are validated for empty predicates and matcher value existence.
> 
> **5. Missing test: selected_upstream condition in IRR step is validated**
> 
> Similar recursion for `iterative_request_router` steps, but no test coverage.
> 
> **6. Missing test coverage: reachability/dataflow validation**
> 
> The PR description mentions "dataflow validation for reachability" ensuring filters with `selected_upstream` conditions only run after load balancers. The diff shows +839 lines in `checks.rs`, but I could not locate corresponding test functions for this validation in the files reviewed. Critical test cases needed:
> 
> * Filter with `selected_upstream` condition without any preceding load_balancer → reject
> * Filter with `selected_upstream` condition after unconditional load_balancer → accept
> * Filter with `selected_upstream` condition where only conditional load_balancer precedes (might not run) → should reject or warn
> * Load_balancer in branch chain providing metadata for top-level filter
> * Branch chains and SkipTo/ReEnter affecting reachability
> 
> **7. Missing test coverage: runtime evaluation logic**
> 
> While the diff shows +308 lines in `condition/request.rs`, I could not verify tests exist for:
> 
> * Condition matches when upstream metadata is present and matches
> * Condition fails when upstream metadata is present but doesn't match
> * Condition fails when no upstream is selected (fail-closed)
> * Both `when` and `unless` variants
> * Partial matches (only protocol matches but provider doesn't, or vice versa)
> 
> These are core runtime behaviors that must be thoroughly tested.
> 
> ### Medium Severity Gaps
> **8. Missing test: serde round-trip for SelectedUpstreamMatch**
> 
> The type derives `serde::Serialize` but no test verifies serialization works correctly. A round-trip test (serialize → deserialize → verify equality) would catch serialization issues.

I checked each requested scenario and added the two direct coverage cases that were missing:

* selected_upstream_provider_only_round_trips_through_serialization verifies provider-only parsing and serialization round-tripping.
* reject_selected_upstream_in_iterative_router_step verifies matcher-value validation recurses into IRR step filters.

The other scenarios were already covered:

* Pair validation: reject_selected_upstream_pair_split_across_clusters
* Inline clusters: accept_selected_upstream_matching_inline_cluster
* Inline branch chains: reject_selected_upstream_in_inline_branch_chain
* Reachability and control flow: the selected_upstream_* tests in pipeline/checks.rs, including unconditional/conditional load balancers, branch chains, SkipTo, top-level and nested ReEnter, and nested exits
* Runtime evaluation: the selected_upstream_* tests in condition/request.rs, covering matching, mismatches, combined predicates, missing metadata, and both when and unless

crates/core/src/config/condition/request.rs
```diff
@@ -89,6 +89,70 @@ pub struct ConditionMatch {
...
+    /// Selected-upstream application metadata published by the load balancer.
+    ///
+    /// Matches the typed protocol/provider of the cluster the load balancer
+    /// selected for this exchange, not request headers or writable metadata.
+    /// Because the value is published only after upstream selection, a filter
+    /// carrying this predicate requires a load balancer guaranteed to run
+    /// before it (enforced at build time). Missing metadata never satisfies
+    /// the predicate (fail-closed).
+    #[serde(default)]
+    pub selected_upstream: Option<SelectedUpstreamMatch>,
```
shaneutt: The PR body answers "Does this introduce a breaking change? No." That holds for YAML config compatibility (this is an optional field with a serde default, so existing configs keep parsing), but not for the Rust API surface. `ConditionMatch` is `pub`, re-exported as `praxis_core::config::ConditionMatch`, has all-public fields, and is not `#[non_exhaustive]`, so a downstream crate can construct it with a struct literal. Adding `selected_upstream` breaks every such literal (and any exhaustive match), which `cargo-semver-checks` classifies as a major-category change (`constructible_struct_adds_field`).

`praxis-proxy-core` is published (`publish = true`) and the repo runs `cargo-semver-checks` in CI (`.github/workflows/semver.yaml`, plus `make semver`). At `0.6.0` a major-category change needs at least a minor bump (0.6.0 -> 0.7.0); if the version stays on a patch, the semver gate will fail. The PR checklist also shows the local `make lint && make test && make test-integration` gate was left unchecked, and `make semver` is not part of it, so this may not have surfaced yet.

Two clean paths: (a) mark `ConditionMatch` `#[non_exhaustive]` so future optional fields stay non-breaking (note that adding the attribute is itself a one-time breaking change, best landed together with the version bump), or (b) accept the field addition and bump the workspace version to reflect it. Either way, worth reconciling with the "no breaking change" statement in the description.

shaneutt (dismissed):

PR #1242, opened 2026-09-21
Remove ring

twghu:
# Remove `ring`, make the crypto provider an explicit build-time choice

Work spans three repositories that must land together. This document covers all
of it; each repository's PR can link here.

| Repository | Branch | State | Size |
|---|---|---|---|
| `praxis-proxy/pingora` | `remove-ring` | committed (`5fc6d32`) | 10 files, +277/−23 |
| `praxis-proxy/praxis` | `fips-compliance-ring` | uncommitted | 39 files, +520/−250 |
| `rustls-openssl` | `fix/fips-self-report` | 4 committed + 1 uncommitted | 3 files, +349/−66 |

## Intent

`ring` is not FIPS validated and cannot be: it compiles statically into the
binary, so there is no module with a cryptographic boundary for a certificate
to attach to, and no `FIPS_mode()` to interrogate. Praxis targets FIPS 140-3
with the cryptography performed by the platform's `libcrypto.so`.

But the deeper problem was not *which* provider — it was that **nobody chose
one**. `pingora-rustls` installed `rustls::crypto::ring` as the process-wide
default from five call sites, and `praxis-tls` read
`CryptoProvider::get_default()` with a fallback to `aws-lc-rs`. Whichever ran
first won. Pingora's connectors are built while a proxy *service* is
constructed, which happens before listeners are wired — so in practice **ring
was serving every handshake, upstream and downstream**, while the code read as
though aws-lc-rs was.

These changes remove the choice from the transport layer and make it an
explicit, single, build-time decision in the application. `ring` leaves the
dependency graph as a consequence, but the change is provider-neutral: the same
code now works with `aws-lc-rs`, `rustls-openssl`, or any other
`CryptoProvider`.

## 1. `pingora` — make the fork provider-agnostic

- Dropped the direct `ring` dependency and the `ring` feature on **both**
  `rustls` and `tokio-rustls`.
- Enabled rustls' `custom-provider`, which removes rustls' implicit fallback to
  whichever built-in its features happen to enable.
- **Deleted `install_default_crypto_provider()`** rather than stubbing it — a
  function with that name that installs nothing is a trap.
- `hash_certificate()` now borrows SHA-256 from the *installed* provider rather
  than calling `ring::digest`, so the digest comes from the same module as
  every other primitive. The `tls13()` detour is required, not stylistic:
  `SupportedCipherSuite::hash_provider()` and `::common()` are both
  `pub(crate)`.
- Added a NIST FIPS 180-2 known-answer test. A length-only assertion would pass
  against SHA-384; the known answer will not.
- Crate-level rustdoc covering provider selection, with worked examples.

### `pingora-s2n` is deliberately unchanged

It still depends on `ring`. It is a separate, mutually exclusive backend
(`#[cfg(feature = "s2n")]` throughout, `optional = true`), nothing we build
selects it, and it would not help: `s2n-tls → s2n-tls-sys → aws-lc-rs`, so as a
compliance route it lands exactly where `aws-lc-rs` already does. Changing it
would mean touching a backend we never compile and taking on its build
toolchain. Recorded so the omission reads as a decision.

## 2. `rustls-openssl` — a fifth patch

Four commits already on this branch route digests, EC keygen and public-key
import through EVP so they reach OpenSSL's FIPS provider.

A fifth was needed, and it is a hard blocker rather than a refinement:
`impl SigningKey for PKey` did not override `public_key()`, which rustls opts
out of by default. `CertifiedKey::keys_match()` begins by asking the key for its
SPKI and returns `InconsistentKeys::Unknown` when there is none. Praxis calls
`keys_match()` on every certificate load, so **no certificate could load at
all**.

```rust
fn public_key(&self) -> Option<SubjectPublicKeyInfoDer<'_>> {
    self.0.public_key_to_der().ok().map(Into::into)
}
```

Same class as the existing four — upstream not implementing what a real
consumer needs — and a candidate for upstreaming alongside them.

## 3. `praxis` — own the choice

- **`crates/tls/src/provider.rs`** (new): `install()`, `installed()`, `name()`.
  `compile_error!` when no provider is selected.
- **No fallback.** `loader.rs`'s `get_default().unwrap_or_else(aws_lc_rs)` is
  now `TlsError::NoCryptoProvider`. This is the line that made ring live.
- **Install first.** `praxis::install_crypto_provider()` as the first statement
  of `main()`, and again in `run_server_with_composition` for embedded callers.
- **`cipher_suite.rs` now maps to `rustls::CipherSuite` wire identifiers**
  rather than `crypto::aws_lc_rs::cipher_suite` constants. The implementation
  resolves against whatever provider is installed, so the config surface never
  needs remapping again.
- **Feature propagation** through core/filter/protocol/server, with
  `default-features = false` on the workspace `praxis-tls` entry — Cargo
  rejects that override on an inherited dependency, so it cannot go on each
  member.

### Selecting a provider

```console
cargo build                                            # OpenSSL (default)
cargo build --no-default-features --features aws-lc-rs # AWS-LC
```

`openssl` is the default because it is the FIPS 140-3 target: the cryptography
is performed by the platform's `libcrypto.so`, not a statically linked Rust
library.

## Verification

Both providers are tested, and the OpenSSL path was verified against the real
library rather than assumed.

```console
$ make test-unit                                # 7715 passed, 20 blocks, 0 failures
$ make test-provider                            # openssl,  546 tests
$ make test-provider CRYPTO_PROVIDER=aws-lc-rs  # aws-lc-rs, 546 tests

$ cargo tree -p praxis-proxy -e normal -i ring
warning: nothing to print.
```

For the fork, with `aws-lc-rs` removed from dev-dependencies so OpenSSL was the
only provider compiled in: the full suite passed identically (608 lib + 9
integration), the test binaries linked `/lib64/libcrypto.so.3`, and the
known-answer SHA-256 test passed — so `libcrypto` was genuinely invoked and
returned the right answer.

Note `make test-unit` alone does **not** exercise aws-lc-rs: its first pass uses
`--all-features`, which links both providers and resolves to openssl by
precedence. `make test-provider` exists for that, naming features explicitly.

## Decisions worth a reviewer's attention

**Enabling both provider features is permitted, with `openssl` winning.** The
original design called for `compile_error!` on zero *or multiple*. Multiple had
to be relaxed: `make test-unit` runs `--all-features`, which enables every
feature by definition and cannot exclude one. Selecting *neither* remains a hard
error — that is the dangerous case, since such a build compiles and then panics
on first use. Allowing both is safe because exactly one provider is ever
installed and `name()` reports which; the cost is that a both-features build
*links* the unused one, which is why a release build selects via
`default = ["openssl"]`. rustls takes the same position with its own `ring` /
`aws_lc_rs` features.

**`--no-default-features` now also drops the crypto provider.** Seven
invocations across `check`, `test-unit` and `lint` were affected. Handled with a
single `CRYPTO_PROVIDER ?= openssl` Makefile variable. Kept strict deliberately:
"no provider selected" is a build nobody should ship, and the error names the
fix.

**Hidden setup lines in doctests.** Seven doctests construct connectors and need
a provider. They use rustdoc's `#` prefix, so the install runs but is not
rendered. Debatable — a reader copying the visible example into a fresh binary
would hit the panic. Dropping the `#` makes it visible.

## Two bugs found that were not the point of this work

**`--validate` and `--dump` would have panicked in production.** Both return
from `main()` before reaching `run_server`, and both build a
`SubRequestConnector`, which constructs a TLS client config. Installing only on
the serving path — which is what the original plan specified — would have
shipped that.

**Praxis' integration suite had been running on `ring`.**
`tests/utils/src/net/tls.rs` installed it directly, justified by a comment
stating both providers arrive "transitively via Pingora". True once; not after
the fork change.

## Test fallout

Around 140 tests across four crates construct connectors directly and so bypass
bootstrap. Each crate gained a test-only helper that installs first, so no
production code carries test scaffolding. The best of these hooks `praxis-tls`'
certificate fixtures rather than individual tests, because needing certificates
and needing a provider are the same set — a new test gets it for free.

Every one of these failures was `custom-provider` doing its job: turning an
implicit provider choice into a loud one. The blast radius is "everywhere that
builds a `ServerConfig` or `ClientConfig`", not "everywhere that called the
install function", so `grep` finds only the first wave.

## Merge prerequisites

1. **The `[patch.crates-io]` block in `praxis/Cargo.toml` must be removed.** It
   points at `../pingora` and `../../rustls-openssl-fips`; CI has neither path.
   Marked in the file.
2. **The Pingora fork must be published** at a version Praxis can depend on.
3. **The `rustls-openssl` patches must land** upstream or be vendored. The
   `public_key()` fix is uncommitted.
4. **These must land together.** Praxis against the new fork, without the
   provider install, panics at first TLS use — which is the intended
   fail-closed behaviour, not a regression.
5. Not yet run: integration, conformance, security, resilience suites.
6. `make test-provider` should be added to CI, or the aws-lc-rs build has no
   coverage.


twghu:
Some conflicts exists, in PR I will reapply changes when remaining work on jsonwebtoken, rcgen, sha/mac is finished.

shaneutt:
closing as complete: cherry picked into #1254 :+1:

PR #1245, opened 2026-09-21
feat(cloud-events): add authentication capability

### LINKED ISSUE #1244: cloud-events: Add simple authorization capability
At least, a cloud-events receiver should require some kind of Authorization. This issue maps the need of allowing Praxis to send some Authorization header, constructed from an environment variable.

While this is not the optimal approach for credential injection, it is enough for allowing cloud-events filter to send its own authentication.

Long term it is expected that the "services" approach allows the upstream cloud-events server configuration to be fully defined on Praxis configuration and re-use some kind of credential injection or any kind of credential provider.

rikatz:
### What does this PR do?
The current change adds the capability of read an environment variable and turn into an authorization header to be sent to the cloud-events receiver

cloud-events filter lacks authentication capability on it. This means that if a cloud-events receiver requires an authorization header, Praxis has no way of providing it.

Fixes #1244 

### Checklist

- [X] Signed off all commits (`git commit -s`)
- [X] Tests added or updated
- [X] Documentation updated (if applicable)
- [X] `make lint && make test && make test-integration` passes locally


crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -337,6 +354,7 @@ impl CloudEventsFilter {
             .chain(cfg.extensions.values())
             .any(|mapping| matches!(parse_source(&mapping.value), Some(CloudEventsSource::ResponseStatus)));
         let _ = (cfg.delivery, cfg.format);
+        let authorization_token = cfg.authorization.as_ref().map(resolve_authorization).transpose()?;
```
shaneutt: Security: this wires in a bearer credential, but nothing requires `destination` to use TLS. `validate_url_target` accepts plaintext `http://` (the `rejects_non_http_destination` test rejects only a missing scheme, port 0, and a fragment; `http://events.example.test/events` passes validation). Before this PR that was harmless because no credential was sent. Now a misconfigured `http://` receiver means the `Authorization: Bearer <token>` header is transmitted in cleartext over the network. Recommend rejecting the config (or at minimum a loud startup warning) when `authorization` is set and `cfg.destination` is not `https`, so an operator cannot silently leak the token. The docs example uses `https`, but nothing enforces it.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -405,11 +424,23 @@ impl CloudEventsFilter {
...
+            HeaderMap::from_iter([(CONTENT_TYPE, HeaderValue::from_static("application/cloudevents+json"))]);
+        if let Some(token) = authorization_token {
+            let value = format!("Bearer {token}");
+            if let Ok(value) = HeaderValue::from_str(&value) {
+                headers.insert(http::header::AUTHORIZATION, value);
+            } else {
+                crate::metrics::record_cloud_events_skipped("invalid_authorization");
+                return false;
+            }
+        }
```
shaneutt: Two observations on this block.

(1) Zeroizing is bypassed here: `token.to_string()` on a `&Zeroizing<String>` derefs to `String` and yields a plain, non-zeroized `String`, and `format!("Bearer {token}")` makes another. Both are dropped without zeroization on every publish, so the `Zeroizing` wrapper only protects the long-lived stored copy, not the per-delivery ones. The final `HeaderValue` is inherently non-zeroized (it goes on the wire), so complete zeroization is not achievable, but the extra `.to_string()` copy is avoidable and the formatted string could be wrapped in `Zeroizing` to keep the discipline consistent with the stated intent.

(2) `HeaderValue::from_str(&format!("Bearer {token}"))` re-validates a token that `resolve_authorization` already validated at build time via `authorization_header()` on the same immutable value, building the exact same `Bearer {token}` string. In production the `else` branch (record `invalid_authorization`, return false) is therefore unreachable and that skip metric can never fire, which is misleading observability. Either drop the re-check, or if you want defense-in-depth reuse `authorization_header()` instead of duplicating the `format!` + `from_str` logic in two places.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -180,6 +189,14 @@ struct CloudEventsFilterConfig {
     delivery_timeout_ms: u64,
 }
 
+/// Environment-backed authorization for the `CloudEvents` receiver.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
+#[serde(deny_unknown_fields)]
+struct CloudEventsAuthorizationConfig {
+    /// Environment variable containing the bearer token.
+    env_var: String,
```
shaneutt: Design note: the scheme is hard-coded to `Bearer` (this config carries only the env var name, and the value is always emitted as `Bearer <value>`). Two consequences worth weighing: (1) receivers needing `Basic`, a raw API-key header, or any non-`Bearer` scheme are unsupported, despite the generic "authentication capability" framing; (2) an operator who sets the env value to a full credential such as `Bearer abc` or `Basic abc` (a natural assumption) gets a doubled/invalid header `Bearer Bearer abc` with no error. The docs do say "bearer token," but consider either accepting the full header value with no forced prefix, or making the scheme configurable, and/or rejecting a value that already begins with `Bearer ` to catch the double-prefix mistake early.

shaneutt, line 197: Right now we're binding to Bearer, but it's likely there will be other options (e.g. `Basic` or raw API key header or something else) I imagine.

Consider how we could improve the schema here down the road, but we don't have to fix it now: this is experimental just make a follow-up issue on the epic and consider this resolved.

crates/filter/src/builtins/http/observability/cloud_events.rs
shaneutt, line 357: We should add logic to block plaintext here and force TLS, and I don't think there's anything enforcing that right now.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -405,11 +424,23 @@ impl CloudEventsFilter {
 
         let destination = self.destination.clone();
         let timeout = self.delivery_timeout;
+        let authorization_token = self.authorization_token.as_ref().map(|token| token.to_string());
```
shaneutt: Should wrap the string here in Zeroize, and no need to make an extra copy then.

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -405,11 +424,23 @@ impl CloudEventsFilter {
...
+        let mut headers =
+            HeaderMap::from_iter([(CONTENT_TYPE, HeaderValue::from_static("application/cloudevents+json"))]);
+        if let Some(token) = authorization_token {
+            let value = format!("Bearer {token}");
+            if let Ok(value) = HeaderValue::from_str(&value) {
+                headers.insert(http::header::AUTHORIZATION, value);
+            } else {
+                crate::metrics::record_cloud_events_skipped("invalid_authorization");
+                return false;
+            }
```
shaneutt: Isn't this unreachable in practice? this is already validated previously? Might wanna consider if we wanna refactor a bit around signalling this to various points.

shaneutt (dismissed):
A few minor comments to resolve, and praxis-bot comments to resolve, and then 👍

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: https://github.com/praxis-proxy/praxis/issues/1253

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: done on the validate_config (~line 820, pushing soon).

In case there's a configuration AND the scheme is not https the configuration will be refused

crates/filter/src/builtins/http/observability/cloud_events.rs
```diff
@@ -405,11 +424,23 @@ impl CloudEventsFilter {
```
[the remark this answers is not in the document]
rikatz: right now fixed with authorization_header creating the bearer token directly, and the header value to be used on the authorization. Should push it soon

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: copying the reference, and wrapping the string on Zeroize inside the authozation_header method

crates/filter/src/builtins/http/observability/cloud_events.rs
rikatz: fixed simplifying the authorization branch and removed the metric as we never reach it

shaneutt (dismissed):
👍

shaneutt (approved):

PR #1249, opened 2026-09-22
test(security): add missing attack vector coverage

### LINKED ISSUE #637: test(security): add missing attack vector test cases
Attack vectors that still need to be added to the security test suite.

**Response-side injection**

- Upstream response header injection via CRLF in header values
- Upstream Set-Cookie header injection via CRLF
- Literal control characters (\x01-\x08, \x0e-\x1f) in response headers

**Protocol parsing**

- Chunked encoding with invalid chunk sizes (non-hex, overflow)
- Mixed chunked + Content-Length on same request (RFC 9112 compliance)
- Request line exceeding max length (expect 414)
- Absolute URI form in request line (GET http://host/path)

**URL and path encoding**

- Fragment identifier in request URI does not influence routing
- Null byte in query string (%00) handled safely
- Double-encoded path segments don't bypass path matching

**TLS**

- Weak cipher suite rejection
- ALPN mismatch between client and upstream

**Proxy-specific**

- SSRF via request to metadata endpoints (169.254.169.254, localhost)
- Proxy loop detection (Via header chain growth)
- Compression bomb (highly compressed body that explodes on decompression)

**Connection lifecycle**

- Keep-alive poisoning (upstream Connection: keep-alive when proxy expects close)
- HTTP/1.1 request pipelining isolation (responses don't mix)
- Premature backend connection close mid-response

AayushSaini101:
### What does this PR do?

Body: gaps — response injection, protocol parsing, URL encoding, TLS, Via/compression, connection lifecycle, shared backends, and docs/proxy-explained.md

Fixes #637 

### Checklist

- [x] Tests added or updated
- [ ] Documentation updated (if applicable)
- [ ] `make lint && make test && make test-integration` passes locally


tests/security/tests/suite/compression_bomb.rs
```diff
@@ -0,0 +1,85 @@
...
+    assert!(
+        raw.len() < 2_000_000,
+        "response must not expand a 1MiB gzip bomb unboundedly (len={})",
+        raw.len()
+    );
+}
+
+// -----------------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
AayushSaini101: Done thanks : )

tests/security/tests/suite/proxy_loops.rs
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2024 Praxis Contributors
+
+//! Proxy-loop / Via-chain attack-vector tests.
+
+use praxis_core::config::Config;
+use praxis_test_utils::{free_port, http_send, parse_header, parse_status, simple_proxy_yaml, start_backend, start_proxy};
```
shaneutt: **convention (blocking):** Six files in this PR are not rustfmt-clean under the repo's nightly config, so `make lint` (which runs `cargo +nightly fmt --all -- --check`, Makefile:391) fails on the `lint` job in tests.yaml. This line is 121 characters (`max_width = 120`); the others are `compression_bomb.rs`, `connection_lifecycle.rs`, `protocol_parsing.rs`, `response_injection.rs` (`hex_literal_case = "Upper"` wants `0x0E`/`0x1F`) and `tests/utils/src/net/backend/specialized.rs`. Running `make fmt` and committing the result fixes all of them.

tests/security/tests/suite/compression_bomb.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    // Proxy may pass through the compressed body, reject, or error — but must
+    // not hang or expand into an unbounded in-memory bomb during the test.
+    assert!(
+        status == 200 || status == 400 || status == 413 || status == 502 || status == 0,
+        "gzip bomb upstream must be handled safely (got {status})"
+    );
+    assert_ne!(status, 500, "must not crash with 500");
+    assert!(
+        raw.len() < 2_000_000,
```
shaneutt: **test:** This bound can't detect decompression: the fixture inflates to 1,000,000 bytes (not 1 MiB as the doc comment says), so a proxy that fully decompressed it would still pass, and the status allowlist also accepts `0` (no response). In practice Praxis never decompresses here (Pingora's compression module is built with decompression off, and the compression filter skips responses that already have `Content-Encoding`), so it would be more useful to pin that: assert status 200, `parse_header(&raw, "content-encoding") == Some("gzip")`, and `Content-Length` equal to `GZIP_1MB_ZEROS.len()`. Compare headers rather than body length, since `http_send` runs the body through `from_utf8_lossy` and the binary bytes won't round-trip. Also fix the doc comment (or regenerate a true 1 MiB fixture).

tests/security/tests/suite/connection_lifecycle.rs
```diff
@@ -0,0 +1,132 @@
...
+    );
+    let status = parse_status(&raw);
+
+    assert!(
+        status == 200 || status == 502 || status == 0,
+        "keep-alive poison upstream must be handled safely (got {status})"
+    );
+    assert_ne!(status, 500, "must not crash with 500");
+
+    let poisoned = parse_header(&raw, "x-poisoned");
```
shaneutt: **test:** With a single client request this can't observe keep-alive poisoning. The forged response sits after the 4-byte `safe` body on the upstream socket, so it could only leak into a *later* request that reuses the pooled upstream connection, and `parse_header` only reads the header block anyway. The backend also drops the socket as soon as it has written. To exercise the vector, keep the upstream socket open after writing (e.g. block on a read), then send a second request through the proxy and assert it gets body `safe` with no `X-Poisoned` header.

tests/security/tests/suite/connection_lifecycle.rs
```diff
@@ -0,0 +1,132 @@
...
+        assert!(
+            !body.contains("poison"),
+            "poison body must not be appended to the client response: {body}"
+        );
+    }
+}
+
+#[test]
+fn http11_pipelining_responses_do_not_mix() {
+    let backend_port = start_backend("pipe-ok");
```
shaneutt: **test:** This can't detect mixed or misordered responses: both requests get the same body (`pipe-ok`), `!status_lines.is_empty()` accepts a single response, every line is already filtered by `starts_with("HTTP/1.")` before the 1.1/1.0 check, and `HTTP/1.1 200 OKHTTP/1.1` can't occur because status lines end in CRLF. It also doesn't pin what actually happens: Pingora has pipelining off by default and Praxis doesn't enable it, so the proxy answers `/a`, closes, and drops `/b`. Suggest `start_uri_echo_backend()` so each body is its path, then read to EOF and assert exactly one response with body `/a` and no second status line. If pipelining is ever enabled, this becomes an ordered `/a` then `/b` check.

tests/utils/src/net/backend/specialized.rs
```diff
@@ -160,6 +160,124 @@ pub fn start_stateful_backend(responses: Vec<(u16, String)>) -> BackendGuard {
...
+///
+/// # Panics
+///
+/// Panics if the server fails to bind or accept connections.
+pub fn start_mid_response_drop_backend() -> u16 {
+    spawn_tcp_server(|mut stream| {
+        stream.set_read_timeout(Some(Duration::from_secs(5))).unwrap();
+        let mut buf = [0_u8; 4096];
+        let _bytes = stream.read(&mut buf);
+        let _sent = stream.write_all(b"HTTP/1.1 200 OK\r\nContent-Length: 100\r\n");
```
shaneutt: **test:** This backend never finishes the header block, so `premature_backend_close_mid_response_returns_502` covers an incomplete-headers failure, and that case already exists: `tests/resilience/tests/suite/slow_client.rs:56` (`backend_mid_response_failure_returns_502`) uses a byte-identical private `start_mid_response_drop_backend` at `:122` and asserts the same 502. Either drop the new test, or move the resilience test onto this shared helper and delete its private copy. The case #637 asks for (upstream sends full headers with `Content-Length: 100` plus a few body bytes, then closes) is still untested; there the proxy has already committed the 200, so the assertion should be a truncated body and a closed client connection rather than a clean complete message (see `tests/integration/tests/suite/streaming_terminal_response.rs:511` for the pattern).

tests/security/tests/suite/protocol_parsing.rs
```diff
@@ -0,0 +1,189 @@
...
+         Connection: close\r\n\
+         \r\n";
+    drop(stream.write_all(follow_up.as_bytes()));
+    let mut buf2 = vec![0_u8; 4096];
+    let n2 = stream.read(&mut buf2).unwrap_or(0);
+    let second = String::from_utf8_lossy(&buf2[..n2]);
+    let second_status = parse_status(&second);
+
+    assert_ne!(first_status, 500, "TE+CL must not cause 500 on first request");
+    if n2 > 0 {
```
shaneutt: **test:** This branch doesn't run against current behavior: when both `Transfer-Encoding` and `Content-Length` are present, Pingora strips CL and disables keep-alive, so the connection closes after the first response. The vector itself is already covered by `cl_te_desync_does_not_poison_connection` and `te_cl_desync_rejected_or_safe` in `request_smuggling.rs`; what none of them assert is the RFC 9112 section 6.1 requirement that the server MUST close the connection after responding. Rather than a follow-up write, read to EOF with `read_to_end` under the existing 5s timeout and assert it ends with `Ok` (server closed, not a timeout) and the buffer holds exactly one status line. A single `read()` for the first response can split header and body, so don't rely on `n2 == 0`.

tests/security/tests/suite/proxy_loops.rs
```diff
@@ -0,0 +1,54 @@
...
+    let status = parse_status(&raw);
+
+    assert!(
+        status == 200 || status == 400 || status == 502 || status == 0,
+        "long Via chain must complete safely (got {status})"
+    );
+    assert_ne!(status, 500, "must not crash with 500");
+
+    // Response must be finite — no unbounded Via growth visible as multi-MB header.
+    if let Some(via) = parse_header(&raw, "via") {
```
shaneutt: **test:** The 64-hop chain is on the request, but this inspects the response `Via`, which Praxis builds from the upstream response headers plus its own entry (`with_body.rs:461`). `start_backend` sends no `Via`, so the client always sees `Via: 1.1 praxis` and both length bounds always pass. Use `start_header_echo_backend()` and assert the forwarded request `Via` keeps the 64 entries and ends with `1.1 praxis`. Note that Praxis has no Via-based loop detection today, so the "proxy loop detection" item from #637 isn't something a test can cover yet; that would be a product change worth tracking separately.

tests/security/tests/suite/response_injection.rs
```diff
@@ -0,0 +1,115 @@
...
+};
+
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+#[test]
+fn upstream_response_crlf_header_value_not_injected() {
+    // Double-CRLF response splitting: embedded blank line + forged status.
+    let malformed = b"X-Safe: ok\r\n\r\nHTTP/1.1 200 OK\r\nX-Injected: evil".to_vec();
```
shaneutt: **test:** This payload doesn't put CRLF inside a header value: the blank line ends the upstream header block, so upstream sends a valid response with just `X-Safe: ok` and a read-until-close body that happens to look like a response. The proxy forwards that as body, and `parse_header` only reads the client's header block, so `x-injected` can never be found here. An obs-fold continuation exercises the real parser path: with `X-Safe: ok\r\n X-Injected: evil`, Pingora accepts obs-fold in upstream responses and folds it to one space, so assert status 200, `parse_header(&raw, "x-safe") == Some("ok X-Injected: evil".into())`, and `parse_header(&raw, "x-injected").is_none()`.

tests/security/tests/suite/ssrf_runtime.rs
```diff
@@ -199,3 +202,37 @@ fn http_proxy_allows_hostname_upstream_with_override() {
...
+        &format!("127.0.0.1:{proxy_port}"),
+        "GET http://169.254.169.254/latest/meta-data/ HTTP/1.1\r\n\
+         Host: 169.254.169.254\r\n\
+         Connection: close\r\n\
+         \r\n",
+    );
+    let status = parse_status(&raw);
+
+    assert!(
+        status == 400 || status == 502 || status == 200 || status == 0,
```
shaneutt: **test:** This accepts exactly what a real SSRF attempt would produce, so it can't detect one: if Praxis connected to the absolute-form authority, the result would be 502 or 0 when 169.254.169.254 is unreachable, or 400 from Azure IMDS when the `Metadata` header is missing (as on GitHub-hosted runners). The only checked branch is 200, and that just looks for the substring `ok`. Suggest aiming the absolute-form at a second local listener, e.g. `start_tcp_tagged_backend("ssrf-target")` on `127.0.0.1:<p2>`, with a matching `Host: 127.0.0.1:<p2>` (Pingora rejects a Host/authority mismatch with 400 before routing), then assert status 200, body equal to the configured backend's `ok`, and no `ssrf-target` in the response.

tests/security/tests/suite/tls_attack.rs
```diff
@@ -0,0 +1,161 @@
...
+    TestCertificates, free_port, https_get, start_backend_with_shutdown, start_tls_proxy, tls_connection_rejected,
+};
+use rustls::ClientConfig;
+
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+#[test]
+fn weak_cipher_suite_client_rejected() {
```
shaneutt: **test:** This duplicates `listener_cipher_suite_restriction_rejects_excluded_suite` in `tests/integration/tests/suite/tls.rs:2202` (same TLS 1.3-only `cipher_suites`, same TLS 1.2-only client; only the ALPN value differs, which doesn't matter when the handshake fails at suite selection). It also isn't a weak-cipher test: a rustls TLS 1.2 client only offers ECDHE AEAD suites. If this should add coverage, configure a TLS 1.2 suite list (the `tls12_ecdhe_*` IDs exist in `crates/tls/src/config/cipher_suite.rs`) and assert a client restricted to an excluded suite is refused while one with an allowed suite succeeds; otherwise drop it.

tests/security/tests/suite/tls_attack.rs
```diff
@@ -0,0 +1,161 @@
...
+    );
+}
+
+#[test]
+fn alpn_mismatch_between_client_and_listener_fails_closed() {
+    let certs = TestCertificates::generate();
+    let backend = start_backend_with_shutdown("alpn-ok");
+    let proxy_port = free_port();
+
+    // Listener does not advertise h2; HTTP/1.1 only.
```
shaneutt: **bug:** The premise is wrong: HTTP TLS listeners advertise ALPN `[h2, http/1.1]` (`crates/protocol/src/http/pingora/listener.rs:30` passes `advertise_http_alpn = true`, see `crates/tls/src/setup.rs:57`). The h2-only client negotiates h2 successfully, Pingora starts an HTTP/2 handshake, the HTTP/1.1 bytes fail the preface, and `tls_connection_rejected` counts the dropped connection as "rejected". No ALPN mismatch ever happens, so a regression in mismatch handling wouldn't be caught. Offer a protocol the listener doesn't advertise (e.g. only `b"h3"`) and assert the handshake itself fails; rustls refuses a non-overlapping offer with `no_application_protocol`.

tests/security/tests/suite/url_encoding.rs
```diff
@@ -0,0 +1,171 @@
...
+        assert!(
+            status_b == 400 || status_b == 0 || status_b == status_a,
+            "fragment URI must be rejected or match non-fragment routing (got {status_b})"
+        );
+    }
+}
+
+#[test]
+fn null_byte_in_query_string_handled_safely() {
+    let backend = start_uri_echo_backend();
```
shaneutt: **test:** The URI echo backend is started but its output is never checked, so the test can't tell whether the query reached upstream unmodified. A literal NUL can't be forwarded (Pingora refuses it and the 502 would already fail this test), but truncation at `%00` (`/search?q=hello`), re-encoding (`%2500`), or decoding to another byte all pass on status 200. When the status is 200, assert `parse_body(&raw) == "/search?q=hello%00world"`.

tests/security/tests/suite/compression_bomb.rs
shaneutt: I don't think this test is detecting decompression: the fixture is 1,000,000 bytes, and the status check even allows `0`. Assert status 200, preserve `Content-Encoding: gzip`, and verify `Content-Length == GZIP_1MB_ZEROS.len()`.

We should check the headers rather than the body, since `http_send` lossy-decodes binary data. Also need to update the fixture’s size documentation.

tests/security/tests/suite/connection_lifecycle.rs
shaneutt: A single request can’t detect keep-alive poisoning; the forged response would only affect a later request reusing the upstream connection. We'll need to keep the upstream socket open after writing, send a second request through the proxy, and assert it returns `safe` without `X-Poisoned`.

tests/security/tests/suite/connection_lifecycle.rs
```diff
@@ -0,0 +1,132 @@
...
+        let body = parse_body(&raw);
+        assert!(
+            !body.contains("poison"),
+            "poison body must not be appended to the client response: {body}"
+        );
+    }
+}
+
+#[test]
+fn http11_pipelining_responses_do_not_mix() {
```
shaneutt: 🤔 this doesn’t detect mixed or misordered responses both requests return the same body, and the assertions allow a single response.

Try with `start_uri_echo_backend()` so responses identify their paths, read to EOF, and assert exactly one response with body `/a` and no second status line (this also verifies the current behavior: pipelining is disabled, so `/b` is dropped)

tests/security/tests/suite/protocol_parsing.rs
shaneutt: Does this branch actually run against the current behavior? 🤔 pingora strips `Content-Length` and closes the connection when both transfer headers are present the desync vector is already covered, add the missing RFC 9112 assertion by reading to EOF with `read_to_end` under the existing timeout, then verify the read completes successfully and the buffer contains exactly one status line. We can't rely on a single `read()` returning zero.

tests/security/tests/suite/proxy_loops.rs
shaneutt: Need to run `make fmt` and `make lint` and commit the generated updates.

tests/security/tests/suite/proxy_loops.rs
shaneutt: This checks the response `Via`, which only contains praxis’ own entry because `start_backend` sends none. Use `start_header_echo_backend()` and assert the forwarded request preserves all 64 entries and ends with `1.1 praxis`. Loop detection isn’t implemented yet, so track that separately from this test.

tests/security/tests/suite/response_injection.rs
shaneutt: payload ends the upstream header block instead of putting CRLF inside a header value, so `x-injected` is forwarded as body and can’t be found by `parse_header`. Need an obs-fold continuation (`X-Safe: ok\r\n X-Injected: evil`) and assert status 200, the folded `x-safe` value, and that `x-injected` is absent.

tests/security/tests/suite/ssrf_runtime.rs
shaneutt: I don't think this case can detect SSRF, a real request to the metadata address would return 502, 0, or 400, while the test only accepts 200 containing `ok`. Need to point the absolute-form URL at a second local listener, set a matching `Host`, and assert the request reaches the configured backend, returns the expected body, and never includes `ssrf-target`.

tests/security/tests/suite/tls_attack.rs
shaneutt: This duplicates `listener_cipher_suite_restriction_rejects_excluded_suite` the ALPN difference doesn’t affect the handshake failure, and the test doesn’t cover weak ciphers. Could just go ahead and add TLS 1.2 cipher-suite coverage using the `tls12_ecdhe_*` IDs, asserting excluded suites fail while allowed suites succeed, to recycle it

tests/security/tests/suite/tls_attack.rs
shaneutt: Wait... does this actually test an ALPN mismatch?

tests/security/tests/suite/url_encoding.rs
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+        assert!(
+            status_b == 400 || status_b == 0 || status_b == status_a,
+            "fragment URI must be rejected or match non-fragment routing (got {status_b})"
+        );
+    }
+}
+
+#[test]
+fn null_byte_in_query_string_handled_safely() {
```
shaneutt: The URI echo backend’s output isn’t asserted so truncation, re-encoding, or decoding of `%00` could pass through here.

tests/utils/src/net/backend/specialized.rs
```diff
@@ -160,6 +160,124 @@ pub fn start_stateful_backend(responses: Vec<(u16, String)>) -> BackendGuard {
...
+}
+
+/// Start a backend that writes a partial response (status +
+/// `Content-Length`) then drops the connection before finishing
+/// headers or body. Used for mid-response failure tests.
+///
+/// # Panics
+///
+/// Panics if the server fails to bind or accept connections.
+pub fn start_mid_response_drop_backend() -> u16 {
```
shaneutt: this seems to duplicate the existing incomplete-header 502 test including its private helper? 🤔

shaneutt (changes requested):
Thanks @AayushSaini101, I have some comments for you to check out LMKWYT 🖖

PR #1250, opened 2026-09-22
refactor(sni): Introduce consolidated `SniMatcher`

alexsnaps:
## tl;dr

New deep module — crates/tls/src/sni_match.rs: `SniMatcher<V>` with a small interface (`build(entries, policy)` + `lookup(sni)`)

## Assumptions:

- The two match semantics are deliberate, not accidental. Inferred "intended" from each side having a test that pins it (`resolve_wildcard_rejects_multi_level for certs`, `longest_wildcard_wins` for routing). But neither test frames itself as an intentional spec divergence, and git blame was muddied by later refactor/move commits — so could not confirm original author intent, only that the behaviour was tested.
- Operators actually rely on the router's multi-label suffix matching. Kept both decision assumes changing *.example.com to stop matching a.b.example.com would break real configs. We have a test proving the behaviour exists, not evidence anyone depends on it.
- The cert path's lack of trailing-dot handling was an incidental inconsistency, not a deliberate cert-specific choice. No comment or test spoke to it either way; we assumed it was an oversight and therefore assumed safe to "fix" by normalizing for both.
- Normalizing a trailing dot on the cert path is safe. We assumed example.com. and example.com are the same host, so serving the example.com cert for example.com. is correct and not a cert-confusion/security risk. Grounded in DNS/RFC understanding, not verified against rustls/client behaviour.
- The ≥3-label rule (*.com) is cert-identity-specific, and routing legitimately wants *.com permitted. Assumed a *.com routing rule is a valid use case worth keeping, so the rule co-varies with SingleLabel rather than applying to both.

### What does this PR do?

It should do _nothing_ (modulo, the trailing dot, see assumptions), it's a refactor to clean things up and avoid things drifting in the future

Fixes #

### Checklist

- [x] Tests added or updated
- [x] `make lint && make test && make test-integration` passes locally


crates/tls/src/sni_match.rs
```diff
@@ -0,0 +1,485 @@
...
+//!
+//! - [`WildcardMatch::SingleLabel`] — certificate identity (RFC 6125, superseded by [RFC 9525]). A wildcard matches
+//!   exactly one leftmost label, and a pattern must have at least three labels (`*.com` is rejected).
+//! - [`WildcardMatch::Suffix`] — routing convenience with no governing RFC. A wildcard matches any number of leading
+//!   labels, longest suffix winning, and `*.com` is permitted.
+//!
+//! Exact matches always win over wildcards. Matching is case-insensitive
+//! ([RFC 4343]); a single trailing dot on the looked-up hostname is ignored.
+//!
+//! See `CONTEXT.md` (TLS / SNI) and `docs/adr/0001-sni-wildcard-match-policies.md`
```
leseb: This file does not exist?

crates/tls/src/sni_match.rs
```diff
@@ -0,0 +1,485 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2024 Praxis Contributors
```
leseb: it's 2026 :)

leseb (changes requested):
nits also got this from AI:

P2 — Public API break: [SniNameError::WildcardTooShallow](https://github.com/praxis-proxy/praxis/blob/eb5d909797129631615e4569da35b92faddf8573/crates/tls/src/sni_name.rs#L59-L64) extends a published, exhaustive enum. Downstream exhaustive matches will stop compiling. Represent this policy error in the new SniMatcherError or reserve it for a breaking release.

but this sounds very unlikely at the moment? the impact is speculative, no need to address this imo

crates/tls/src/sni_match.rs
alexsnaps: Not in this repo indeed. Let me fix that. See the first assumption in this PR's description tho. 
I still don't know whether that was intentional yes or no tbh. As I say, given they were tests, I assumed they were. But... since AI writes the tests, we're left with the infamous question "who tests the tests?" ;P

crates/tls/src/sni_match.rs
alexsnaps: Time flies!

alexsnaps:
>Public API break

I consciously ignored that bit. Given where we are in terms of release and cadence. Now if there is a scheme for that, let me know (whether it is fine in this case yes or no actually). I have so many breaking changes coming up (if I get to them)

alexsnaps:
> * The two match semantics are deliberate, not accidental. Inferred "intended" from each side having a test that pins it (`resolve_wildcard_rejects_multi_level for certs`, `longest_wildcard_wins` for routing). But neither test frames itself as an intentional spec divergence, and git blame was muddied by later refactor/move commits — so could not confirm original author intent, only that the behaviour was tested.

This assumption is probably the most important one. If this was _not_ deliberate this can be simplified quite a bit!

Wildcard SNI matching lives in one module (`praxis_tls::sni_match::SniMatcher`) but is parameterised by an explicit `WildcardMatch` policy, because the two callers answer to different authorities and must(?) keep different semantics.
 - *Certificate selection* uses `SingleLabel`: a wildcard matches exactly one leftmost label (`*.example.com` matches `app.example.com` but not `a.b.example.com`) and patterns need at least three labels - this is the RFC 6125 / RFC 9525 certificate-identity rule, and a CA will not issue a `*.com` certificate. 
 - *SNI routing* uses `Suffix`: a wildcard matches any number of leading labels with the longest configured suffix winning, and `*.com` is permitted — a routing convenience no RFC governs.

The divergence is deliberate, not a bug to unify: both behaviours are covered by existing tests (`resolve_wildcard_rejects_multi_level` for certs, `longest_wildcard_wins` for routing). Collapsing them to one semantics would either break routing (converge single-label) or serve `*.example.com`
certificates for `a.b.example.com`, violating certificate identity (converge suffix).

leseb (approved):

alexsnaps:
Still no write access for me, so one of you will have to click that button  ¯\_(ツ)_/¯

crates/tls/src/setup/sni.rs
```diff
@@ -749,8 +708,13 @@ mod tests {
...
             let resolver = build_sni_resolver(&certificates).expect("resolver build");
-            let exact_ptr = Arc::as_ptr(resolver.certs.get("api.example.com").expect("exact cert")) as usize;
-            let wildcard_ptr = Arc::as_ptr(resolver.wildcard_certs.get("example.com").expect("wildcard cert")) as usize;
+            // Derive the certificate identities through the public lookup:
+            // the exact name yields the exact cert, a single-level subdomain
+            // yields the wildcard cert.
+            let exact_cert = resolver.lookup(Some("api.example.com")).expect("exact cert");
+            let exact_ptr = Arc::as_ptr(&exact_cert) as usize;
+            let wildcard_cert = resolver.lookup(Some("other.example.com")).expect("wildcard cert");
+            let wildcard_ptr = Arc::as_ptr(&wildcard_cert) as usize;
```
shaneutt: **test:** The fixture now derives `exact_ptr` and `wildcard_ptr` from `resolver.lookup(...)`, which is the function these properties test, so `exact_beats_wildcard` no longer proves its name: if exact matches stopped taking precedence, `lookup("api.example.com")` would return the wildcard cert, `exact_ptr` would be that same pointer, and the property would still pass. Before this change the fixture read the two certs straight from the internal maps, independent of the code under test. A cheap fix is to assert in the fixture that the identities differ, which also lets the message replace the inline comment: `assert_ne!(exact_ptr, wildcard_ptr, "exact name must resolve to the exact cert, not the wildcard cert");`. Low severity since `SniMatcher` has `exact_match_wins`, but the resolver-level property should still be able to fail.

crates/tls/src/sni_match.rs
```diff
@@ -0,0 +1,482 @@
...
+            .map(|(pattern, value)| ((*pattern).to_owned(), (*value).to_owned()));
+        SniMatcher::build(owned, policy).expect("test matcher should build")
+    }
+
+    /// Look up `sni` and return the matched value as `&str`, if any.
+    fn resolve<'matcher>(matcher: &'matcher SniMatcher<String>, sni: &str) -> Option<&'matcher str> {
+        matcher.lookup(sni).map(String::as_str)
+    }
+
+    // --- exact matching (both policies) ---------------------------------------
```
shaneutt: **convention:** `docs/developing/conventions.md` (Separator Comments) says separators must be full-width and explicitly "Never: `// --- Section Name ---`", and under "Inside `#[cfg(test)] mod tests`" puts test utilities at the end under a `// Test Utilities` separator. This module has five short-form separators (here and at 344, 385, 427, 471) and `make`/`resolve` at the top; it's the only file in `crates/` that does either. Dropping the per-group separators and moving the two helpers to the end under a full-width `// Test Utilities` separator brings it in line.

crates/tls/src/sni_match.rs
```diff
@@ -0,0 +1,482 @@
...
+pub enum WildcardMatch {
+    /// A wildcard matches exactly one leftmost label (RFC 6125 / RFC 9525
+    /// certificate-identity semantics): `*.example.com` matches
+    /// `app.example.com` but not `a.b.example.com`. Patterns must have at
+    /// least three labels, so `*.com` is rejected at build.
+    SingleLabel,
+
+    /// A wildcard matches any number of leading labels, longest configured
+    /// suffix winning: `*.example.com` also matches `a.b.example.com`. A
+    /// routing convenience with no governing RFC; `*.com` is permitted.
```
shaneutt: **question:** On the open "is the divergence deliberate?" question: for routing it looks deliberate. The generated operator docs already promise "exact-match lookup first, then longest-suffix wildcard match" (`docs/filters/tcp/traffic_management/sni_router.md:10`, predating this PR), so moving `sni_router` to single-label would break a documented contract rather than tidy an accident. One data point this module doesn't cover yet: the HTTP router has a third wildcard host matcher (`host_matches` in `router/matching.rs:100`) that defaults to single-level and opts into suffix matching via `multi_level_subdomain_matching`, which `router.md` says Gateway API needs. So the codebase treats multi-level as a real routing need, but the two routers disagree on the default. Not blocking here; if this is meant to be the single source of truth, a follow-up could route HTTP `host_matches` through `WildcardMatch` so the routers can't drift, and this doc could note the default difference (HTTP: single-level unless opted in; SNI: always suffix).

PR #1254, opened 2026-09-23
feat(fips): all crypto through the system OpenSSL, a FIPS build on UBI 9, and FIPS scanner in CI

shaneutt:
There's now a standard build (non-FIPS) and an opt-in FIPS build. In the FIPS build Praxis is doing all of its cryptography in the system OpenSSL.

Changelog:

* **One crypto provider.** rustls stays the TLS engine, but every primitive goes through the OpenSSL-backed provider from our Pingora fork (`quixotic-plecostomus-rustls-openssl`). No build-time switch, no ring, no aws-lc in the binary. Library consumers must now call `praxis_tls::provider::install()` before building a listener or connector, and they can turn the policy engine off (`praxis-filter` without default features). (cherry-picked from @twghu)
* **FIPS build and image.** `make release-fips` and `make container-fips` build with the policy engine off (its deps carry their own crypto) on `ubi9/ubi` with Red Hat's `rust-toolset`, shipping on `ubi9/ubi-minimal`. Base images are pinned by digest and their Red Hat signatures are verified before every build. Release, nightly and manual publish push the same tags with a `-fips` suffix. The standard Alpine image is unchanged in shape, but now links Alpine's OpenSSL dynamically since there is no bundled crypto anymore.
* **Runtime behaviour.** Startup logs the provider and kernel FIPS status. `PRAXIS_REQUIRE_FIPS=1` refuses to start unless both say FIPS, naming what is missing. TLS 1.2 requires Extended Master Secret everywhere (FIPS wants it and rustls only counts a config as FIPS with it).
* **Compliance tooling, in `cargo xtask fips`.** `report` mirrors what Red Hat's scanner (openshift/check-payload) checks: crate denylist in the dependency graph, bundled-crypto symbols, the cargo-auditable manifest, and source guards, each finding with a reason and a fix. `verify-image` proves a UBI digest is Red Hat's. `make fips-check` runs the report inside the UBI builder; `make fips-scan` runs check-payload itself against the image with warnings fatal. `make lint` now includes the graph check and a `fips` workflow runs the whole chain on every PR.
* **cargo-auditable manifest.** The FIPS binary embeds its crate list. It is built from cargo's SBOM precursor rather than `cargo metadata`, because metadata resolves weak features (`ring?/alloc` from rustls-webpki) and puts `ring` in the manifest of a binary that never compiled it, which the scanner fails on by name. The details and the removal condition are in `docs/developing/fips.md`.
* Docs: `docs/operating/fips.md` for operators, `docs/developing/fips.md` for the tooling.
* **MSRV to 1.92**

Manually tested in addition to automated tests. The `make fips-check` target currently reports no findings on UBI 9 and check-payload passes the image with `--fail-on-warnings`. I haven't tested on a FIPS-enabled RHEL 9 host yet. @hexfusion if you wouldn't mind trying this out and testing it yourself I would appreciate it. :vulcan_salute:

```console
make fips-check          # build on UBI 9 with Red Hat's toolchain and print the report
make container-fips   # praxis:<version>-fips
make fips-scanner && make fips-scan
```


nerdalert (approved):
Reviewed the provider initialization, feature gating, standard and UBI container changes, FIPS enforcement, EMS handling, compliance tooling, and operator/developer documentation. I did not find a blocking code issue. The implementation consistently routes cryptographic operations through system OpenSSL, installs the provider before TLS configuration is constructed, and fails closed when FIPS is required but unavailable. All current checks are green.

